PR #2, opened 2020-03-27
copy and paste all files from new_datastore_ 533c325913ee7d78c1d2848a…

andybue:
…51225a1600e27d8f


andybue:
Now, the repo is on the same level as Reference/new_datastore commit 533c325913ee7d78c1d2848a and includes PRs from Joakim

PR #23, opened 2020-04-16
22 integrate sfc improvements

jdhoffa:
@Clare2D I think these are the main changes I made. This was pretty difficult to pull in, so hopefully this works. Please review the files changed and merge when you feel its ok to.


Clare2D:
So I will merge this and try and investigate why we need the mean company weight. It seems that was the primary change made? There are a lot of spacing variations that make it tricky to find the core of the problem there. The company sector changes are fine.

PR #30, opened 2020-04-27
Initialize package

maurolepore:


DESCRIPTION
```diff
@@ -2,11 +2,16 @@ Package: PACTAanalysis
 Title: What the Package Does (One Line, Title Case)
```
maurolepore, line 2: @andybue please change this to something meaningful.

DESCRIPTION
```diff
@@ -2,11 +2,16 @@ Package: PACTAanalysis
...
-             role = c("aut", "ctr", "cre"),
-             email = "maurolepore@gmail.com",
-             comment = c(ORCID = "0000-0002-1986-7988")),
+    c(person(given = "Clare",
+             family = "Murray",
+             role = c("aut", "cre"),
+             email = "clare@2degrees-investing.org"),
+      
+     person(given = "Andy",
+             family = "Büchler",
```
maurolepore, line 11: TODO: Please ensure this will work for everyone. Maybe search the unicode equivalent.

maurolepore:
@andybue, thanks for meeting me today.

This PR transforms the PACTA_analysis repo into the PACTAanalysis R package (we removed the underscore because it's invalid in R package names). 

Notes:

* While the structure is now that of an R package, the package is "empty" -- for example, there is nothing in R/ or data/. 

* All the useful code (.R files) is still in the root. They should continue to work as they have so far, but they are not part of the package. R packages take code not in the root but inside the R/ folder. (All .R code is now listed in .Rbuildignore to allow the R package to build despite the invalid .R code in the root.)

* As you more files from the root to R/ you may also want to add tests. Tests go in the directory tests/testthat/. You can automate the process with `usethis::use_r()` and `usethis::use_test()`. The the [documentation of the usethis package](https://usethis.r-lib.org/reference/index.html) for these and many other useful functions.

* To extend this PR you can use the command line instructions at the end of this (and every) PR. They look like this:

image

I'm lazy and forgetful so instead I use `usethis::pr_fetch()`. See more helpful PR helpers [here](https://usethis.r-lib.org/reference/index.html#section-pull-requests).

* Finally, you will see this PR has some comments like TODO and FIXME. You can search for those literal strings with RStudio using a global search (Shift + Control + F).

Let me know what questions you have.

Clare2D:
Hi Mauro, I just saw this Pull request recently and am not sure whether we should revert back and have another look at this. Ie. Does it still make sense to create this repo as a package? Or is it difficult at this stage?

maurolepore:
If this PR has been needless so far I suggest we close it. We can reopen it anytime if we later find it useful.

> Does it still make sense to create this repo as a package? Or is it difficult at this stage?

In general, I think any duplicated code is best to extract into a function and maybe a package. But I recognize that some code is more suitable not to a package but to an analysis script. I realize I lack the vocabulary to articulate what exactly that script-like code is compared to package-like code. I'll try to find something clean online; but in the meantime my best attempt it this: Anything that relates to a specific project, dataset, or directory in your files system should be in a script -- not a package. 

A week ago Jackson suggested we should add to the r2dii package-ecosystem some of the code you are working on. He mentioned the Reference repo but maybe there is something here too. So maybe we should all discuss how to collaborate on this.

Clare2D:
Thanks Mauro. It was the intention to add the PACTA analysis to the r2dii.analysis package, however I'm a bit afraid that this will be extremely complicated. This should ideally be created as a package as it is used repeatedly like the banks software, and is already being used externally. I see this as the next major task! 
For information, the Reference repo has been superseded by this repo. 
I'll close this for now then.

maurolepore:
Thanks. Then I'm happy to help package it; if that's sounds good to you, we need to first check with Klaus when to do so.

Clare2D:
Great! Yep, we should have a planning session for resourcing and next steps
at some point soon. Will get back to you about when that should be.

PR #39, opened 2020-07-01
Current web functionality

Clare2D:


Clare2D:
We won't merge this branch into master but keep this as a branch.

PR #41, opened 2020-07-13
round exchange_rate_usd and unit_share_price

jacobvjk:
This change avoids most issues of duplication of ISINs due to tiny floating point diffs. Such problems caused an error when left_joining portfolio data and financial data, as duplicates of distinct holdings were introduced.
@Clare2D


Clare2D:
Thanks for this Jacob. I'm just wondering about the second commit and whether the "duplicate_isins_check.R" should be added as a file. I would rather envisage that this happens in the data_preparation code approximately here: 
https://github.com/2DegreesInvesting/data_preparation/blob/f42592a318f64759ebf40a6158c460c4cc435229/0_data_preparation_functions.R#L114 
Ie after the file is read in

Clare2D:
The rounding looks good though

jacobvjk:
> Thanks for this Jacob. I'm just wondering about the second commit and whether the "duplicate_isins_check.R" should be added as a file. I would rather envisage that this happens in the data_preparation code approximately here:
> https://github.com/2DegreesInvesting/data_preparation/blob/f42592a318f64759ebf40a6158c460c4cc435229/0_data_preparation_functions.R#L114
> Ie after the file is read in

That is very true, it shouldn't be in here. I forgot to delete the file before pushing. Will fix and update this.

Clare2D:
Thanks!

jacobvjk:
removed

Clare2D:
thanks

PR #46, opened 2020-07-21
data QA markdown 0.1.0

jacobvjk:
@Clare2D since I cannot create new branches on the 2Dii main fork, this would get merged right into master, if I open a PR. Up to you, we can do that. It would not destroy anything, but maybe you could just create another such branch on the 2dii fork off master, so we can merge it into that one instead?


Clare2D:
I believe I've made the branch. Let me know if that didn't work!

Clare2D:
Although I think merging this automatically into Master is fine. I envisage this as one script that is run at any stage and uses generally the 1_project_init script to set up paths etc. Make sense? Maybe including a check to see that the required files are there as well?

jacobvjk:
Yea, that sounds like a good idea. I will add this functionality before opening the actual PR

jacobvjk:
The files are now being read in automatically from the corresponding PACTA project
I also changed the base branch to "qa"
Will open the PR for review
@Clare2D

PR #47, opened 2020-07-22
Qa

Clare2D:


wip_pacta_data_qa.Rmd
```diff
@@ -0,0 +1,545 @@
...
+
+* **total_holdings** is the number of holdings per investor and portfolio
+* **sum_value_usd** is the aggregated value of the holdings in USD by investor and portfolio
+* **sum_market_value** .... see comment below
+* **pct_holdings** is the share of holdings of a specific PF, relative to the overall number of holdings in the analysis. It's the weight of the PF in pure terms of different holdings.
+* **value_weight** is the share of the value of a specific PF in USD, relative to the overall value across all PFs in the analysis. It shows how much value each of the PFs make up for
+
+**OPEN:**
+
+*sum_market_value* only makes sense if all holdings in a portfolio share the same currency. Otherwise, the overall value can look artificially inflated or deflated. Additionally grouping by currency would alleviate this, but the information on the PF level will not be readable anymore.
```
Clare2D, line 88: Yeah I think this is the challenge. I think remove this "sum_market_value" as it could confuse I would suggest

wip_pacta_data_qa.Rmd
```diff
@@ -0,0 +1,545 @@
...
+datatable(investors_overview)
+```
+
+### Total holdings and value per investor and portfolio
+Number of holdings and aggregated value in USD of the portfolio input - grouped by investor and portfolio
+
+* **total_holdings** is the number of holdings per investor and portfolio
+* **sum_value_usd** is the aggregated value of the holdings in USD by investor and portfolio
+* **sum_market_value** .... see comment below
+* **pct_holdings** is the share of holdings of a specific PF, relative to the overall number of holdings in the analysis. It's the weight of the PF in pure terms of different holdings.
```
Clare2D, line 83: Weight of the portfolio in all of the overall investor's portfolios? I'm not sure I get this one and the next.

wip_pacta_data_qa.Rmd
```diff
@@ -0,0 +1,545 @@
...
+## Currencies
+
+### currencies in portfolio and missing currency info
+1. Overview of all currencies present in the portfolio and their respective overall market values (in units of that currency)
+2. Overview of holdings with missing exchange rate information. For successful analysis, exchange rate information is required, so it should be checked if exchange rates can be manually added in these case.
+
+**Notes:**
+
+* Multiple entries of the same ISIN in the missing-table sometimes due to different holding IDs.
+* **market_value_miss_currency** is the aggregated value of the holdings with missing currency information per Investor and PF.
```
Clare2D, line 125: This is great. Perhaps a statement as to whether there are any missing currencies

wip_pacta_data_qa.Rmd
```diff
@@ -0,0 +1,545 @@
...
+## Missing Bloomberg data
+
+### Overview of holdings with missing Bloomberg data
+Check if data for these ISINs can be downloaded from Bloomberg and added to DataStore by Taylor
+Use the .csv file generated by the PACTA_analysis script to pass on the information. This DataTable is supposed to help understand more about individual cases.
+For ISINs that are not in DataStore, check if these are unlisted items (??)
+
+* **share_missing_value** shows the overall impact per ISIN and asset_type
+
+**OPEN:** should financial_sector == 'Other' be filtered out for convenience?
```
Clare2D, line 165: It could be.. I don't think it matter so much. 
I think directly sorting by "share_missing_value" could be more valuable. 
Also if you can change the .x and .y to reflect what they are that should help. I think the sum_value_usd.x is a duplicate of the value_usd

wip_pacta_data_qa.Rmd
```diff
@@ -0,0 +1,545 @@
...
+  filter((flag == "Invalid or missing ISIN" | isin == "MissingValue") &
+           investor_name != "Meta Investor") %>% 
+  select(investor_name, portfolio_name, isin, company_name,  holding_id, #market_value, currency,exchange_rate,
+         value_usd, asset_type, financial_sector, fund_isin, flag) %>% 
+  group_by(fund_isin) %>% 
+  mutate(sum_value_usd = round(sum(value_usd, na.rm=T))) %>% 
+  ungroup() %>% 
+  mutate(
+    share_missing_value = round(
+      sum_value_usd/(investors_overview %>% filter(investor_name=="Meta Investor") %>% pull(sum_value_usd)), 7
```
Clare2D, line 212: Possibly removing the "Meta Investor" reference here could help make this more useful at some point. Ie choose the investor.

wip_pacta_data_qa.Rmd
```diff
@@ -0,0 +1,545 @@
...
+
+**Actions:**
+
+* if the ISIN is a "MissingValue" because it is not available in the BBG database, there is nothing we can do (??)
+* if the ISIN seems to have an odd format, check (potentially with the project partner) if the identifier is really aan ISIN, or some other identifier such as a FIGI. If so, update the identifier.
+* the invalid ISIN might also just be a typo, verify this for cases that look close to valid.
+* These actions are taken on the portfolio input level, so no need to involve the data team
+
+```{r}
+ISIN_missing_invalid <- result_data_qa %>%
```
Clare2D, line 202: the share_missing_value values here seem to be all the same despite differences in the value_usd. Not sure why

wip_pacta_data_qa.Rmd
```diff
@@ -0,0 +1,545 @@
...
+      )
+    ) %>% 
+  arrange(desc(share_missing_value), fund_isin)
+datatable(ISIN_missing_invalid)
+
+```
+
+## Missing ALD
+
+### overview Equity holdings w/o ALD
```
Clare2D, line 222: I think a little more description of what this is is required. I am not sure from the first take on this.

wip_pacta_data_qa.Rmd
```diff
@@ -0,0 +1,545 @@
...
+           financial_sector != "Other" &
+           investor_name != "Meta Investor") %>% 
+  select(financial_sector, isin, company_name, market_value, value_usd) %>% 
+  group_by(financial_sector, isin, company_name) %>% 
+  summarise(sum_value_usd = round(sum(value_usd, na.rm=T))) %>% 
+  ungroup() %>% 
+  mutate(share_total_value = round(
+    sum_value_usd/(total_overview %>% filter(asset_type=="Equity") %>% pull(sum_value_usd)),6)) %>% 
+  arrange(financial_sector, desc(share_total_value))
+datatable(no_ALD_EQ_detail)
```
Clare2D, line 259: This is great! I hope it is actually used, but this generally makes lots of sense!

wip_pacta_data_qa.Rmd
```diff
@@ -0,0 +1,545 @@
...
+```
+
+## Non-PACTA sector with ALD
+
+### overview Equity holdings, non-PACTA sector but with ALD
+Overview of Holdings in non-PACTA sectors for which we do find ALD.
+
+1. Overview on sector level, including count and share of overall equity value affected
+2. Overview on sector-company level, including overall equity value affected
+
```
Clare2D, line 315: I think this also requires an additional sentence or two about what this is/means

wip_pacta_data_qa.Rmd
```diff
@@ -0,0 +1,545 @@
...
+  arrange(sectors_with_assets)
+datatable(non_pacta_ALD_EQ)
+
+non_pacta_ALD_EQ_detail <- result_data_qa %>% 
+  filter(asset_type == "Equity" &
+           has_asset_level_data == T &
+           financial_sector == "Other" &
+           investor_name != "Meta Investor") %>% 
+  select(sectors_with_assets, isin, company_name, market_value, value_usd) %>% 
+  group_by(sectors_with_assets, isin, company_name) %>%
```
Clare2D, line 342: If you add in the bloomberg_id/corporate_bond_ticker it will make it easier to make the changes to the fin_sector_overrides directly!

wip_pacta_data_qa.Rmd
```diff
@@ -0,0 +1,545 @@
...
+
+1. List of fund ISINs and their Values impact on the overall PF value, filtered by direct_holding == FALSE and asset_type == "Funds"
+2. List of fund_isins that have ISIN == "MissingValue" and are therefore omitted by the first table. This, however, seems to come from the normalise_fund_data() function and may not be necessary. It is introduced on purpose for fund holdings, whose overall sum weight is smaller than 1. 
+
+**Actions:**
+
+* Check if the funds with higher impact are available in the fund_data file.
+* If they are not, try to get the corresponding fund data from LIM and add the data to the fund_data table before the next run.
+
+
```
Clare2D, line 411: Cool. We'll have to check this when there is some funds data

wip_pacta_data_qa.Rmd
```diff
@@ -0,0 +1,545 @@
...
+```
+
+
+## Value by asset type
+**OPEN:**
+
+Target and Actions unclear
+One idea seems to be to check to share of sovereign bonds...
```
Clare2D, line 454: Yep, I believe this is mainly about S.bonds. Laura has an additional analysis about these. And sometimes they need to be identified as a separate asset type.

wip_pacta_data_qa.Rmd
```diff
@@ -0,0 +1,545 @@
...
+## Port Weight check
+This table checks if the portfolio weights of the data inputs and outputs correspond.
+We expect that this is the case, i.e. the difference in port_weight should be negligible.
+
+1. Check if equity portfolio inputs match the outputs in terms of weighted value per company
+2. Check if corporate bond portfolio inputs match the outputs in terms of weighted value per company
+
+**Actions:**
+
+* If this is not the case, dig into what could have lead to changes of portfolio weights during the analysis.
```
Clare2D, line 474: What are the potential reasons here for this to be different 😕. I think this has caused a bit of confusion in the past, as sometimes differences are acceptable, but it's been flagged. So we should possibly question this one again.

wip_pacta_data_qa.Rmd
jacobvjk: Yea, I'm just a bit unsure what to use for the missing currency table then, since this is where I need some info about a value in the original currency... maybe leave it up for discussion on Monday?

wip_pacta_data_qa.Rmd
jacobvjk: sum_value_usd.x is only a duplicate, if the isin is only held by one investor in one portfolio. otherwise it sums across all these.. grouping by isin. But I will definitely improve the column names.

jacobvjk: I will also remove the sum_value_usd_total, as I don't see the value in repeating the same value in each row. just needed for the calculation.

wip_pacta_data_qa.Rmd
jacobvjk: Yea, I'll see how easy/hard it is to have a single/multiple choice type filter in DT::datatable. Not quite sure about it. Might already be on the verge of becoming a small app then.
Would be really cool though

wip_pacta_data_qa.Rmd
jacobvjk: looks like I grouped by fund_isin (which is NA for all these) instead of by isin. Not sure why I did that. Will check with Laura on Monday if this was how we did that and if so why. But for the time being, will change to isin level. Seems more reasonable indeed.

wip_pacta_data_qa.Rmd
jacobvjk: Yea, I found that surprising as well. Let's see what the discussion on Monday brings

jacobvjk:
I will, of course, improve the exact placement of text and tables within sections. Right now its always all the text first and then some poorly explained tables. this is just the working state. Same goes for the naming of the columns and table header.

jacobvjk:
@Clare2D this PR is outdated, can be rejected and closed

PR #48, opened 2020-07-23
Bug temp remove duplicate isins

jacobvjk:
This is a bit of a hacky solution for now, but I thought I just add the csv temporarily and read from there which isins to remove, until this is solved on the DB side.
@Clare2D


1_portfolio_check_initialisation.R
```diff
@@ -30,7 +30,7 @@ source("0_portfolio_input_check_functions.R")
 source("0_global_functions.R")
 source("0_sda_approach.R")
 
-project_name <- "Laura_test_project_2019_jk"
+project_name <- "liechtenstein_p2020"
```
jdhoffa, line 33: While we're here, I propose we extract the "project_name" to a config file. 
This way we won't have merge conflicts every time somebody is working on a different project

non_distinct_isins.csv
```diff
@@ -0,0 +1,237 @@
+company_id,bloomberg_id,company_name,country_of_domicile,ticker,corporate_bond_ticker,figi,isin,cusip,sedol1,asset_type,security_icb_subsector,security_bics_subgroup,security_bclass4,security_mapped_sector,cfi_code,security_type,maturity_date,issue_date,amount_issued,amount_out,coupon_value,par_price,par_amount,current_face_amount,accrued_interest,security_amount_out_native,unit_share_price,position_market,market_value,current_shares_outstanding,current_shares_outstanding_all_classes,free_float_percent,free_float_shares,exchange_rate_usd,moody_rating,s_and_p_rating,fitch_rating,financial_timestamp
```
jdhoffa, line 1: This repo is public, and this financial data seems to me that it should be private/ not committed to version control lol... 
@Clare2D thoughts?

1_portfolio_check_initialisation.R
Clare2D: That sounds like a good plan. The project_name and the file paths below would then be in a config file in the repo I guess.. Can you implement this?

non_distinct_isins.csv
jacobvjk: I totally agree, I just don't see how financial data can be entirely kept out of this as long as it is not solved on DataStore.. Of course, I can reduce the list to only ISINs, but I'm not sure how much better that is. Open to suggestions here

Clare2D: ISIN only makes sense here as a better alternative!

1_portfolio_check_initialisation.R
Clare2D: Moved to an issue.

non_distinct_isins.csv
jdhoffa: Hmm. If you could put the dataset on dropbox somewhere (maybe in the project directory?). You could:
* .gitignore this file, so as to not commit it to version control
* access the file using `rdrop2` and an oauth token (see #50)

There would just need to be a sensible place on dropbox to upload this file, @Clare2D thoughts?

Clare2D: This is a temporary solution only and needs to be used by everyone using the code, so we can't ignore it. Maybe as a longer solution we can start to think about the rdrop2, but I don't think now is the best time to start on that.. but eventually that could work for a few different things.

jdhoffa: Ah understood. I thought this was "project specific", didn't realize everyone would need the fix. 
I'll leave it to you to decide then!

jacobvjk:
@jdhoffa the conflict should be resolved now

jdhoffa:
Merged
@jacobvjk your safe to delete that branch

non_distinct_isins.csv
Clare2D: Ahh the actual solution to this should be that this fix happens in data_preparation not in PACTA_analysis. Let's get to the solution with Taylor asap and see when we can write the test into data prep and remove this solution!

Issue #49, opened 2020-07-23
Move Project name and file paths to yml file and ignore this file


Clare2D:
While we're here, I propose we extract the "project_name" to a config file. 
This way we won't have merge conflicts every time somebody is working on a different project

_Originally posted by @jdhoffa in https://github.com/2DegreesInvesting/PACTA_analysis/pull/48#discussion_r459296636_

Clare2D:
Jacob are you able to do this? Both the project_name and the true/false on external/internal and the file paths. 
Thanks!

jacobvjk:
I think I can figure that out, will add it to the agenda

jdhoffa:
Closed by #51

PR #51, opened 2020-07-24
New config for project paths

jacobvjk:
create new config file project_settings.yml within this repo

read:

- project_name
- twodii_internal
- project_location_ext
- data_location_ext

from this config file rather than hard coding them within the R code

Every PACTA user will have to specify the entries of this file prior to running the PACTA_analysis code in the future
The variable names and their content have been kept equal to what they were within the code

@jdhoffa @Clare2D 
any wishes and/or comments?


0_portfolio_input_check_functions.R
```diff
@@ -857,7 +857,7 @@ get_and_clean_fin_data <- function(fund_data){
   rm_duplicates <- rm_duplicates %>% distinct(isin) %>% pull(isin)
   fin_data_raw <- fin_data_raw %>%
     filter(!(isin %in% rm_duplicates))
-  
+
```
jdhoffa, line 860: not sure I agree with this

1_portfolio_check_initialisation.R
```diff
@@ -30,16 +30,31 @@ source("0_portfolio_input_check_functions.R")
...
+# the project_name will determine the name of the folder that is created for this project
+# Set twodii_internal to TRUE to run the analysis on an internal 2dii laptop
+# This setting uses the dropbox connection for data import
+# Set twodii_internal to FALSE, tu use external data locations
+# Specify these data locations in the config file "project_settings.yml" in the repo
+
+cfg <- config::get(file = "project_settings.yml")
+stopifnot(exists("cfg")  == T)
+stopifnot(cfg %>% class() == "list")
+stopifnot(cfg %>% length() == 2)
```
jdhoffa, line 47: this could be a function, something like: 

``` r 
check_valid_cfg <- function(cfg){
  stopifnot(exists("cfg")  == T)
  stopifnot(cfg %>% class() == "list")
  stopifnot(cfg %>% length() == 2)
  
  invisible(cfg)
}
``` 

would make it easier to add more checks in the future, and keep them organized

jacobvjk: sure, will add this

jacobvjk: changes pushed, pls review
@jdhoffa @Clare2D

jacobvjk:
Nothing further to add on this one from my side

jdhoffa:
Closes #51

jdhoffa:
Closes #49

Clare2D:
Hi guys, given that this is a fairly big change to how people use this code, it would be good to flag this to the pacta team that this change has happened so that people are aware how to use it. The "read me" also needs to be updated as this is a code that has been used by externals in the past and should be kept up to date for them too.

jdhoffa:
@Clare2D how would you feel about this type of change being documented on the [2dii blog](https://2degreesinvesting.github.io/), and then circulated to the PACTA team? 

I tend to prefer this approach, over the "notify people one by one over slack" because it keeps a track-record of all changes to the PACTA interface (a kind or "release notes")

Clare2D:
That sounds fine to me! As long as this happens before people try to use it and get confused!

PR #52, opened 2020-07-27
Input data qa

jacobvjk:
rebased with master and renamed the markdown file


jacobvjk:
this should be added to the other PR, just to make everything a bit cleaner

jacobvjk:
Or in case of conflict, we might just reject the old #47 PR and and use this one here instead @Clare2D (the old one will probably conflict with the new config file, so this might be an easier cleanup)

Clare2D:
Will you let me know when you're ready for this to merge? Or is it already do you think? Happy for the other PR to be deleted

jacobvjk:
I'll let you know, hopefully today

jacobvjk:
This state will now automatically export the pacta_data_qa.html file to the processed_inputs folder in the project directory AND should not care about there being only one of the main asset types (EQ or CB).
I will clearly not win a style comp with this (neither coding nor formatting), but I think it the markdown can already be helpful in this state.
I would be happy to merge this @Clare2D (take a look if you feel like it @jdhoffa )

Clare2D:
Thanks Jacob

PR #54, opened 2020-07-27
Update readme

jdhoffa:
I did a couple of things while I was here: 
1. "README.md" now includes correct instructions (and a few typo fixes)
2. project_settings.yml has some sensible defaults. I would propose that we avoid pushing changes to that files onto master (ever) so that we dont needlessly edit it a million times for different projects.  
3. README is now generated from a `.Rmd` instead of an `.md`. This makes it easier to include and evaluate R code chunks, if we decide we want to down the road. (Just edit the .Rmd instead and click knit)


jdhoffa:
Closes #54

project_settings.yml
```diff
@@ -1,8 +1,8 @@
 default:
 
-  project_name: new_config_test
+  project_name: name_of_the_project_folder
 
   project_internal:
     twodii_internal: FALSE
```
jacobvjk, line 6: I think this should be TRUE as a default @jdhoffa

jacobvjk: although I am seeing it used to be FALSE.. it's the same to me really. Rest looks good to me

README.Rmd
```diff
@@ -0,0 +1,184 @@
...
+  project_internal:
+    ...
+    data_location_ext: ~/path/to/data
+```
+
+   - Please define ```data_location_ext``` by copying the full path to the folder, in which you stored the data files (set of rda files) provided Asset Resolution or the 2DII.
+
+   - Please consider: 
+     - copying in a path from windows requires to change all "\" to "/". 
+     - path needs to be a string ("...")
```
jacobvjk, line 124: also pretty sure it automatically turns it into a string, although I don't know if that is true for all setups of course

project_settings.yml
jdhoffa: I went back and forth on this one haha... 
`TRUE` is a better default when we are using it, but `FALSE` is a better default for when externals (EIOPA etc) are using it, and i tend to feel like we should make it easier for them...

README.Rmd
jdhoffa: you're right! I left that in from the old README, removing it

project_settings.yml
jdhoffa: @Clare2D thoughts?

jdhoffa:
Ok, I think the changes here are relatively small. im gonna go ahead and merge, feel free to revert if you think it should change!

PR #59, opened 2020-07-30
Pacta analysis sfc updates

jacobvjk:
just opening a conversation on the bug fix regarding the aggregation of company pf weights


0_reporting_functions.R
```diff
@@ -345,6 +344,14 @@ ReportGeneration <- function(){
...
   }
   
+  if(!IncCoalRetirement){
+    text <- removetextlines("CoalCapChart")
+  }
+  
+  if(!IncRenewableAddition){
+    text <- removetextlines("RenewableCapChart")
+  }
+
```
jacobvjk, line 354: @jdhoffa @Clare2D shopuld we care about these being added to ReportGeneration()? My intuition says no, since we're about to get rid of it anyways...?

0_reporting_functions.R
```diff
@@ -377,27 +384,28 @@ ReportGeneration <- function(){
...
-  text$text <- gsub("start_year+5",start_year+5,text$text,fixed=T)
-  text$text <- gsub("start_year",start_year,text$text)
-  text$text <- gsub("investor_name_select",investor_name_selectClean,text$text)
-  text$text <- gsub("portfolio_name_select",portfolio_name_selectClean,text$text)
+  text$text <- gsub("Startyear+10",start_year+10,text$text,fixed=T)
+  text$text <- gsub("Startyear+5",start_year+5,text$text,fixed=T)
+  text$text <- gsub("Startyear+5",start_year+5,text$text,fixed=T)
+  text$text <- gsub("Startyear",start_year,text$text)
+  text$text <- gsub("InvestorName",investor_name_selectClean,text$text)
+  text$text <- gsub("PortfolioName",portfolio_name_selectClean,text$text)
```
jacobvjk, line 392: this looks a bit like the old naming convention is reintroduced? should that be changed?

0_reporting_functions.R
```diff
@@ -377,27 +384,28 @@ ReportGeneration <- function(){
...
-  text$text <- gsub("scenarioText",reportdata$scenarioText,text$text)
-  text$text <- gsub("scenarioName",reportdata$scenarioName,text$text)
-  text$text <- gsub("scenarioValue",reportdata$scenarioValue,text$text)
-  text$text <- gsub("scenarioDescription",reportdata$scenarioDescription,text$text)
-  text$text <- gsub("scenarioTemp",reportdata$scenarioTemp,text$text)
+  text$text <- gsub("ScenarioText",reportdata$scenarioText,text$text)
+  text$text <- gsub("ScenarioName",reportdata$scenarioName,text$text)
+  text$text <- gsub("ScenarioValue",reportdata$scenarioValue,text$text)
+  text$text <- gsub("ScenarioDescription",reportdata$scenarioDescription,text$text)
+  text$text <- gsub("ScenarioTemp",reportdata$scenarioTemp,text$text)
```
jacobvjk, line 403: not sure which naming conventions these are, but still worth a look

0_reporting_functions.R
```diff
@@ -431,7 +439,7 @@ ReportGeneration <- function(){
     
     ReportName <- paste0("ClimateAlignmentReport_",strtrim(investor_name_select,20),"_",strtrim(portfolio_name_select,20))
     
-    # if (InvestorType %in% c("InvestorSingle","InvestorMeta")){ReportName <- paste0("ClimateAlignmentReport_",investor_name_select)}
+    if (investor_type %in% c("InvestorSingle","InvestorMeta")){ReportName <- paste0("ClimateAlignmentReport_",investor_name_select)}
```
jacobvjk, line 443: this feels like it should be validated

3_run_analysis.R
```diff
@@ -32,47 +33,57 @@ if(file.exists(equity_input_file)){
     
     port_eq <- calculate_weights(port_raw_eq, "Equity", grouping_variables)
     
-    port_eq <- port_eq %>% filter(port_weight > 1e-6)
```
jacobvjk, line 35: any reason why this should be removed?

jdhoffa: nope, can add this back

4_report_maker.R
```diff
@@ -42,15 +39,11 @@ portfolio_overview$portfolio_name <- clean_punctuation(portfolio_overview$portfo
 report_list <- get_report_list(portfolio_overview)
 
 
-#template <- read_utf8_tex(paste0(getwd(),"/Templates/",templateversion,".tex"))
+template <- readLines(paste0(dirname("/Templates/",templateversion,".tex"),encoding = "UTF-8"))
```
jacobvjk, line 42: this crashes the report for me, not sure why and what it's for

jdhoffa: the `read_utf8_text()` line or the `readLines`?
This functin loads up the "default" template tex file that is used in report generation. The biggest hurdle we had with getting report printing to work on multiple computers was file encoding, so we forced the encoding to be UTF-8 in hopes that that would fix things.

0_reporting_functions.R
jdhoffa: Nay, unfortunately the old naming convention is still present in the Template .tex files (afaik). 
Maybe open an issue to align the Template files to the new convention? and then we can change these back.

0_reporting_functions.R
jdhoffa: see above

jdhoffa: Here for example: https://github.com/2DegreesInvesting/Templates/blob/77db076c87b5a136854d8a1a3d417ded1a3215fa/GeneralTemplateInput_v6_SB_en.tex#L255

0_reporting_functions.R
jdhoffa: Here for example: https://github.com/2DegreesInvesting/Templates/blob/77db076c87b5a136854d8a1a3d417ded1a3215fa/GeneralTemplateInput_v6_SB_en.tex#L255

Actually not sure if it's worth fixing, because I think we need to be moving away from this type of Report generation altogether, and towards bookdown (or gitbook as people erroneously call it hehe)

jacobvjk: I mean, as long as we have two computers that still have a functioning version of the Report generation.. maybe we can decide this together with @Clare2D on Monday

4_report_maker.R
jacobvjk: the readLines does not work.. one issue seems to be that dirnames expects one input, which can be fixed by wrapping the three inputs in another paste0(), but even then, I won't find the corresponding directory

jacobvjk:
There are definitely still some problem in the graph and report generation.
`"Equity Market Filter removed All Data"` Still occurring and removing data in graph generation.. 
`Error in ReportGeneration() : object 'IncRenewableAddition' not found` this is new to me..

0_reporting_functions.R
Clare2D: Yep as Jackson suggests, I don't think this is worth changing. If we want to use a different template, we'd need to go through all templates to fix this. Hopefully, we move on soon!

0_reporting_functions.R
Clare2D: does this work? The investor_type is defined in one of the first functions in 4_report_maker so if you've changed it there then that's ok. Otherwise, this line is about naming the output files, so it's not so important.

Clare2D:
> 
> 
> There are definitely still some problem in the graph and report generation.
> `"Equity Market Filter removed All Data"` Still occurring and removing data in graph generation..
> `Error in ReportGeneration() : object 'IncRenewableAddition' not found` this is new to me..

That IncRenewableAddition is about whether we have the "Renewable Additions" chart included and should be defined within the parameter file. At some points this chart hasn't worked, so this excluded this chart

0_reporting_functions.R
Clare2D: Not sure I get the question - but this just removes the section for the "Renewable Additions chart" that you mention below as well. This can be removed if you're printing the chart, or if you remove this section manually from the tex file

Clare2D:
PS. Good luck embarking on this! As we renamed things in the rest of the code, the cleaning and renaming of the graphing code was something I considered too much work! I think there are many small gains to be made though.

jacobvjk:
We might also want to consider using the same startyear configuration for the analysis and the report. I just got a bunch of errors for that not being in sync. I am guessing it should always be the same though? @Clare2D

Clare2D:
Yes I think this would be a good thing to implement. Also including checks to see that market files and indicies files are also aligned on this.

jacobvjk:
It's funny you would mention that, as misaligned market files are exactly the next problem I just ran into..

Clare2D:
haha. Idealy you set it up, so it just reruns the market files with the correct start date....

jacobvjk:
I just ran this on the Liechtenstein data, and set up a new test project for that. weirdly, the outcomes look quite different in terms of units (I checked a 5years chart). not sure where that comes from. I will have to focus on EIOPA in the afternoon, but this feels a bit shaky and will have to be revisited

jacobvjk:
Before I continue with this it would be great to disentangle a bit, what's actually supposed to happen here, and ideally break it down into several issues. I have found problems with the following aspects here so far:
-  Graphs generation
-  Report generation
-  The company weighting for bonds does not seem to be fixed satisfactory
-  Unified start year across configs is desirable
-  Global/GlobalMarket filter issue
-  Market and peer references? seems to produce issues frequently, where are the currently correct indices?
-  Units of results shifting substantially, when running this on the same project that I ran the current master version on.

maurolepore:
What's the status of this PR? Should we review/close?

jacobvjk:
@Clare2D ?

Clare2D:
Yep feel the same. The reporting errors shouldn't be the same issues moving forward either.

PR #60, opened 2020-08-04
Fix bics sector bug

jacobvjk:
@Clare2D @jdhoffa there was a bug in the create_audit_file() function. It still contained "bics_sector" instead of "securit_mapped_sector" (which it seems to me is the new setup?). When external users merged master into their forks, this breaks the analysis. Would be good to review this asap, so I can get back to the Norwegian Financial Supervisors and inform them to update their fork again


Clare2D:
The only thing to note is that we do need bics_sectors at some point in order to use the gitbooks

jacobvjk:
good to know.. I'll make an issue to remind us of that

PR #64, opened 2020-08-06
Web tool: update parameter and docker settings, add logging, updated path logic

jacobvjk:
Bit unsure if the parameter files should be copied into these directories in the repo as well...


jacobvjk:
one could argue btw that saving the processed input files does not require the project name as that is always "working dir"
`save_if_exists(portfolio_total, portfolio_name_, paste0(proc_input_path, "/", project_name,"_total_portfolio.rda"))`

jacobvjk:
@Clare2D I added a bunch of messages to write to the log. Some of them ended up in `0_portfolio_input_check_functions.R` which, now that I reflect on it, might not be ideal in case we merge in new functionalities from master.. problem is, they cannot be moved right into the web script 1 in case they reflect an error raised in `0_portfolio_input_check_functions.R`, which would stop the script before logging anything.. tbd I guess.
Maybe it would also be wise to make some test portfolios that will raise these error messages?

Clare2D:
> 
> 
> one could argue btw that saving the processed input files does not require the project name as that is always "working dir"
> `save_if_exists(portfolio_total, portfolio_name_, paste0(proc_input_path, "/", project_name,"_total_portfolio.rda"))`

This makes sense and I would support this change!

jacobvjk:
The latest commit saves every portfolio that is passed to the tool in a separate folder in 30_proc_inp and (web_tool_script_1.R) and in the analysis (web_tool_script_2.R) it loops over all given portfolios, reading and analysing the proc input folders one at a time, and writes each result into a separate folder per portfolio name.
@Clare2D

Clare2D:
Let's discuss however I think the code breaking is a good option.

jacobvjk:
So just a short summary of what this PR does:
- New function write_log() writes error messages to the 00_Log_Files folder
- Tests for such messages are added, some are part of the standard input check, some are unique to the webtool script
- the folder structure is as follows: the 6 standard project folders are written into the working_dir folder, which in turn is on the repo level
- the script can process one or multiple portfolios, depending on the input file, but it will create a sub-folder for each portfolio_name. The slicing of portfolios happens on Constructiva's side.
- if the script is called from the web, it will take the command line input as the portfolio name and look for a corresponding file as input.
- the script can be run in a docker container, for which the image is provided by Constructiva. All other changes outlined above also work in this setting

0_web_functions.R
```diff
@@ -187,6 +204,11 @@ read_web_input_file <- function(input_file_path){
   
   if(data_check(input_file) == FALSE){
     warning("Input file not readable")
+    ifelse(nrow(input_file) == 0,
+           write_log(msg = "Input file has 0 rows. Please ensure the uploaded file is not empty."),
+           write_log(msg = "Input file could not be transformed into a data.frame.
```
Clare2D, line 209: These are two different error messages - is the second meant to be the "stop"?

parameter_files/WebParameters_2dii.yml
```diff
@@ -1,7 +1,7 @@
 default:
     paths:
-        project_location_ext: C:/Users/clare/Desktop/TestGit/PACTA_analysis/
```
Clare2D, line 3: We should probably make this a GitIgnore file? Or is there an option to keep a "template" version of this and changes are ignored?

web_tool_script_2.R
```diff
@@ -45,161 +45,185 @@ analysis_inputs_path <- set_analysis_inputs_path(twodii_internal, data_location_
...
-    port_own_eq <- aggregate_portfolio(company_own_eq) 
+    list_investors_eq <- unique(port_raw_all_eq$investor_name)
     
-    # Create combined outputs
-    company_all_eq <- bind_rows(company_pw_eq, company_own_eq)
-    
-    port_all_eq <- bind_rows(port_pw_eq, port_own_eq) 
-    
-    if(has_map){
+    for (e in 1:length(list_investors_eq) ){
```
Clare2D, line 75: Should this not be "portfolio_name"? and list_portfolios_eq ?

0_web_functions.R
jacobvjk: ehm, no. interesting case. Will have a look

jacobvjk: Ah, I remember. the messages are two diff arguments to the ifelse() function.
If data check is FALSE AND the input file has zero rows, you get the first message, If data check is FALSE AND the input file does not have exactly nrows == 0, the second message highlights that the file could not even be transformed into a data frame. within data_check() that reflects is.data.frame == F
So this is expected

parameter_files/WebParameters_2dii.yml
jacobvjk: I've been wondering how best to that kind of thing for a while now. Have been talking to Jackson recently, be also did not know. I mean, we can add it to gitignore, but then we will have to assume that everyone has this file installed... I guess we could stop tracking it and see what happens when the next person pulls code from the repo...

web_tool_script_2.R
jacobvjk: I don't think that's necessary. I wrapped a whole other loop around the calculations that goes through the file names, which are, at east in this case, the portfolios. I left this ne in here, just in case there actually is more than one investor in a file. But I think it could just as well be removed

jacobvjk:
I added the break after finding missing cols. I think it might be reasonable also to break in case columns have wrong classes, but maybe I am being too restrictive here.

parameter_files/WebParameters_2dii.yml
jacobvjk: @Clare2D since the discussion on this topic does not seem overly promising neither on stackoverflow nor on our slack channel, we could just go for the same logic as in the normal Pacta code and copy the parameterfiles into the 10_Parameter_File folder... that would at least be consistent

Clare2D:
As discussed:
- Move the ald_scen_eq & _cb to outside of the "file" loop. 
- Changing in "identify_portfolios" the warning on multiple investors to a "stop"
- Overall structure of the file_names file and whether how it's being created makes sense.

Clare2D:
Am I right to merge this then?

jacobvjk:
I still need to verify with Wim if he passes me an identifier in the portfolio_name column or not. Because if not we would currently run separate analyses for each pf in a grouped pf. I think an alternative could be to simply collapse the different pf names into one long string in such a case within R... that would circumvent the problem, but is kind of ugly.

0_portfolio_test.R
```diff
@@ -208,7 +208,7 @@ aggregate_map_data <- function(portfolio){
   
   portfolio <- portfolio %>% 
     ungroup() %>% 
-    group_by(!!!rlang::syms(grouping_variables), allocation, 
+    group_by(!!!rlang::syms(grouping_variables), #allocation,
```
Clare2D, line 211: So this change doesn't make sense to me - Was this causing a problem? Because I am fairly certain that we need to also group by allocation here. This is just for the mapping data..

Clare2D (commented):
Just the one small thing with the map function.

PR #66, opened 2020-08-07
fund work

vintented:


Clare2D:
Hey Vincent, just checking a few things - are both versions of the fund_analysis_functions used? If not, can you delete one of them? I've changed the name of the 4_fund_analysis to 5_ to avoid confusion with 4_report_maker. I'm not so sure that the 3_xxx makes logical sense here either. 

This all being said, I'm not sure that PACTA_analysis is even the best place for this file to live. It seems that the majority of this is retail work right? In which case it would be more logical to source the PACTA_analysis functions into a different repository and run from there. Or is there something I'm missing here?

vintented:
I am keeping both versions for now because I do not want to mess with the original one which contains Klaus's code. Yes, I agree. Mainly retail stuff and actually it does not rely on PACTA_analysis functions so it could in theory live anywhere. @2diiKlaus asked me to move over the work to PACTA_analysis, so it is ultimately his call. I am happy to move things elsewhere.

Clare2D:
Ok makes sense.

0_fund_analysis_functions_v2.R
```diff
@@ -0,0 +1,1210 @@
...
+  # weight value 
+  .data <- .data %>% 
+    dplyr::group_by(!!!rlang::syms(id_cols)) %>% 
+    dplyr::summarise(weighted_value := stats::weighted.mean(.data[[values_from]], .data[[weights_from]])) %>% 
+    dplyr::ungroup()
+  # rename output
+  if (!is.null(values_to)) {
+    .data %>% 
+      plyr::rename(c("weighted_value" = values_to))
+  } else if (is.null(values_to)) {
```
maurolepore, line 497: In this case it seems enough to use `} else {`  instead of `} else if (condition) {`

0_fund_analysis_functions_v2.R
```diff
@@ -0,0 +1,1210 @@
...
+  # check parameters 
+  lapply(character_parameters, function(x) {stopifnot(is.character(x) & !is.null(x))})
+  lapply(character_null_parameters, function(x) {stopifnot(is.character(x) | is.null(x))})
+  # check names 
+  crucial_names <- c(values_from, numerator_group, denominator_group, id_cols)
+  r2dii.utils::check_crucial_names(.data, crucial_names)
+  # check values
+  assertr::verify(.data, class(.data[[values_from]]) == "numeric")
+  # return 
+  return(.data)
```
maurolepore, line 1170: This function is called primarily for its side effect. Return `invisible(.data)` instead of `return(.data)`.
https://principles.tidyverse.org/out-invisible.html

0_fund_analysis_functions_v2.R
```diff
@@ -0,0 +1,1210 @@
...
+}
+
+summarise_group_share <- function(
+  .data,
+  id_cols = NULL,
+  numerator_group = "asset_type", 
+  denominator_group = "portfolio_name",
+  values_from = "value_usd",
+  values_to = NULL,
```
maurolepore, line 1179: I see all these arguments have defaults. Are they "detail" arguments?
https://principles.tidyverse.org/args-data-details.html

0_fund_analysis_functions_v2.R
```diff
@@ -0,0 +1,1210 @@
...
+  ...
```
maurolepore, line 1180: The ellipsis argument (`...`) should come between the "data" arguments and the "details" argument.
https://principles.tidyverse.org/dots-position.html#dots-position

0_fund_analysis_functions_v2.R
```diff
@@ -0,0 +1,1210 @@
...
+    dplyr::group_by(!!!rlang::syms(c(group_vars, numerator_group))) %>% 
+    dplyr::summarise(group_share = sum(.data[[values_from]], ...)) %>% 
+    group_by(!!!rlang::syms(group_vars)) %>% 
+    dplyr::mutate(group_share = .data$group_share / sum(.data$group_share, ...)) %>% 
+    dplyr::ungroup()
+  # rename output
+  if (!is.null(values_to)) {
+    .data %>% 
+      dplyr::rename("{values_to}" := .data$group_share)
+  } else if (is.null(values_to)) {
```
maurolepore, line 1207: Here, `} else {` is enough.

0_fund_analysis_functions_v2.R
```diff
@@ -0,0 +1,1210 @@
...
+    return(.data)
```
maurolepore, line 1208: Reserve `return()` for returning early.
See 3.3 at https://style.tidyverse.org/functions.html

0_fund_analysis_functions_v2.R
```diff
@@ -0,0 +1,1210 @@
...
+  # check parameters 
+  stopifnot(is.character(id_cols))
+  stopifnot(is.character(weights_from))
+  stopifnot(is.character(values_from))
+  stopifnot(is.character(values_to) | is.null(values_to))
+  # check class 
+  data %>% 
+    assertr::verify(class(.data[[weights_from]]) == "numeric") %>% 
+    assertr::verify(class(.data[[values_from]]) == "numeric")
+}
```
maurolepore, line 469: This function is called for primarily for its side effects. Return `invisible(data)`.
https://principles.tidyverse.org/out-invisible.html

PR #67, opened 2020-08-11
Error logging

jacobvjk:
Add log function to global_functions and fix the bug with Global/GlobalMarket equity market filter. The logging function will be used more extensively in the web tool as it can be used to return somewhat readable error messages to the user/uploader of a portfolio file.
@Clare2D


jacobvjk:
@Clare2D are you fine with these changes?

0_portfolio_input_check_functions.R
```diff
@@ -64,13 +64,39 @@ clean_portfolio_col_types <- function(portfolio, grouping_variables){
   
   portfolio$currency <- if_else(portfolio$currency == "Euro","EUR",portfolio$currency)
   
+  if(is.character(portfolio$investor_name) == FALSE) {
```
Clare2D, line 67: What will happen if one of these columns is not in the file? Or is this check previously?

0_global_functions.R
```diff
@@ -315,4 +315,15 @@ check_valid_cfg <- function(cfg){
   stopifnot(cfg$project_internal$twodii_internal %>% is.logical() == T)
   
   invisible(cfg)
+}
+
+
+#write error log for input portfolio - msg should be a string containing the error message
+write_log <- function(msg, ...) {
```
Clare2D, line 322: Question perhaps whether these should be labelled "error_messages" or more a Warning message. But this isn't too critical.

0_portfolio_test.R
```diff
@@ -9,6 +9,12 @@ get_ald_scen <- function(portfolio_type){
     
     ald <- read_rds(paste0(analysis_inputs_path, "/equity_ald_scenario.rda"))
     
+    # fix naming issue. if equity market list is either Global or GlobalMarket, allow both
+    # values. Else retain the input from the config file
+    equity_market_list <- ifelse(c("Global", "GlobalMarket") %in% equity_market_list,
```
Clare2D, line 14: Does this function work in that, if one of Global or GlobalMarket is in the list then they are both added? Or does it work that they both have to be in the list?

jacobvjk: if either one is in the list, both are added, so that it doesn't matter anymore which one to choose. Multiple people ran into issues with this over the last few weeks (that's also what caused the problem at the Norwegian financial supervision btw) and this has solved it for me every time.

0_portfolio_input_check_functions.R
jacobvjk: That is being checked in the function before this one (check_missing_cols() I believe). So theoretically it should not happen

PR #70, opened 2020-08-18
Input data qa

jacobvjk:
@Clare2D here's the updated qa PR. this should also easily transition into master after merging into qa


Clare2D:
Just one thing - we should not be uploading the results files to GitHub as this is a public repo - are you able to remove these and add the correct line to gitignore? Thanks!

jacobvjk:
sure @Clare2D the results somehow and weirdly have sneaked in from the current_web_functionality... I just removed them, as these folders are not required in the normal pacta case and will not be generated by running the script anyway

Clare2D:
Great. I'll merge now then.

PR #73, opened 2020-08-26
Align script 3

jacobvjk:
@Clare2D 
- Propagate bics sector through portfolio files (for interactive report)
- Note: I added this to the create_ald_flag() function in 0_portfolio_input_check_functions.R --> This is absolutely not optimal, unless we also want this change in the normal pacta code. Otherwise it will be disappearing with the next merge of master into current_web_functionality.. I'm happy to change this, if we don't want that. It just seemed like the easiest place to do this.
- changed the output path of the interactive report so it sticks to the web_tool conventions. This means it copies all necessary files from the create_interactive_report repo to the output directory. this will allow us to open the output file in the desired format.
- updated pacta-data to 2020 inputs and changed the debt_fin_data file to cover bics_sectors.
- minor beautification of output paths

**OPEN:** the integration should be discussed with constructiva. They would need this repo as well.
**OPEN:** the new 2020 pacta-data also needs to go to wherever Constructiva makes the tool read this data from


0_portfolio_input_check_functions.R
```diff
@@ -706,10 +704,9 @@ create_ald_flag <- function(portfolio, comp_fin_data, debt_fin_data){
   if (data_check(portfolio_other)){
     portfolio_other <- portfolio_other %>% mutate(has_asset_level_data = NA,
                                                   sectors_with_assets = NA,
-                                                  has_ald_in_fin_sector = NA,
-                                                  bics_sector = NA)
+                                                  has_ald_in_fin_sector = NA)
   }else{
-    portfolio_other <- portfolio_other %>% add_column("has_asset_level_data","sectors_with_assets","has_ald_in_fin_sector", "bics_sector")
+    portfolio_other <- portfolio_other %>% add_column("has_asset_level_data","sectors_with_assets","has_ald_in_fin_sector")
```
Clare2D: I like the simplicity of this function! Can't get clearer as to intention on this!

0_portfolio_input_check_functions.R
```diff
@@ -1514,4 +1511,27 @@ add_other_to_sector_classifications <- function(audit){
...
   
   audit
-}
\ No newline at end of file
+}
+
+
+add_bics_sector <- function(portfolio, comp_fin_data, debt_fin_data){
+  #join in bics sectors for EQ and CB
+  portfolio_eq <- portfolio %>% filter(asset_type == "Equity") %>%
```
Clare2D, line 1519: Will this still work if the filter reduces the portfolio to 0 lines?

jacobvjk: Yes, tested it by filtering out all equity holdings before running this. Goes through without complaints and will let you know that there's no EQ in portfolio when trying to create the portfolio subset in the next step. As before.

PR #74, opened 2020-08-26
re-introduce bics_sectors as needed in bookdown

jacobvjk:
Add the bics sector via comp_fin_data and debt_fin_data to the portfolio. this keeps the code consistent with the web tool and is required for the bookdown report.
@Clare2D


Clare2D:
Looks great

PR #76, opened 2020-09-07
Report with eq or cb

jacobvjk:
IMPORTANT:
Please ideally first review and merge this: https://github.com/2DegreesInvesting/create_interactive_report/pull/14 to ensure functionality of the interactive report is up to date.

This (draft) PR:
- enables report creation for portfolios with only of asset type of bonds or equity
- checks if results exist per asset_type before loading the data for processing in the report function
- if results do not exist for an asset type, a corresponding tibble with NAs is created and passed instead
- this leads to successful report creation even if there are only bonds or only equity holding present
- In the report, asset type drop downs will simply have only one of the two values. Graphs with the default of the missing type are simply omitted.

OPEN:
- currently, the code does not check if at least one of the two types are given. It is not tested, what happens if both are missing. Could be an error, could be an empty report..
- In some test runs, bond holding always got the portfolio flag `valid_input == FALSE`, which leads to an error upon creation of the report. This has to be investigated but might come from some portfolio preparation step, possibly due to joining financial data in unexpeced formats

@Clare2D


jacobvjk:
the bonds issue from above is fixed. fin_data had too many columns and was causing colname troubles

jacobvjk:
@Clare2D this is the complementary PR for the handling of only EQ or CB portfolios in the report. Still needs to be reviewed and merged

PR #79, opened 2020-09-14
moved weight calc to 2_proj_...R

jacobvjk:
status
- moved calc of weights to script 2 in master.
- write new files

open:
- remove the relevant parts from script 3
- read in relevant new files
- do the same for web script


maurolepore:
Is this alive? Should be review/close?

jacobvjk:
@Clare2D will this still be relevant for @2diiKlaus ?

Clare2D:
it is still necessary - but perhaps it is easier to duplicate this calculation for Klaus. I had a quick look at it the other day and am a bit concerned about what makes sense on this. Given that there is a clear distinction of functionalities of the two scripts, perhaps we just add this calculation in a separate function that create the values that Klaus needs separately?

Clare2D:
We decided to leave this and pass this to Klaus to manage.

2diiKlaus:
Where can I find the function used in PACTA @jacobvjk @Clare2D ?

Clare2D:
This is however a fairly major structural change to move this forward. I would suggest we use the same function in duplicate to create the outputs you need. 

https://github.com/2DegreesInvesting/PACTA_analysis/blob/f4ba7c8083466545d5772e145c6514f81830acb9/0_portfolio_test.R#L85-L103

cjyetman:
resolved by #79

PR #80, opened 2020-09-14
Try runnign web tool following Jacob's guide

maurolepore:
This is work in progress. Here I draft an [rmarkdown file](https://github.com/maurolepore/PACTA_analysis/blob/mauro-wip/try-web-tool.md) that implements in code the instructions from the [Google Doc](https://docs.google.com/document/d/1wmSPN_7TnNkB4_wH67XL9tjNXZGrfdTa4F4jWn59v_8/edit) that Jacob wrote.

I intend to check the instructions and help detect potential issues. Eventually I may be able to suggest some ways to simplify or automate the workflow.

PR #81, opened 2020-09-15
Follow up: Try runnign web tool following Jacob's guide 

maurolepore:
This PR follows #80. Now from my fork of 2DegreesInvesting instead of from Jacob's fork.


maurolepore:
Beyond helping this project achieve specific goals, I would like to help the team work more smoothly. And I see one issue that adds friction.

This project needs PR to be merged into the branch `current_web_functionaliry` instead of the standard branch `master`. This makes the workflow more complex than the typical GitHub workflow.

For example, this is how I update my branch `pr` against `origin/pr` (my fork) and `upstream/master` (the main repo I forked from):

```bash
git merge origin && git merge upstream/master
```

 Instead, for this project I need to sync with

```bash
git merge origin && git merge upstream/current_web_functionality
```
 
This is no big deal but adds friction: I need to remember that I have a peculiar configuration, and I can't use tools that expect the standard setup (e.g. `usethis::pr_sync()`) -- which forces me to do manual work from the terminal that is error prone.

--

The solution is to stick to the GitHub workflow and keep only one long-lived branch:`master`; all other branches are short-lived, feature branches.

maurolepore:
@jacobvjk, I found an [error in line 34 of web_tool_script_1.R](https://github.com/maurolepore/PACTA_analysis/blob/mauro-wip/try-web-tool.md#exploring-the-error). It'd be great if you can fix it.

maurolepore:
@jacobvjk this is ready for review. See https://github.com/2DegreesInvesting/PACTA_analysis/pull/81#issuecomment-692904592

web_tool_script_1.R
```diff
@@ -26,15 +26,9 @@ source("0_global_functions.R")
 source("0_web_functions.R")
 source("0_json_functions.R")
 
-if (rstudioapi::isAvailable()) {
-  portfolio_name_ref_all <- c("TestPortfolio_Input")
-  working_location <- dirname(rstudioapi::getActiveDocumentContext()$path)
-  set_web_parameters(file_path = paste0(working_location,"/parameter_files/WebParameters_2dii.yml"))
-} else {
-  portfolio_name_ref_all = get_portfolio_name()
```
jacobvjk, line 34: I don't think we can remove this part, as this is where the script takes the information about the portfolio from, when called via command line. The webtool will mostly be called via command line in production, and get_portfolio_name() ensures it gets the correct information.
Same goes for Docker paramters instead of web parameters config. But happy to change it for the first part of the IF-statement

web_tool_script_2.R
```diff
@@ -19,16 +19,9 @@ source("0_portfolio_test.R")
...
 source("0_web_functions.R")
 
-
-if (rstudioapi::isAvailable()) {
-  portfolio_name_ref_all <- c("TestPortfolio_Input")
-  working_location <- dirname(rstudioapi::getActiveDocumentContext()$path)
-  set_web_parameters(file_path = paste0(working_location,"/parameter_files/WebParameters_2dii.yml"))
-} else {
-  portfolio_name_ref_all = get_portfolio_name()
-  working_location <- getwd()
```
jacobvjk, line 29: see above

web_tool_script_3.R
```diff
@@ -17,17 +17,11 @@ library(fs)
 library(jsonlite)
 library(fst)
 
-if (rstudioapi::isAvailable()) {
-  portfolio_name_ref_all <- c("TestPortfolio_Input")
-  working_location <- dirname(rstudioapi::getActiveDocumentContext()$path)
-  set_web_parameters(file_path = paste0(working_location,"/parameter_files/WebParameters_2dii.yml"))
-} else {
-  portfolio_name_ref_all = get_portfolio_name()
-  working_location <- getwd()
```
jacobvjk, line 26: see above

web_tool_script_3.R
```diff
@@ -17,17 +17,11 @@ library(fs)
...
-  set_web_parameters(file_path = paste0(working_location,"/parameter_files/WebParameters_docker.yml"))
-}
+portfolio_name_ref_all <- c("TestPortfolio_Input")
+working_location <- here::here()
+set_web_parameters(file_path = paste0(working_location,"/parameter_files/WebParameters_2dii.yml"))
 
-working_location <- paste0(working_location, "/")
+working_location <- paste0(here::here(), "/")
```
jacobvjk, line 24: agree with this but of course also for all 3 scripts

jacobvjk: and I am guessing here package will have to be added

parameter_files/WebParameters_2dii.yml
```diff
@@ -1,8 +1,8 @@
 default:
     paths:
-        project_location_ext: /Users/jacobkastl/Documents/git/PACTA_analysis/
-        data_location_ext: /Users/jacobkastl/Documents/git/pacta-data/2019Q4/
-        template_location: /Users/jacobkastl/Documents/git/create_interactive_report/
+        project_location_ext: ~/git/PACTA_analysis/
```
jacobvjk, line 3: I like the idea but I think this comes with the problem that it forces the user to have all repos in a directory called git. While I am personally happy with that, I am not sure everyone working on this has the same setup

jacobvjk (changes requested):
I will add some of the other changes you mentioned in the markdown file soon, especially the assertions/expectations.
I think we should discuss together with @Clare2D the idea of moving back to one main branch for all. While I agree it would be a lot nicer to maintain, Clare brought up some reasons as to why this was done in the first place. In either case, I feel like an overhaul of the repo will make it a lot easier to move back to one main branch logic.

web_tool_script_1.R
maurolepore: Thanks for checking. 

Okay, if the code does what you expect when called interactively and from the terminal, then it may only need a new  `else` block to define the behavior when called from a non interactive session but not in the terminal -- like I did by knitting an rmarkdown document. I'm unsure here because I don't understand what `commandArgs()` does (my experiments didn't clarify help).

If the `if()` `else` code gets long you may abstract the details in something like this (here in pseudo code). The code should do the same but should be easier to understand at first glance:

```r
get_parameters <- function() {
  user_environment <- get_user_environment()
  switch(
    user_environment,
    "rstudio" = get_rstudio_params(),
    "docker" = get_docker_params(),
    "shell" = get_shell_params()
    stop("Unknown user environment")
  )
}
```

parameter_files/WebParameters_2dii.yml
maurolepore: Yeah, here I was hard-wiring my own path. I was tempted to do something like this but suspected it wouldn't work:

```yaml
        project_location_ext: PACTA_analysis
        data_location_ext: pacta-data/2019Q4
        template_location: create_interactive_report
        project_location_ext: PACTA_analysis
```
But the code could be adapted so that it does work, for example, consider this pseudocode:

```r
project_location_ext_child_path <- config::get(file = "WebParameters_2dii.yml")$project_location_ext
project_location_ext <- here::here(project_location_ext_child_path)
```

maurolepore:
@jacobvjk, 

How about polishing the report I wrote to serve two useful purposes:

1. Help other people run the web tool as a parametrized rmarkdown report. The same report -- with example data -- could be included in README so that the instructions are not just words but code that actually runs.

2. Help us check the code still runs and that the instructions to run the web tool remain current. This we could do by rendering the report on GitHub actions, triggered by every new commit. We could improve the assertions I wrote to throw informative messages if the report fails to render. Is there anything like this already?

What do you think?

jacobvjk:
Definitely like the idea, but I doubt I will get to it before going on holidays on Friday. On my list for when I'm back though, @maurolepore

maurolepore:
The file try-web-tool.Rmd is now parametrized:

* You may see the rendered output [here](https://github.com/maurolepore/PACTA_analysis/blob/mauro-wip/try-web-tool.md).
* The button _Knit > Knit with params ..._ now gives you an interface like this one. Notice the parent path is not hard coded; it will reflect the system of the user:

image

You can also pass parameters programatically. For an example see the file [render-with-params.R](https://github.com/maurolepore/PACTA_analysis/blob/mauro-wip/render-with-params.R). This can be used, for example, during development to test the output of the web tool remains the same -- locally or on GitHub actions.

I hope @jacobvjk or @Clare2D can review this PR and discuss. I have no rush to merge it and I'm happy to discuss what I did here and happy to hear your suggestions for chage.

Clare2D:
A little slow on the uptake here Mauro! Will check this out properly though, as this looks really nice!

maurolepore:
This PR now includes an integration test on GitHub actions. The gh-actions workflow run-integration-test.yaml clones the repositories pacta-data/ and create_interactive_report/ alongside this repository PACTA_analysis/, then renders an .Rmd file that runs the entire PACTA_analysis based on @jacobvjk's guide. 

This integration test will run on every push and PR, meaning that we'll know whether or not a new commit brakes something in PACTA_analysis.

Right now, the workflow runs on mac os only, but I plan to extend it to other platforms. We an use this information to tell our users which environments we have tested this tool with, which could replace the more fragile, hard-coded descriptions of the environments that I see on README.

maurolepore:
I'll close this PR to resubmit the cleaner version: #90

PR #87, opened 2020-09-25
Fix wrong syntax

maurolepore:
This PR fixes a syntax error. Please review carefully -- I have no tests to 
ensure this doesn't break anything.

### DIFF
--- a/0_stress_test_functions.R
+++ b/0_stress_test_functions.R
@@ -52,7 +52,7 @@ if (rstudioapi::isAvailable()) {
 ### Load in Audit File ###
 
 
-process_audit_filt <- function(all_holdings, ){
+process_audit_filt <- function(all_holdings){
   port <- merge(all_holdings, revenue, by = "isin", all.x = T)
   
   port$Value.USD <- as.numeric(port$Value.USD)


maurolepore:
@Clare2D, this PR seems safe to merge -- I did the same change in #90 and the integration test continued to pass.

Clare2D:
I'll merge now and get @jacobvjk to check the spelling of "filk" later ;)

jacobvjk:
@maurolepore @Clare2D these functions have little to do with what we currently run as the stress test. This must be a pretty old version, probably used in the BoE study and I would suggest I do not try to dive into this as we should use the new version from this Stress Testing repo anyway https://github.com/2DegreesInvesting/StressTestingModelDev

maurolepore:
Thanks Jacob. It's good to know the change should be safe. If you are confident that is [dead code](https://refactoring.com/catalog/removeDeadCode.html), best practice is to remove it. Git can always find it should we later need it again.

jacobvjk:
@Clare2D since this is the PACTA_analysis repo, not the Stress testing one, I feel like I should validate with you if removing the boe stress testing section here is okay. What do you think? Is this being used be anyone at all?

maurolepore:
If it's unclear, I suggest you open another issue to remember, and come back to it once we have more tests -- i.e. a safety net to know if removing this code would actually break anything useful.

PR #89, opened 2020-09-26
Add packages.R to centralize packages

maurolepore:
This commit adds the new file packages.R. I found all packages
with the terminal command at the top of packages.R. An alternative
to find packages is env::dependencies(), but it seems to not find
packages on GitHub.

The next step is to remove all calls to library(), to keep the code
clean. Instead, you can now `source("packages.R")`.

Another option is to add a DESCRIPTION file and list the packages
under the field "Imports". That is nice because it brings features
from R packages; but for now this commit keeps it simple.


maurolepore:
@Clare2D, in this PR I suggest we should centralize the scattered calls to `library()`. If this sound like a good idea, we could either (a) merge it now and later add another PR to `source("packages.R")` where it's needed, or (b) extend it now to  `source("packages.R")` where it's needed.
If you dislike the idea, feel free to close it.

Clare2D:
I like the idea of adding the packages to a DESCRIPTION file but as you suggest this method is really clean and simple. I will merge now!

PR #90, opened 2020-09-26
Integration test

maurolepore:
Ralates to #81 -- this is a cleaner version.

This PR adds an [integration test](https://github.com/maurolepore/PACTA_analysis/blob/integration-test/integration-test.md). It runs PACTA_analysis based on 
[@jacobvjk's instructions](https://docs.google.com/document/d/1wmSPN_7TnNkB4_wH67XL9tjNXZGrfdTa4F4jWn59v_8/edit). 

You can run this integration test manually with
```r
rmarkdown::render("integration-test.Rmd")
```
or automatically on push or pull request to GitHub, which runs it on the
three main platforms and with multiple versions of R. For details see the 
file "[.github/workflows/run-integration-test.yaml](https://github.com/maurolepore/PACTA_analysis/blob/integration-test/.github/workflows/run-integration-test.yaml)" or the [Actions tab of 
my fork](https://github.com/maurolepore/PACTA_analysis/actions?query=branch%3Aintegration-test), which looks like this:


maurolepore:
@Clare2D, this PR is ready for your review; to understand what it does you may follow the links above and ask me questions. It should be safe to merge, and will make merging future PRs safer. It has multiple commits but they should all be squashed -- you may do ahead and do it or say LGTM and I can squash-merge myself.

Clare2D:
This looks really great Mauro. I've been through your instructions and have understood most of what the test is trying to do. 
There are a few things that might make sense to discuss with @jacobvjk just before integration. 

- Not all users running the PACTA_analysis repo need the create_interactive_report to work. If they don't will this cause errors? Thinking about this, perhaps this test just ensures that the create_interactive_report would work and it doesn't matter if an external user then doesn't need this as they are not running this test. 

- We should discuss the webtool scripts and branch. It seems that merging the branch this is the better thing to do. 

- The "clean_files" are something that generally could be refactored a little. @jacobvjk  should probably be in this discussion as well. Basically we want to reduce the running time for the script but doing some precleaning of the larger data files that we generally use. These files could also be cut down in size (excess columns) and then saved in a temp location (perhaps?).

maurolepore:
> Not all users running the PACTA_analysis repo need the create_interactive_report to work. 

> If they don't will this cause errors? 

No. The only way for integration-test.Rmd to cause errors is when the "user" of PACTA_analysis pushes a commit that changes the underlying code in a way that makes integration-test.Rmd throw an error. And whoever does that, should more precisely be referred to not as a "user" but as a "developer" or "contributor". (The "user" and "developer/contributor" may be the same person, working with different goals -- in "user" mode, I want to get something out of the system; in "developer/contributor" mode I want to change the system.)

> Thinking about this, perhaps this test just ensures that the create_interactive_report would work
 
Yes, it ensures that `create_interactive_report()` would work -- but not just `create_interactive_report()`. A test for  specifically `create_interactive_report()` would be a "unit" test. AFAIK, we lack such unit tests. The file "integration-test.Rmd" is instead an integration" test -- that is, it tests that the entire system works. 

This integration test is our first line of defense against introducing errors. Now that we have a working system, we need to ensure it remains working whatever we do. This enables a top-down approach: we start by capturing the working state of the system from end to end (i.e. the repos pacta-data/, PACTA_analysis/, and create_interactive_report/), then we add more granular tests to capture the working state of each module of the system (e.g. each repo independently), and each unit in each module of the system (e.g. each of the main functions of each repo). The process will expose bugs (that we can then fix) and uncertainties (that we'll need to discuss and disambiguate express in tests), and leave the system in a more predictable and robust state.

If someone was climbing a wall, they would now be hooked to the top of the wall -- but only there. If they fell, we could now know without having to look at the climber -- we can just sense the increased tension of the rope at the top hook. But we wouldn't know precisely where they fell. As we add more hooks and sensors along the wall, we'll be ever more able to tell where precisely they fell and what move might have cause the fall. Eventually we'll have hooks in every tricky spot of the wall, and all the moves will be relatively safe. Then a climber might still fall but the consequences should be minimal.


> it doesn't matter if an external user then doesn't need [`create_interactive_report()`].

Exactly.

> as they are not running this test.

Running "integration-test.Rmd" will unlikely be their primary goal for changing the system, but it will be one way to check their intended change doesn't deteriorate the system in an obvious way. A developer or contributor should always run all tests before they propose a change (via a PR) to the system. Now they can run  `rmarkdown::render("integration-test.Rmd")` locally; but even if they don't, `rmarkdown::render("integration-test.Rmd")` will now run on GitHub actions (automatically, with the three main platforms, and with multiple versions of R). Passing this integration test should now be the lowest bar to merge a commit into the main line of development.

> We should discuss the webtool scripts and branch. It seems that merging the branch this is the better thing to do.

Great. Maybe this could be the focus of a call (let's try find a suitable time). Having a single "main line of development" is a good idea. But I still don't know the reason we currently have two lines of development (`master` and `current_web_functionality`); once we discuss that, I should be able suggest how to restructure the project (merging the two lines is one option; another option might be to extract some parts into a separate repository). This topic seems to deserve its own issue.

> The "clean_files" are something that generally could be refactored a little. @jacobvjk should probably be in this discussion as well. Basically we want to reduce the running time for the script but doing some precleaning of the larger data files that we generally use. These files could also be cut down in size (excess columns) and then saved in a temp location (perhaps?).

I haven't looked at the "clean_files" yet. It seems you developed a cache system. This topic seems to deserve its own issue.

jacobvjk:
@maurolepore @Clare2D 

- the integration test itself: lgtm
- regarding the branch/repo discussion: would be great to talk about this this week, as we should also consider how the stress testing modules will fit into the future repo structure and this should ideally already be considered when I make the changes in that for P2020
- regarding the "clean_files" or the pacta_data more generally: I think for the time being there is not pressure to get this refactored until the P2020 deadline. It's quick and easy to convert the csvs and as long as we agree on formats and locations for those files, we should be good. I would say the time is spent better focussing on the other remaining issues. @maurolepore given we want to basically get rid of the pacta-data repo in favor of directly placing the files on an SFTP server hosted by or webdev agency Constructiva, we need to consider the implications of that for your suggestions, especially the integration test.

maurolepore:
In this conversation I see two independent discussions which I extracted to issues of their own:

* Can we use a single main line of development? #92 
* Refactor "clean_files": precleaning of the larger data files to save computation time #93

Regarding the integration test, it seems both are okay to merge, so I'll go ahead and do it.

PR #98, opened 2020-09-29
Reorganize #97: Revert offending commit

maurolepore:
- first step functionality webtool.
- flexibility for webtool included
- update fund file name in code
- Start of options for webtool
- start of funds processing
- Added start year check
- corporate bonds fix, type classification
- filter small holdings
- json output for audit charts
- New web tool scripts
- Web functions for json files from Andy
- New Git Ignore
- Added todo
- Fixed for no fund file available
- Improving on saving of cleaned files
- Delete files to change branch
- Added the function export_report_content_variables_json
- cleaned old references
- Add improved file read in functionality and parameter files
- cleaned up folder file structures
- update gitignore
- Added parameter files
- formatted json files for webtool
- removed ref to old audit chart and webtext
- update git ignore
- refix json function
- adjust set portfolio parameters
- cleaned timestamp inputs and removed punctuation clean function
- Fixed meta portfolio code
- changes to the input file reading to check for input arguments
- new read and write file functions that take the extension type from the file name
- reworked parameter file
- sourced functions and gitignore update
- Move up function and removed defunct test
- Resave of currencies file Shift of webparameters to be relevant for environment
- Cleaning and adding debt fin data file
- ungroup added
- Clean up web tool script
- Update with web folders
- new analysis parameters
- Add in saving by portfolio
- Updated parameter files
- clean script
- Update git ignore and fix script
- Add bics-sector to audit file. It should now come through the financial data.
- Inclusion of the create_interactive_report
- Updates to parameter files
- Initial updates to web_script_2
- move back to just working_dir
- correction of parameter file endings
- fixbics_sector bug in create_audit_file()
- keep market_cap and bond_debt_out in import comp_fin_data for pf_emissions
- add Global to default EEquity.Market.List in param file
- fix duplicate grouping by allocation in aggregate_map_data()
- change output directories to ./project/40_results/portfolio_filename/
- added function to write error messages into log file
- Added log messages to check for data consistency
- changed naming of error logfile
- correct creation of subfolders in grouped pf input case
- move function def of write_log to 0_global_functions.R
- orrected file paths
- fixed some logging messages
- break code when missing cols, rather than warning
- break code when missing cols, rather than warning
- surface imports of scenario and masterdata for eq and cb pout of the loops
- correct the file paths
- changed deletion of past results to only delete within the project folder
- merge different portfolio names into the same one, if multiple portfolio names given
- remove unnecessary loop over pf and on obsolete variable
- remove obsolete loops over pf in analysis
- reset to default paths and files
[...]

PR #97, opened 2020-09-29
Webtool dir fixes

jacobvjk:
same as here: https://github.com/2DegreesInvesting/PACTA_analysis/pull/96

PR #96, opened 2020-09-29
Webtool dir fixes

jacobvjk:
This PR incorporates the following changes to the path structure and file naming conventions, as dicsussed with Constructiva:

- move file_names.csv to a lower level folder (it was moved from /30_Processed_Inputs to /30_Processed_Inputs/pf_name/. This ensures wrong file names are not kept between runs, but are solely written to the directory which reflects the corresponding portfolio.
- ensure git tracks the subfolders /30_Processed_Inputs, 40_Results and 50_Outputs by adding placeholder .gitkeep.txt files and adapting the overall .gitignore command to allow tracking these placeholders.
- use investor_name and portfolio_name from the parameter files, namely from pf_name_Input_PortfolioParameters.yml. The entries to this parameter file are provided by Wim when calling the tool and ensure that investor and portfolio names match the grouping as requested by the user.

OPEN:

- should we remove the pf_name prefix for all files in the /30_Processed_Inputs directory for consistency purposes? Downside: this will look a bit different to the offline version. Not necessarily a problem, but something to keep in mind/consider
- resolving conflicts

PR #99, opened 2020-09-29
Rework dir fixes (#96, #97, #98)

maurolepore:
Follow up #96, #97, #98


maurolepore:
@jacobvjk, this PR is the same as your #96 and #97  but with the commits rearranged. 

* I rebased your PR on top of current_web_functionality to get a linear history that that I could understand more easily.
* I reseted all commits and re-commited to separate simple from more complex changes.
* I used `git bisect` to detect which commits broke the integration test ([video](https://youtu.be/uFOnezhSxAg)).
* I gathered the bad commits into a single one and move it to the top of the history.
* I reverted the offending commit.

I hope this helps you to detect the problem and reapply the changes I reverted, with the bug fixed. But feel free to
reopen your PR and close this one -- if that makes more sense to you.

.gitignore
```diff
@@ -33,10 +33,16 @@ vignettes/*.pdf
...
+!**/working_dir/40_Results/.gitkeep.txt
+**/working_dir/50_Outputs/*
+!**/working_dir/50_Outputs/.gitkeep.txt
 
 setup.R
 data/isins_missing_bbg_data.csv
 data/fund_isins_without_bbg_data.csv
+
+# Mac files
+working_dir/DS_STORE
```
maurolepore, line 48: I think I've seen a file called .DS_Store. Would this catch it?

PR #100, opened 2020-09-29
Rework dir fixes

maurolepore:
Follow #99 (I know, I'm a nightmare -- sorry).


maurolepore:
Ok, #99 works well. I'll stick to that one.

PR #102, opened 2020-09-30
Add package infrastructure

maurolepore:
This PR add the minimal infrastructure to work with tests confortably.
It's safe because it touches no code and only builds infrastructure 
around existing files. The only files that I touched are the .rda files
in data/ -- which I saved from data-raw/ in a way that avoids 
an error about magic numbers. The takeaway is this: I epect the 
data behave the same than before, but if there is something
wierd about data, then this may be the cause.


maurolepore:
Follow at #103

PR #103, opened 2020-09-30
Add package infrastructure

maurolepore:
This PR add the minimal infrastructure to work with tests confortably.
It's safe because it touches no code and only builds infrastructure
around existing files. The only files that I touched are the .rda files
in data/ -- which I saved from data-raw/ in a way that avoids
an error about magic numbers. The takeaway is this: I epect the
data behave the same than before, but if there is something
wierd about data, then this may be the cause.


maurolepore:
@Clare2D and @jacobvjk, if you can please add your author details in DESCRIPTION

DESCRIPTION
```diff
@@ -0,0 +1,31 @@
...
+      person(given = "Jacob",
+             family = "Kastl",
+             role = c("aut"),
+             email = "jacob.kastl@gmail.com"), 
+      person(given = "2 Degrees Investing Initiative",
+             role = c("cph", "fnd"),
+             email = "contact@2degrees-investing.org"))
+Description: For more information visit
+    <https://2degrees-investing.org/>.
+License: MIT + file LICENSE
```
maurolepore, line 23: @Clare2D, are we okay with MIT license?

Clare2D:
This is fine. Thanks

jacobvjk:
LGTM

maurolepore:
I'm merging this despite this:

```
Error: At least one argument must be supplied (input file).n
```

This error comes from get_portfolio_name()`, which I'll try debug and test in a separate PR.

PR #106, opened 2020-10-01
Dry setup

maurolepore:

PR #108, opened 2020-10-02
Naming 30 inputs

jacobvjk:
This PR will:

- remove the pf name as a prefix of file names in the folders "30_...", "40_...", "50_..." as this info is already given by the folder name (this is the change discussed with Wim)
- adapt the integration test so it adds the .gitkeep.txt files after deleting  and recreating the corresponding directories (this change just ensures the .gitkeep files do not get accidentally removed by running the test, rendering them useless)

@Clare2D @maurolepore I am not expecting this to pass the test on gh actions, but it did locally, so I would argue to merge this


Clare2D:
grafik
LGTM ;) Having spoken with CJ, I'm not worried about this.

integration-test.Rmd
```diff
@@ -95,6 +95,7 @@ ensure_empty_directory <- function(directory) {
   }
 
   dir_create(directory)
+  file_create(path(directory, ".gitkeep.txt"))
```
maurolepore, line 98: 1. Is the result the same as to not deleting .gitkeep.txt? 

I think .gitkeep.txt has some useful comments that explain why the file exist. But here 
we would end up with an empty file, right?

maurolepore: Would it be better to instead not remove .gitkeep.txt? (See https://github.com/2DegreesInvesting/PACTA_analysis/pull/106/commits/f1786137e5f912474d7f2629b6cb45fd8dc882cd)

web_tool_script_1.R
```diff
@@ -187,12 +187,12 @@ export_audit_information_jsons(audit_file_ = audit_file %>% filter(portfolio_nam
...
-save_if_exists(audit_file, portfolio_name, paste0(proc_input_path_, "/",portfolio_name,"_audit_file.rda"))
-save_if_exists(emissions_totals, portfolio_name, paste0(proc_input_path_, "/",portfolio_name,"_emissions.rda"))
+save_if_exists(audit_file, portfolio_name, paste0(proc_input_path_, "/audit_file.csv"), csv_or_rds = "csv")
+  
+save_if_exists(portfolio_total, portfolio_name, paste0(proc_input_path_, "/total_portfolio.rda"))
+save_if_exists(eq_portfolio, portfolio_name, paste0(proc_input_path_, "/equity_portfolio.rda"))
+save_if_exists(cb_portfolio, portfolio_name, paste0(proc_input_path_, "/bonds_portfolio.rda"))
+save_if_exists(portfolio_overview, portfolio_name, paste0(proc_input_path_, "/overview_portfolio.rda"))
+save_if_exists(audit_file, portfolio_name, paste0(proc_input_path_, "/audit_file.rda"))
+save_if_exists(emissions_totals, portfolio_name, paste0(proc_input_path_, "/emissions.rda"))
```
maurolepore, line 197: It looks like the same change happens in 3 places. If so, can you DRY? (See https://github.com/2DegreesInvesting/PACTA_analysis/pull/106/commits/8d8a98d5a7ff62d55c3d9f5d1b1f86ed832d6d66)

maurolepore (changes requested):
Did you find a solution to stop .DS_Store from showing up everywhere?

I see this file here: "working_dir/.DS_Store"

maurolepore:
Thanks @jacobvjk, I made some general and specific comments above. Also:

> I am not expecting this to pass the test on gh actions, but it did locally, so I would argue to merge this

The failure I see relates to permissions. Can you avoid that problem if you push directly to the source repo instead of to your fork? I need to figure out a solution that allows you to push from your fork, but until then, I think it's best to push to the source repo and benefit from the test than to push to your fork and by-pass the test.

jacobvjk:
> The failure I see relates to permissions. Can you avoid that problem if you push directly to the source repo instead of to your fork?

sure @maurolepore  https://github.com/2DegreesInvesting/PACTA_analysis/pull/110 I will close this one

PR #110, opened 2020-10-02
Naming 30 inputs (directly to upstream)

jacobvjk:
This PR will:

- remove the pf name as a prefix of file names in the folders "30_...", "40_...", "50_..." as this info is already given by the folder name (this is the change discussed with Wim)
- adapt the integration test so it adds the .gitkeep.txt files after deleting and recreating the corresponding directories (this change just ensures the .gitkeep files do not get accidentally removed by running the test, rendering them useless)

@Clare2D @maurolepore I am not expecting this to pass the test on gh actions, but it did locally, so I would argue to merge this

this replicates the following PR (https://github.com/2DegreesInvesting/PACTA_analysis/pull/108) but I pushed directly to upstream to avoid permissions issues in the GH Actions


maurolepore:
Thanks! The permissions are now okay. The failure  I see should go away if we merge #106 first.

PR #115, opened 2020-10-05
Merging in Current Web Functionality

Clare2D:
Initially checking the merge and will see what needs to be resolved.

PR #116, opened 2020-10-05
Details after we merge current_web_functionality into master

maurolepore:
This PR contains the few commits that reflect the discussion between @Clare2D and @jacobvjk in #115.

As I suggested in https://github.com/2DegreesInvesting/PACTA_analysis/pull/115#issuecomment-703747504, I think it's best to merge `current_web_functionality` into `master` first, then follow up with this "details" separately. (So this PR should stay as a draft until we merge  `current_web_functionality` into `master`.)


maurolepore:
I messed up the git history of this branch so I'll resubmit.

PR #118, opened 2020-10-06
New PR to merge current_web_functionality into master

Clare2D:


web_tool_script_3.R
```diff
@@ -0,0 +1,211 @@
...
+
+repo_path <- template_path
+template_dir <- fs::path(template_path, "template_beta")
+company_charts_dir <- fs::path(template_path, "company_charts", "Mixed_Portfolio")
+output_dir <- paste0(outputs_path,"/", portfolio_name_ref_all, "/")
+project_name <- "working_dir"
+investor_name <- file_names$investor_name
+portfolio_name <- file_names$portfolio_name
+# start_year <- 2020
+scenario <- "B2DS"
```
jacobvjk, line 53: is this the really the desired scenario?

working_dir/10_Parameter_File/WebParameters.yml
```diff
@@ -0,0 +1,9 @@
+default:
```
jacobvjk, line 1: shouldn't this be `WebParameters_2dii.yml` and `WebParameters_docker.yml` ?
Also, I think at the current stage of the code this should have an entry for `template_location`

@Clare2D

jacobvjk (changes requested):
see comments

web_tool_script_3.R
jacobvjk: this is not a required change, just something to think about

integration-test.Rmd
```diff
@@ -0,0 +1,330 @@
...
+  contents <- dir_ls(path, recurse = TRUE)
+  has_files <- any(map_lgl(contents, is_file))
+  has_files
+}
+
+out_1 <- path("working_dir", "30_Processed_Inputs")
+
+expect_false(dir_has_files(out_1))
+source("web_tool_script_1.R")
+expect_true(dir_has_files(out_1))
```
maurolepore, line 258: I see this error (which surprisingly before used to be just a warning). Can we add the file "fund_data.fst" or remove the code that uses it?

```r
Loading PACTA.analysis
Quitting from lines 246-259 (integration-test.Rmd) 
Error: Error: '/home/runner/work/PACTA_analysis/PACTA_analysis/pacta-data/2019Q4/cleaned_files/fund_data.fst' does not exist.
Execution halted
```

maurolepore (changes requested):
This is the only problem I see. If you know what's going on, let me know; otherwise I'll try debug it myself and hopefully push a fix ASAP -- here is a reproducible example; notice the error at the end:

``` r
devtools::load_all()
#> Loading PACTA.analysis
test_integration()
#> processing file: integration-test.Rmd
#> 
#> Attaching package: 'scales'
#> The following object is masked from 'package:purrr':
#> 
#>     discard
#> The following object is masked from 'package:readr':
#> 
#>     col_factor
#> 
#> Attaching package: 'reshape2'
#> The following object is masked from 'package:tidyr':
#> 
#>     smiths
#> 
#> Attaching package: 'tidyselect'
#> The following object is masked from 'package:testthat':
#> 
#>     matches
#> 
#> Attaching package: 'jsonlite'
#> The following objects are masked from 'package:rlang':
#> 
#>     flatten, unbox
#> The following object is masked from 'package:purrr':
#> 
#>     flatten
#> Loading PACTA.analysis
#> Quitting from lines 246-259 (integration-test.Rmd)
#> Error: '/home/mauro/git/pacta-data/2019Q4/cleaned_files/fund_data.fst' does not exist.
```

<sup>Created on 2020-10-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

integration-test.Rmd
jacobvjk: That depends on where the test reads that data from @maurolepore we can't add the file to the pacta-data gh repo, as it is more than 100mb in size. we could add it to the data folder locally though and the idea is that this data will be transferred to constructiva via sftp in the future, not via github. So in that sense, we can add it. But we might have to reflect these changes in the test and docker workflows..

maurolepore: Thanks! And what should happen if the file isn't available? (Before I used to get a warning only -- not an error -- so I suspect there is a way for the analysis to continue without this file).

jacobvjk: there is indeed. the portfolio can also be processed without fund data. It is the same discussion we are having in the pacta2020 slack channel and @cjyetman opened a GH issue for this: https://github.com/2DegreesInvesting/PACTA_analysis/issues/123

working_dir/10_Parameter_File/WebParameters.yml
jacobvjk: forget this comment, I did not realise this file is in `/working_dir/10_parameter_files`  this should not interfere with aything

maurolepore:
Please follow up at https://github.com/2DegreesInvesting/PACTA_analysis/pull/127.

PR #114, opened 2020-10-05
Include stress test

jacobvjk:
This PR will:

- source a webtool version of the main stress test calculation from https://github.com/2DegreesInvesting/StressTestingModelDev
- automatically write the stress test results into /working_dir/40_Results
- remove the old BoE Stress test scripts from this repo

Issue #125, opened 2020-10-06
get rid of .DS_Store forever


cjyetman:
Though `.DS_Store` has already been added to the `.gitignore` file, it still exists at the root and numerous sub-directories. This is likely due to those files being added to the git cache before `.DS_Store` was added to the `.gitignore` file, so they are still being tracked. I believe if we make one PR that deletes all of the `.DS_Store` files that are currently in the repo, the gitignore command will prevent any of them coming back after that.

cjyetman:
this appears to have been addressed in a number of reason merges... closing

PR #127, opened 2020-10-06
Revert the latest three problematic commits, and fix gh-actions workflow

maurolepore:
This PR reverts the three latest commits, to restore a state where the integration test passes.

The reverted commits are not lost, and can be reverted back to restore the changes you
made. However, I suggest we add those changes in smaller steps, adding one commit
at a time and checking the tests continue to pass.

This approach should allows us to notice if any change introduces a braking change, and
fix it before moving on.

This is motivated by the fact that both @cjyetman and me want to fix a bug each, 
but this is hard to do against a broken state because this obscures any new
bugs that we we might unintentionally introduce.

-- 

Surprisingly, reverting the commits was insufficient to make the integration test pass
and I had to add a tiny commit (ed68290) that fixes that problem (although I can't
explain why I and won't bother to getting to the bottom of it). This tiny fix should
also fix the problem I see in https://github.com/2DegreesInvesting/PACTA_analysis/pull/114#issuecomment-704473677.


maurolepore:
@jacobvjk, after your review, feel free to merge but please do NOT squash-merge. Instead, please use the option "Create a merge commit", which I think will be best for later re-adding the commits I'm reverting here.

.gitignore
```diff
@@ -1,38 +1,10 @@
-### MacOS specific ###
```
jacobvjk, line 1: @maurolepore should we really remove all these lines from .gitignore?

1_portfolio_check_initialisation.R
```diff
@@ -30,26 +30,16 @@ source("0_portfolio_input_check_functions.R")
...
-
-# within the "project_setting.yml" config file, set the project_name, the twodii_internal switch,
-# and the external data locations, if necessary.
-# the project_name will determine the name of the folder that is created for this project
-# Set twodii_internal to TRUE to run the analysis on an internal 2dii laptop
-# This setting uses the dropbox connection for data import
-# Set twodii_internal to FALSE, tu use external data locations
-# Specify these data locations in the config file "project_settings.yml" in the repo
-
-cfg <- config::get(file = "project_settings.yml")
```
jacobvjk, line 44: when we merge this with master, we definitely need to go for the config version (i.e. master version)

jacobvjk (commented):
I will merge this after our daily call with constructiva, as this changes the curernt_web_functionality dramatically and I dont want to interfere with the work last minute

.gitignore
cjyetman: yes, please keep at least `.DS_Store` in `.gitignore` see #125

maurolepore: @jacobvjk, this PR reverts commits that I did not author. 

image

The next step is to re-add this commits but in much smaller steps so we ensure the tests continue to pass along the way. The best person to re-add those commits are the authors of the reverted commits, because they know the intention behind those changes.

Alternatively, I can try brake down all those changes in little bits that make sense to me and ask questions as they arise.

PR #128, opened 2020-10-06
Show/solve conflicts between packages more explicitely

maurolepore:
This PR makes explicit how we prefer to solve conflicts between functions from 
different packages. The way I declared the preferences is my guess; I need
someone with experience on this repo to review it. 

For the record, I initially submitted this commit in https://github.com/2DegreesInvesting/PACTA_analysis/pull/114
because I thought it would fix a problem I saw there. Finally I realized the problem
was fixed by a different commit, so I reverted it there and pushed it here -- to
keep each PR as focused and small as possible.


integration-test.Rmd
```diff
@@ -26,15 +26,18 @@ knitr::opts_chunk$set(
...
-suppressPackageStartupMessages(library(glue))
-suppressPackageStartupMessages(library(fs))
+library(tidyverse)
+library(devtools)
+library(testthat)
+library(config)
+library(rlang)
+library(renv)
+library(glue)
+library(fs)
```
maurolepore, line 36: I no longer suppress package startup messages. The output is long but informative.

integration-test.Rmd
```diff
@@ -26,15 +26,18 @@ knitr::opts_chunk$set(
...
 library(here)
+library(conflicted)
+conflicted::conflict_prefer("filter", "dplyr")
+conflicted::conflict_prefer("lag", "dplyr")
```
maurolepore, line 40: Getting some help from the package "conflicted". Eventually I would prune to a minimum the packages we depend on.

maurolepore (commented):
Here are some highlights of this PR.

integration-test.Rmd
jacobvjk: sounds good and agreed. I will take a look into best practices on how to minimise dependencies and check which ones best to remove

jacobvjk (approved):
As the main changes are the ones you already mentioned in the comments I would say:
LGTM

integration-test.Rmd
maurolepore: I think @cjyetman once lead a ds-incubator on the topic?
A good place to get some ideas is https://www.tidyverse.org/blog/2019/05/itdepends/

cjyetman: very much in favor of reducing dependencies... and thereby reducing conflicts, and reducing package version issues, and reducing system setup time/complexity, and reducing run time, and reducing other random problems 😄

also be aware that the order in which packages are loaded will affect which conflicting functions take precedence... and when you run multiple scripts and/or do not reset your session, what's already loaded in your environment can change what the script you're looking at appears to suggest as the load order (e.g. the last package to load may have already been loaded by a previous script)

`tidyverse` is an obvious one to try to get rid of.... I believe that `library(tidyverse)` is intended as a shortcut for interactive use, but it's probably much safer in a script to only load the specific packages that are needed from `tidyverse`

PR #129, opened 2020-10-07
read_file() now reads RData saved as .rds and .rda

### LINKED ISSUE #124: deal with inconsistent use of .rda extension
In some cases, RDS files are saved with the `.rda` extension. This causes a bunch of confusion, and breaks functionality of,  for instance, `read_in_file()`

https://github.com/2DegreesInvesting/PACTA_analysis/blob/acdedabadccdcdeb05cc2e0384f9c78e77a5abd6/0_global_functions.R#L24-L43

maurolepore:
Closes #124

This PR is ready for review but I make it a draft because it is not ready to be merged -- first we need to merge #127. Once that's done I may  need to rebase `current_web_functionality` and resubmit.

### DIFF
--- a/0_global_functions.R
+++ b/0_global_functions.R
@@ -1,6 +1,6 @@
 # Checks whether a variable is a dataframe. Considers also logicals and null values.
 data_check <- function(df){
-  
+
   if (is.data.frame(df)){
     if(nrow(df) > 0){
       check <- TRUE
@@ -10,9 +10,9 @@ data_check <- function(df){
   }else{
     check <- FALSE
   }
-  
+
   return(check)
-  
+
 }
 
 # Checks whether a value is null or blank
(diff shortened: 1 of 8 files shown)


R/read_file.R
```diff
@@ -0,0 +1,80 @@
...
+#' @param file_name Character; path to a file storing a single object.
+#'
+#' @noRd
+#' @examples
+#' # Odd case: .RData saved with extension .rda
+#' path <- tempfile(fileext = ".rda")
+#'
+#' data1 <- data.frame(x = 1)
+#' saveRDS(data1, path)
+#' read_file(path)
```
cjyetman, line 12: Either the description or the concept behind this example/test is off. What is happening in the code here is that an RDS file is being saved with the file extension ".rda"

R/read_file.R
```diff
@@ -0,0 +1,80 @@
...
+    rlang::warn("The `path` does not exist: {path}")
+    return(invisible(path))
+  }
+
+  abort_unsupported_extension(path)
+
+  extension <- fs::path_ext(path)
+  switch(
+    extension,
+    "rda" = read_rda(path),
```
cjyetman, line 36: I think we should consider adding ".rds" as a supported extension also, that way if we ever begin the process of properly naming exported files, this function will continue to work in place, e.g. if a script uses this function to read in an RDS file that is named "data.rda" and we eventually fix the upstream code that creates "data.rda" to export that data with the proper file name/extension "data.rds" then the script could continue to function without modification.

R/read_file.R
```diff
@@ -0,0 +1,80 @@
...
+  if (!supported_extension(path)) {
+    rlang::abort(glue::glue("Unsupported `path` extension: {path}"))
+  }
+
+  invisible(path)
+}
+
+supported_extension <- function(path) {
+  fs::path_ext(path) %in% c("rda", "csv", "fst")
+}
```
cjyetman, line 53: I don't understand/agree with the idea of creating single line functions that are only ever called once. In my opinion, it substantially degrades the readability of the code. I would prefer this type of code to be inline with the code that calls it so that it's much easier to follow.

R/read_file.R
```diff
@@ -0,0 +1,80 @@
...
+  tryCatch(
+    # Unfortunately, legacy code saved .rds files with the extension .rda
+    readr::read_rds(file_name),
+    error = function(e) read_rdata(file_name)
+  )
+}
+
+read_rdata <- function(path) {
+  e <- new.env()
+  load(path, envir = e)
```
cjyetman, line 65: I'm pretty sure that `load()` invisibly returns a character vector of the all the names of the objects it created, which could then easily be tested for `length() == 1L` and used as a single character value if passed.

R/read_file.R
```diff
@@ -0,0 +1,80 @@
...
+}
+
+abort_if_too_long <- function(dataset) {
+  nn <- length(dataset)
+  if (nn > 1L) {
+    rlang::abort(glue::glue("`path` must have a single object but has {nn}."))
+  }
+
+  invisible(dataset)
+}
```
cjyetman, line 80: This also seems like something that could very easily, and much more succinctly be inlined in the parent/calling function.

0_global_functions.R
```diff
@@ -21,310 +21,289 @@ is_blank_na <- function(x){
...
-
 save_file <- function(df, file_name){
-  
-  
+
+
   extension_type = substring(file_name,nchar(file_name)-4+1)
-  if (!extension_type %in% c(".rda", ".csv", ".fst")){stop("Can't read in file. File extension type not applicable.")}  
-  
+  if (!extension_type %in% c(".rda", ".csv", ".fst")){stop("Can't read in file. File extension type not applicable.")}
```
cjyetman, line 28: Why is the `save_file()` function reporting that it "Can't read in file."?

0_portfolio_input_check_functions.R
```diff
@@ -1,54 +1,55 @@
+devtools::load_all()
```
cjyetman, line 1: I still don't understand what this function does. Its help file describes it as "load_all loads a package. It roughly simulates what happens when a package is installed and loaded with library()."

Which package is it loading? What is the purpose of simulating "what happens when a package is installed and loaded" at the beginning of an analysis process? If the/a package needed to be loaded, why not load it with `library()` here?

parameter_files/WebParameters_2dii.yml
```diff
@@ -1,8 +1,8 @@
 default:
     paths:
-        project_location_ext: /Users/jacobkastl/Documents/git/PACTA_analysis/
-        data_location_ext: /Users/jacobkastl/Documents/git/pacta-data/2019Q4/
-        template_location: /Users/jacobkastl/Documents/git/create_interactive_report/
+        project_location_ext: /home/mauro/git/PACTA_analysis/
+        data_location_ext: /home/mauro/git/pacta-data/2019Q4/
+        template_location: /home/mauro/git/create_interactive_report/
```
cjyetman, line 5: can we use `~/git/PACTA_analysis/` etc. here so that it's portable?

R/read_file.R
```diff
@@ -0,0 +1,80 @@
...
+#' read_file(path)
+#'
+#' # More natural case: .RData saved as .rda
+#' data2 <- data.frame(x = 1)
+#' save(data2, file = path)
+#' read_file(path)
+#'
+#' # Fails by design: .RData with more than 1 object
+#'
+#' save(data1, data2, file = path)
```
cjyetman, line 21: I believe the first argument of `save()` should be a list of objects, rather than `...`

R/read_file.R
maurolepore: I'm influenced by "Clean code" by Robert C. Martin. The name of the function acts as a comment to explain what the function does, but hides the implementation details to make it easier to read and easier to maintain. The high-level function does not need to know the implementation details of the lower level functions; this ensures that when the low level details change, the high level functions remain the same.

If the function does what its name promises, the reader can understand it from the overview that the high level function provides. When the reader wants to know the details, they can jump to the definition of the function. The code reads like a magazine article, with an overview at the top (high level function) and details becoming finer grained as you descend.

R/read_file.R
maurolepore: :+1: That's what `abort_if_too_long(dataset)` does (even though we disagree in whether or not we should expose or abstract the details -- but that's very welcome professional criticism).

image

0_portfolio_input_check_functions.R
maurolepore: The PACTA_analysis repo contains an R package (see DESCRIPTION). Here, `devtools::load_all()` sources all R files in R/ and all datasets in data/. It is using the package in "developer mode" instead of "user mode" with `library()`; this is a hack (and likely a bad idea) to avoid the more verbose approach:

```r
source("R/some-file.R")
source("R/another-file.R")
load("data/some-data.rda")
load("data/some-more-data.rda")
```

maurolepore: Eventually, we may extract the package elsewhere and use it with `library()`. But for now, I find it convenient to have the R/ functions alongside the PACTA_analysis code so git tracks the refactoring moves I do.

R/read_file.R
maurolepore: From `?load()` and this experiment, I think both are equivalent:

``` r
data1 <- data2 <- data.frame(x = 1)

path1 <- tempfile(fileext = ".rda")
save(data1, data2, file = path1)

path2 <- tempfile(fileext = ".rda")
save(list = c("data1", "data2"), file = path2)

e1 <- new.env()
base::load(path1, envir = e1)

e2 <- new.env()
base::load(path2, envir = e2)

identical(ls(e1), ls(e2))
#> [1] TRUE
```

<sup>Created on 2020-10-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

0_portfolio_input_check_functions.R
cjyetman: OK, I think I understand it now. :-)

But this raises the question for me then, why are this file `0_portfolio_input_check_functions.R` and similar files like `0_global_functions.R` not also in the `./R` sub-directory?

parameter_files/WebParameters_2dii.yml
maurolepore: The path from home to PACTA_analysis/ and its siblings is generated problematically, in integration-test.Rmd (see `[make_paths_portable()](https://github.com/2DegreesInvesting/PACTA_analysis/blob/current_web_functionality/integration-test.md#configurations)`). This is a horrible hack I came up with to deal with the inflexible way the configuration file is managed. I'm happy to improve that in a separate PR.

R/read_file.R
cjyetman: oops....

from the help file...
```
save(..., list = character(),
```

```
...  the names of the objects to be saved (as symbols or character strings).
list  A character vector containing the names of objects to be saved.
```

0_global_functions.R
maurolepore: I didn't work on this function but it seems similar to the one I did work on. I'll open an issue to follow up. Hopefully we can reuse some of the code of this PR.

In any case, I'm now sure if it pays of to maintain this wrappers around functions that read and save files. The "convenience" they add comes at the cost of having to maintain too many lines of code and at the risk of introducing lots of bugs. 

I'm not attached to this PR and I would love to kill `read_file()` if and whenever it is safe.

maurolepore:
Thanks a lot @cjyetman for your comprehensive review. I'll pushed the commits that addressed your suggestions, or resolved the review-conversations with an explanation of why I didn't do what you suggest.  If you have super strong opinions about the changes I didn't do, let me know and I can reconsider.

I'll pass it on to @jacobvjk before merging.

0_portfolio_input_check_functions.R
maurolepore: Great question. 
I move code to R/ as I touch it, and see an opportunity to extract something that stands alone, or need to test something (like `read_file()`) -- using the convenient paring between R/file.R and tests/testthat/test-file.R that  works automatically with testtaht.

cjyetman: So then, would you suggest/support the idea, at least in the long run, of moving scripts like `0_global_functions.R` and/or the functions contained in that script to the `R/` sun-directory?

Just to understand if that is the long-term strategy... because for me I think that was part of the confusion, we are dependent here now on some R-package like structure, functions, tools, etc., but then much of this repo doesn't look anything like an R package to me (not even sure if it would successfully build unless all of these extraneous files were added to `.Rbuildignore`), as you say it looks more like a data science pipeline.

maurolepore: > would you suggest/support the idea, at least in the long run, of moving scripts like 0_global_functions.R and/or the functions contained in that script to the R/ sun-directory?

Yes, I think that's one good strategy to refactor this code base. To me, it makes sense because I don't want to change any existing behavior unless I can test its impact. And the structure R/ + tests/testthat makes testing functions easy. I wouldn't go crazy moving all functions; instead I would move functions as we need to work on them (as we did in this in this PR).

> we are dependent here now on some R-package like structure, functions, tools, etc.

Yes, but only minimally and for convenience. The scripts don't _need_ to know that R/ has functions that also belong to a package -- i.e. we could remove the call to `devtools::load_all()` and replace it with calls to `source("R/....R")` and `read_file("data/....")`. And even if we do that, the package structure is convenient because I can test all functions with a keyboard shortcut.

> not even sure if it would successfully build unless all of these extraneous files were added to .Rbuildignore)

Yup, you guessed it -- all these extraneous files are in .Rbuildignore. R CMD check passes cleanly, and I plan to add a gh-actions workflow to check on every push.

maurolepore:
@jacobvjk, please review at your earliest convenience.

PR #134, opened 2020-10-07
Add gh-actions workflow to run R CMD check

maurolepore:


maurolepore:
This PR is safe in that it touches no code. It only adds a workflow to run R CMD check on GitHub actions (which also runs all tests in tests/testthat).

The workflow file is what you get with `usethis::use_github_action_check_full()`, except that I removed checks for versions of R we don't support (we support R >= 3.5).

.github/workflows/R-CMD-check.yaml
```diff
@@ -0,0 +1,102 @@
+# NOTE: This workflow is overkill for most R packages
+# check-standard.yaml is likely a better choice
+# usethis::use_github_action("check-standard") will install it.
```
cjyetman, line 3: I don't know anything about this, but it sounds like we should at least consider the "check-standard" version of this.

.github/workflows/R-CMD-check.yaml
```diff
@@ -0,0 +1,102 @@
...
+#
+# For help debugging build failures open an issue on the RStudio community with the 'github-actions' tag.
+# https://community.rstudio.com/new-topic?category=Package%20development&tags=github-actions
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
cjyetman, line 13: now that `current_web_functionality` is the default branch, does that mean that here "master" refers to `current_web_functionality`, or does it actually refer to the `master` branch? and is that what is intended?

cjyetman:
I made a couple inline comments. Otherwise I'm not really qualified to review this. ;-)

.github/workflows/R-CMD-check.yaml
maurolepore: Great catch! Thanks! 
As is, I think this would be watching only on pushes and PR with the base branch `master`. I could change it to `current_web_functionality` but I instead I'll wait until we merge #136 -- which should be soon. Then `master` will be the correct base branch.

.github/workflows/R-CMD-check.yaml
maurolepore: Thanks for promoting simplicity. Unfortunately the [check-standard.yml](https://github.com/r-lib/actions/blob/master/examples/check-standard.yaml) checks only with R-release and R-devel, but this repo supports R >= 3.5. Our other workflow ([run-integration-test.yaml](https://github.com/2DegreesInvesting/PACTA_analysis/blob/current_web_functionality/.github/workflows/run-integration-test.yaml)) also checks the full range from R 3.5 to R-devel.

The complexity doesn't worry me; this workflow is well maintained and I follow it closely because I use it in all the packages we have on CRAN.

maurolepore:
Will merge after #136

PR #136, opened 2020-10-07
Re-add three commits I reverted recently

### LINKED ISSUE #92: Can we use a single main line of development?
I moved this discussion from #90 to here.

> We should discuss the webtool scripts and branch. It seems that merging the branch this is the better thing to do.

-- @Clare2D 

> regarding the branch/repo discussion: would be great to talk about this this week, as we should also consider how the stress testing modules will fit into the future repo structure and this should ideally already be considered when I make the changes in that for P2020

-- @jacobvjk

### LINKED ISSUE #120: master or current_web_functionality
After commit acdedabadccdcdeb05cc2e0384f9c78e77a5abd6, `current_web_functionality` is still set as the default branch of the repo, and based on the name of that commit ("Merge branch 'master' into current_web_functionality") that seems like it should be the case. Or not? Should we be primarily using "master" now, or "current_web_functionality"?

maurolepore:
Closes #120 
Closes #92 

This PR re-adds the changes that I recently reverted. It merges the branch
`current_web_functionality` into master. This is not clear from seeing this:

 

(Sorry that the name `re-add-two-reverted-commits` is a bad name -- as this PR actually reverts not two but three commits.)

However, you can confirm by merging locally this PR into master. You should 
see "Already up to date" (assuming your local `master` is in syn with the 
remote `master` in this repo). This means that once we merge this PR we can 
change the default branch from `current_web_functionality` to `master`.

DETAILS

These two commits reverted without conflicts:

* 9ffb854 - Revert "Moved file call into loop to save processing time"
* 3fdbd94 - Revert "Merge branch 'master' into current_web_functionality" `**`

This commit reverted with conflicts, but solving them was straight forward:

* 2472408 (to-revert) Revert "Changed function name read_in_file and use of this"

I kept the name `read_file()`, and flagged a TODO and FIXME.

Also, this PR adds three files to .gitignore to pass R CMD check. It's a minor change
that I should have added elsewhere but here it is and I can't be bothered :)


.Rbuildignore
```diff
@@ -44,3 +44,6 @@
 ^LICENSE\.md$
 ^docker\.sh$
 ^dry$
+^README\.Rmd$
+^pacta_data_qa\.Rmd$
+^project_settings\.yml$
```
maurolepore, line 49: This change does not come from reverting commits.

0_portfolio_input_check_functions.R
```diff
@@ -76,215 +76,216 @@ clean_portfolio_col_types <- function(portfolio, grouping_variables){
...
-  
+
   fin_data <- fin_data %>% bind_rows(fin_data_na)
-  
+
   fin_data <- fin_data %>% select(-security_mapped_sector) %>% rename(security_mapped_sector = sector)
 
-  fin_data %>% group_by(security_mapped_sector) %>% filter(is.na(security_mapped_sector)) %>% summarise(count = n())
+  # FIXME: This seems to serve no purpose as it's not assigned to any variable
+  fin_data %>% group_by(security_mapped_sector) %>% filter(is.na(security_mapped_sector)) %>% summarise(count = n(), .groups = "drop")
```
maurolepore, line 225: Should we remove this line or should we assign it to `fin_data`?

0_portfolio_input_check_functions.R
```diff
@@ -342,74 +343,75 @@ check_mapped_assets_flag <- function(fin_data){
...
+
   actual_sectors <- unique(fin_data$security_mapped_sector)
-  if(any(!actual_sectors %in% c(sector_list, other_sector_list, "Other"))){
-    stop("Additional Sectors in fin_data")
-  }
 
+  ## TODO: Remove dead code
+  # if(any(!actual_sectors %in% c(sector_list, other_sector_list, "Other"))){
+  #   stop("Additional Sectors in fin_data")
+  # }
```
maurolepore, line 381: We should remove this, right?

0_portfolio_input_check_functions.R
maurolepore: Also, `.groups = "drop"` seems to be added recently. We want it, right?

maurolepore (commented):
@jacobvjk, I wrote a few comments to drive your attention to some things I'd like you to please check.

0_portfolio_input_check_functions.R
cjyetman: `.groups = "drop"` I believe is necessary with v1.0+ of `dplyr` to prevent a warning in the console

.Rbuildignore
jacobvjk: noted

jacobvjk (approved):
LGTM

PR #139, opened 2020-10-08
Remove save_file()

### LINKED ISSUE #133: Refactor (or kill) save_file()
https://github.com/2DegreesInvesting/PACTA_analysis/pull/129/files/ff03a383e0601bca66548a013395b78a1a3a7418#r500834517

I label this as enhancement but I suspect a lot of bugs hide in this fuction.

maurolepore:
Closes #133 

This PR makes the code base a little smaller and less error prone by removing 
`save_file()`, which has too much logic for the little value it adds.

### DIFF
--- a/0_global_functions.R
+++ b/0_global_functions.R
@@ -1,6 +1,6 @@
 # Checks whether a variable is a dataframe. Considers also logicals and null values.
 data_check <- function(df){
-  
+
   if (is.data.frame(df)){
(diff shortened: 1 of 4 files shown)


cjyetman:
👍 I’m in favor of this. The only situation where I think this function might be beneficial is if we wanted to maintain saving RDS files with the “.rda” extension, which hopefully we’re moving away from instead.

Issue #137, opened 2020-10-08
Prune dependencies


maurolepore:
https://github.com/2DegreesInvesting/PACTA_analysis/pull/128#discussion_r501291778

Clare2D:
Hi. I think this is relevant here: I'm getting this error in web_script_3
`Either pick the one you want with `::` 
* dplyr::lag
* stats::lag
Or declare a preference with `conflict_prefer()`
* conflict_prefer("lag", "dplyr")
* conflict_prefer("lag", "stats")`

Clare2D:
In addition to this I've already added: 
 ```conflict_prefer("filter", "dplyr") ```

maurolepore:
1. Are you up to date with `master`? 

```bash
git checkout your-branch
# Assuming origin points to 2DegreesInvesting
git fetch origin
git merge origin/master
```
2. Are you sourcing after restart?
3. Are you running on your host computer or on docker?

--

I can't reproduce your problem. I restart and source script 1 successfully, same with script 2 and 3. I expect the conflicts to be
solved via `use_r_packages()` -- which internally calls `resolve_conflicts()`.

In the output of every scripts I see:

```r
[conflicted] Will prefer dplyr::filter over any other package
[conflicted] Will prefer dplyr::lag over any other package
[conflicted] Will prefer dplyr::mutate over any other package
[conflicted] Will prefer here::here over any other package
[conflicted] Will prefer dplyr::rename over any other package
[conflicted] Will prefer dplyr::summarise over any other package
[conflicted] Will prefer dplyr::arrange over any other package
```

PR #140, opened 2020-10-08
Centralize calls to library()

### LINKED ISSUE #91: Replace calls to library() with source("packages.R")
#88 introduced packages.R. We can now replace calls to `library()` with `source("packages.R")`.
This will remove duplication and thus leave the system cleaner.

It would be nice to write a helper that updates packages.R. The ui may be like 
`spelling::update_wordslist()` or `usethis::use_tidy_description()`. Maybe it should
live in r2dii.usethis.

maurolepore:
Closes #91 
Relates to #137 

This PR centralizes all calls to `library()` in a single place. This removes
duplication, makes the code base smaller, and makes it easier to prune
packages because now it is clearer which packages we depend on.


cjyetman:
What is the `dry` directory for? Why is it called that?

I think maybe a more descriptive filename than `packages.R` might be helpful.

Otherwise, I support this plan and I don’t see any obvious errors.

web_tool_script_1.R
```diff
@@ -5,19 +5,7 @@ options(encoding = "UTF-8")
 # devtools::install_github("2DegreesInvesting/r2dii")
 # install.packages("fst")
```
cjyetman, line 6: should these commented out install lines also be removed from here and/or added to `dry/packages.R`?

0_fund_processing.R
```diff
@@ -1,5 +1,4 @@
-library(tidyverse)
```
jacobvjk, line 1: @Clare2D this file is not needed for anything, right?

jacobvjk (commented):
LGTM
from my side, as long as CJ's question can be answered

web_tool_script_1.R
maurolepore: I'll submit a separate PR to remove dead code like this one.

maurolepore: https://github.com/2DegreesInvesting/PACTA_analysis/issues/144

.Rbuildignore
```diff
@@ -43,4 +43,4 @@
 ^data-raw$
 ^LICENSE\.md$
 ^docker\.sh$
-^dry$
+^deduplicate$
```
maurolepore, line 46: DRY is short for Don't Repeat Yourself (https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). 
I removed the jargon, favoring a more explicit name -- thanks CJ.

0_fund_processing.R
maurolepore: If so, maybe flag this in #144

R/utils.R
```diff
@@ -1,12 +1,18 @@
...
+  path <- fs::path(
+    "deduplicate",
+    "set_portfolio-name-ref-all_working-location_and_web-parameters.R"
+  )
+  source(path)
+}
+
+use_r_packages <- function(path = "deduplicate/load-and-attach-r-packages.R") {
+  source(path)
 }
```
maurolepore, line 18: This helper hides the details about where and how we load and attach packages -- which are a low level detail that is likely to change. This way the change will happen in just one place.

web_tool_script_2.R
```diff
@@ -1,34 +1,22 @@
+options(r2dii_config = paste0(par_file_path,"/AnalysisParameters.yml"))
+
+devtools::load_all()
+use_r_packages()
```
maurolepore, line 4: I reorganized scripts headers to start with options, then sourcing all functions and data, then attaching packages, then anything else.

maurolepore (commented):
Here are some comments to highlight the most interesting changes.

.Rbuildignore
cjyetman: hahahaha… awesome, learn something new every day 😉

PR #141, opened 2020-10-08
Style .R and .Rmd files

maurolepore:
This PR styles R code systematically. Using a consistent style makes
the code easier to read. Also, the styler removes whitespace that 
makes git diff harder to understand -- but see CJ's recent discovery:


cjyetman:
👍 👍 👍  thank you!

I’m not actually reviewing the changes, because it’s so much and I assume you used some automated process for that.

One suggestion, in a broad sense, not just in this repo, is to check/set/be aware of the settings in the `*.Rproj` file which can enforce some coding style issues, like stripping white space from line endings. It looks like this repo is already setup properly like that, and I guess that’s why this is happening now, because RStudio is enforcing those rules which apparently were not in effect when the files were first created.

cjyetman:
For clarification of my comment above, I believe the *project level* settings, which are stored in the `*.Rproj` file, which is then typically included in the git repo, is ideal because then those settings are shared across any forks… as opposed to relying on individual user settings which are not necessarily in sync.

The short version of all this is… I support the inclusion of a `*.Rproj` file in the repo to help sync up and enforce some basic level code style choices. 😉

maurolepore:
Thanks CJ, 

Indeed, this PR was created with the something like:

```r
files <- fs::dir_ls(regexp = "[.]R$", recurse = TRUE)
purrr::walk(files, styler::style_file)
```

Indeed, it looks like [my rstudio preferences](https://github.com/maurolepore/dotfiles/blob/master/rstudio-prefs.json) got into the .Rproj file, maybe when I created the package.

I'll also add a gh-actions worflow to style the package-code from any PR-comment -- by starting it with the keyword "/style".

.github/workflows/pr-commands.yaml
```diff
@@ -0,0 +1,55 @@
+on:
+  issue_comment:
+    types: [created]
+name: Commands
+jobs:
+  document:
+    if: startsWith(github.event.comment.body, '/document')
+    name: document
```
maurolepore, line 8: We can now document the package-code (i.e. the code in R/, tests/ and so on) by starting a PR-comment on GitHub with "/document"

.github/workflows/pr-commands.yaml
```diff
@@ -0,0 +1,55 @@
...
+          git config --local user.email "actions@github.com"
+          git config --local user.name "GitHub Actions"
+          git add man/\* NAMESPACE
+          git commit -m 'Document'
+      - uses: r-lib/actions/pr-push@v1
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+  style:
+    if: startsWith(github.event.comment.body, '/style')
+    name: style
```
maurolepore, line 33: We can now style the package-code (i.e. the code in R/, tests/, and so on) by starting a PR-comment on GitHub with "/style"

Issue #122, opened 2020-10-06
shell variables in docker.sh should be quoted


cjyetman:
It is my understanding that when using shell variables that contain strings (as in [docker.sh](https://github.com/2DegreesInvesting/PACTA_analysis/blob/current_web_functionality/docker.sh)), one should always quote them to avoid problems, e.g. a path with spaces in it.

for example...
`data=$(dirname $(pwd))/pacta-data`
should be...
`data=$(dirname "$(pwd)")/pacta-data`

and...
```
docker run --rm -ti \
  --mount type=bind,source=$analysis,target=/bound
```
should be...
```
docker run --rm -ti \
  --mount type=bind,source="$analysis",target=/bound
```

maurolepore:
Solved

PR #142, opened 2020-10-08
Run the web tool on docker in a single step, or in two steps

### LINKED ISSUE #121: docker.sh should not be run as a shell script
The [docker.sh](https://github.com/2DegreesInvesting/PACTA_analysis/blob/current_web_functionality/docker.sh) seems like it is intended to be a shell script which can be run directly from the shell command line, but it seems like the second half of the script is intended to be run inside of the docker container which is initiated by the first half of the script. If the script is fully run from a shell command line, then the docker container is initiated, but the Rscript commands are run in the host/local environment, which seems wrong and also potentially mucks up the hosts R environment, e.g. setting your packages repo to a Linux specific repo that may not be appropriate for the host machine.

I suggest separating these into two scripts with clear instructions in the comments about how and where to run them.

maurolepore:
Closes #121 

This PR adds two workflows to run the web tool on docker. 

[Instructions](https://github.com/2DegreesInvesting/PACTA_analysis/tree/121-docker/bin)


bin/start-containter-with
```diff
@@ -0,0 +1,16 @@
...
+analysis=$(pwd)
+data=$(dirname $(pwd))/pacta-data
+report=$(dirname $(pwd))/create_interactive_report
+stresstest=$(dirname $(pwd))/StressTestingModelDev
+
+docker run --rm -ti \
+  --mount type=bind,source=$analysis,target=/bound \
+  --mount type=bind,source=$data,target=/pacta-data \
+  --mount type=bind,source=$report,target=/create_interactive_report \
+  --mount type=bind,source=$stresstest,target=/StressTestingModelDev \
```
cjyetman, line 14: see issue #122 shell variables should be quoted

bin/README.md
```diff
@@ -0,0 +1,16 @@
+## Running the web tool on transitionmonitor's docker-container
+
+Ensure your working directory is PACTA_analysis, then run one of these workflows:
```
cjyetman, line 3: maybe a comment about the Docker image `2diirunner:1.3.0` being necessary?

cjyetman:
Thanks for this effort!

This seems like it should work, but I haven’t tested it myself, as I don’t yet have the git skills to jump between PRs while I’m in the process of making changes on another branch.

bin/README.md
maurolepore: Thanks!

bin/start-containter-with
```diff
@@ -2,15 +2,15 @@
...
 # docker image import https://tool.transitionmonitor.com/downloads/2diirunner_1.3.0.tar.xz 2diirunner:1.3.0
 
-analysis=$(pwd)
-data=$(dirname $(pwd))/pacta-data
-report=$(dirname $(pwd))/create_interactive_report
-stresstest=$(dirname $(pwd))/StressTestingModelDev
+analysis="$(pwd)"
+data="$(dirname $(pwd))/pacta-data"
+report="$(dirname $(pwd))/create_interactive_report"
+stresstest="$(dirname $(pwd))/StressTestingModelDev"
```
cjyetman, line 8: while the quotes in `analysis="$(pwd)”` should be harmless, the more important place to put them is when you access the variable within other stuff you’re combining it with, for instance….
`data="$(dirname $(pwd))/pacta-data”`
should be
`data=$(dirname "$(pwd)")/pacta-data`

otherwise, for instance, on my machine which has a space in the working directory path, the `$(pwd)` part that gets added to the string that makes the `$data` variable only passes the path up to the first space

bin/start-containter-with
```diff
@@ -2,15 +2,15 @@
...
 
 docker run --rm -ti \
   --mount type=bind,source=$analysis,target=/bound \
   --mount type=bind,source=$data,target=/pacta-data \
   --mount type=bind,source=$report,target=/create_interactive_report \
   --mount type=bind,source=$stresstest,target=/StressTestingModelDev \
```
cjyetman: these lines should also quote the shell variable strings…
```
  --mount type=bind,source="$analysis",target=/bound \
  --mount type=bind,source="$data,target=/pacta-data \
  --mount type=bind,source="$report",target=/create_interactive_report \
  --mount type=bind,source="$stresstest",target=/StressTestingModelDev \
```

bin/run-web-tool
```diff
@@ -1,3 +1,3 @@
 #! /bin/bash
 
-$(pwd)/bin/start-containter-with "/bound/bin/run-r-scripts"
+"$(pwd)/bin/start-containter-with" /bound/bin/run-r-scripts
```
cjyetman, line 3: Pretty sure this should be
`"$(pwd)"/bin/start-containter-with /bound/bin/run-r-scripts`
or if you want to prepare for the future possibility of changing the sub-directory path to something else that also might have a space in it
`""$(pwd)"/bin/start-containter-with" /bound/bin/run-r-scripts`

bin/start-containter-with
maurolepore: I don't get why we would quote only the inner variable and not the outer one:

```bash
$(dirname "$(pwd)")
$(outer_variable "$(inner_variable)")
```

In [this example]( https://www.jimhester.com/post/2017-10-13-docker/) Jim
Hester quotes the entire expression `"/opt/$(basename $(pwd))"`

```bash
docker run -v "$(pwd)":"/opt/$(basename $(pwd))" ...
```

This is more for me to learn what's going on than a question about what to do -- which I don't mind either way.

bin/start-containter-with
maurolepore: Oops, thanks!

bin/start-containter-with
maurolepore: Oh, okay, I think I get it now. Thanks.

bin/start-containter-with
```diff
@@ -0,0 +1,16 @@
...
+data=$(dirname "$(pwd)")/pacta-data
+report=$(dirname "$(pwd)")/create_interactive_report
+stresstest=$(dirname "$(pwd)")/StressTestingModelDev
+
+docker run --rm -ti \
+  --mount type=bind,source="$analysis",target=/bound \
+  --mount type=bind,source="$data",target=/pacta-data \
+  --mount type=bind,source="$report",target=/create_interactive_report \
+  --mount type=bind,source="$stresstest",target=/StressTestingModelDev \
+  2diirunner:1.3.0 "$1"
```
maurolepore, line 15: @cjyetman, looking good now?

bin/start-containter-with
cjyetman: it’s primarily a concern when a shell variable is expanded… that is when, if it has a space in it, often only the chunk up until the first space is expanded, so…

say you have (in pseudo-code)…
```
foo = "/path/with a/space/"
bar = "/home/$foo"

bar will return
/home/path/with

foo = "/path/with a/space"
bar = "/home/"$foo""

bar will return
/home/path/with a space/
```

cjyetman: that’s actually a pretty bad example

maybe a better explanation is that in `data="$(dirname $(pwd))/pacta-data”`, `dirname $(pwd)` is a complete command of which its output is being used as a variable

so for example…
```
root@bd0b8a010eba:/bound# foo="/path with/a/space/"
root@bd0b8a010eba:/bound# echo $foo
/path with/a/space/
root@bd0b8a010eba:/bound# dirname $foo
/
with/a
root@bd0b8a010eba:/bound# dirname "$foo"
/path with/a
root@bd0b8a010eba:/bound# 
```

Issue #135, opened 2020-10-07
Improved assignment of config objects to global environment (via sourcing dry/file.R)


jacobvjk:
As described by @maurolepore in this past PR, there is a better way to load configs into the global environment than what is current PACTA_analysis practice. Once we merge the 'current_web_functionality' into the 'main' branch, this would be a good opportunity to change that across the repo

https://github.com/2DegreesInvesting/PACTA_analysis/pull/114

maurolepore:
I edited the title slightly to not prescribe the implementation details. The best way should be clearer once we dive into this issue and write some tests.

maurolepore:
https://github.com/2DegreesInvesting/PACTA_analysis/pull/155/commits/4aaaa4b1d59ddcac147048a5bdf1189eca9d70cf#r502613979

PR #155, opened 2020-10-09
148 par file path

maurolepore:


web_tool_script_2.R
```diff
@@ -9,12 +9,15 @@ source("0_portfolio_test.R")
...
 
+devtools::load_all()
 setup_project()
 
 working_location <- paste0(working_location, "/")
 
 set_webtool_paths()
 
+# just done once
+options(r2dii_config = paste0(par_file_path,"/AnalysisParameters.yml"))
```
maurolepore: For the record, this is a good example to show why #135  is important. 

Here, the variable `par_file_path` is available "magically": It is not visible where it is defined. 
This code should not work, but it does because `set_webtool_paths()` defines `par_file_path` -- beyond
its own scope -- on the global environment. This use of `<<-` is a very bad idea. 

```r
set_webtool_paths <- function() {
  project_location <<- paste0(working_location, "working_dir/")

  log_path <<- paste0(project_location, "00_Log_Files")
  par_file_path <<- paste0(project_location, "10_Parameter_File")
  raw_input_path <<- paste0(project_location, "20_Raw_Inputs")
  proc_input_path <<- paste0(project_location, "30_Processed_Inputs")
  results_path <<- paste0(project_location, "40_Results")
  outputs_path <<- paste0(project_location, "50_Outputs")
}
```

PR #156, opened 2020-10-09
Centralize calls to library() (#140)

maurolepore:
This commits somehow made its way into a PR where it's not relevant. 
I'm pulling it here to keep the other PR focused.


maurolepore:
Closes #91 
Relates to #137 

This PR centralizes all calls to `library()` in a single place. This removes
duplication, makes the code base smaller, and makes it easier to prune
packages because now it is clearer which packages we depend on.

PR #163, opened 2020-10-10
Detect and use more packages

maurolepore:
When running the web-tool on docker I noticed a few packages
were missing.


jacobvjk (approved):
lgtm

PR #164, opened 2020-10-10
Extend the base docker image to include required R packages

maurolepore:
Here is how to update all repos and run the web tool: 

<https://asciinema.org/a/364647>

I published the docker image that includes the required R packages
so that we don't need to re-install required R packages every time
we run the web-tool on docker. This should allow us to iterate and
detect bugs more quickly.

You can pull it from its dockerhub repo with:

```bash
docker pull maurolepore/2diirunner-with-packages
```

For now I published it in my own dockerhub account as
[maurolepore/2diirunner-with-packages](https://hub.docker.com/r/maurolepore/2diirunner-with-packages). We can
later move it to a 2DII account. 

This is quite experimental and I would appreciate your help before
we merge this PR.


R/utils.R
```diff
@@ -13,6 +13,42 @@ setup_project <- function() {
...
+
+  out <- c(
+    dkr[1:marker_start(dkr)],
+    pkg,
+    dkr[marker_end(dkr):length(dkr)]
+  )
+  writeLines(out, path)
+
+  invisible(path)
+}
```
maurolepore, line 45: This function helps update the calls to `Rscript -e "install.package(...)" in the Dockerfile. It overwrites the existing
Dockerfile with packages detected in "deduplicate/load-and-attach-r-packages.R".

docker/2diirunner-with-packages/README.md
```diff
@@ -0,0 +1,13 @@
...
+docker image import https://tool.transitionmonitor.com/downloads/2diirunner_1.3.0.tar.xz 2diirunner:1.3.0
+```
+
+To build this image locally, change directory to this folder, then run:
+
+```bash
+docker build --tag 2diirunner-with-packages .
+```
+
+Reference: <https://ropenscilabs.github.io/r-docker-tutorial/04-Dockerhub.html>
```
maurolepore, line 13: This file explains how I created this image and how to use it to run the web tool

docker/2diirunner-with-packages/Dockerfile
```diff
@@ -0,0 +1,55 @@
+FROM maurolepore/2diirunner:1.3.0
```
maurolepore, line 1: This Dockerfile will trigger a build on dockerhub with every push -- now configured only to the branch `docker`, but eventually it should be configured with `master`.

maurolepore: Likely `FROM maurolepore/2diirunner:1.3.0` should be `FROM maurolepore/2diirunner:latest`, so we don't have to rewrite this Dockerfile when transitionmonitor increment the version number.

docker/run-r-scripts
```diff
@@ -0,0 +1,13 @@
...
+cd /bound
+
+portfolio=${1:-TestPortfolio_Input}
+# FIXME: Eventually we want to run the Rscripts directly, but it's easier
+# to detect problems if you run them via integration-test.Rmd because some of
+# its tests should allow you to narrow down a bit the cause of the problem
+Rscript -e "rmarkdown::render('integration-test.Rmd')"
+# Rscript --vanilla web_tool_script_1.R portfolio
+# Rscript --vanilla web_tool_script_2.R portfolio
+# Rscript --vanilla web_tool_script_3.R portfolio
```
maurolepore, line 13: This should help debug.

docker/start-containter-with
```diff
@@ -12,5 +10,5 @@ docker run --rm -ti \
   --mount type=bind,source="$data",target=/pacta-data \
   --mount type=bind,source="$report",target=/create_interactive_report \
   --mount type=bind,source="$stresstest",target=/StressTestingModelDev \
-  2diirunner:1.3.0 "$1"
+  maurolepore/2diirunner-with-packages $1
```
maurolepore, line 13: This is the key. Now we run the image that already includes all required R packages.

maurolepore:
@AlexAxthelm, if you happen to have some time, please have a look at what we have here. 
You are our Docker guru and maybe you can quickly suggest improvements.

docker/start-containter-with
maurolepore, line 13: @AlexAxthelm, we now run this ackward script as a terminal command, 
1. Would you use docker-compose instead? Can you draft the yaml file?
2. I haven't yet studied the difference between `--mount` and `-v` (volumes), do you have any opinion to direct my research?

R/imports.R
```diff
@@ -0,0 +1,3 @@
+#' @importFrom glue glue
+#' @importFrom rlang abort
```
maurolepore, line 2: This doesn't relate to this PR -- I should probabbly extract it to a separate PR.

cjyetman:
Honestly, I think this is a bad idea unless we're 100% sure that Constructiva will use this exact Docker image on their web server. There are too many parameters that could change: OS, OS version, R version, packages installed, version of packages installed, etc. We're 2 weeks away from this supposedly going online, and if we start developing this around a Docker image that's totally different than what they're going to use, then we're setting ourselves for disaster I fear.

@jacobvjk do you know what the current status is with getting Constructiva's update Docker image?

R/imports.R
cjyetman: agreed

R/read_file.R
```diff
@@ -39,13 +39,13 @@ read_file <- function(path) {
     "fst" = fst::read_fst(path),
     "rda" = read_rda(path),
     "rds" = readr::read_rds(path),
-    rlang::abort("Extension is supported but something unexpected happened.")
+    abort("Extension is supported but something unexpected happened.")
```
cjyetman, line 42: changes in this file also seem unrelated to this PR

R/utils.R
```diff
@@ -13,6 +13,48 @@ setup_project <- function() {
...
+  conflicted::conflict_prefer("lag", "dplyr")
+
+  invisible(path)
+}
+
+packages_path <- function() {
+  "deduplicate/load-and-attach-r-packages.R"
+}
+
+create_install_calls_for_dockerfile <- function(path = packages_path()) {
```
cjyetman, line 28: seems like this functions purpose is duplicated in the `update_dockerfile_packages` function?

R/utils.R
```diff
@@ -13,6 +13,48 @@ setup_project <- function() {
...
+}
+
+marker_start <- function(lines) {
+  grep("# update-dockerfile-packages-start", lines)
+}
+
+marker_end <- function(lines) {
+  grep("# update-dockerfile-packages-end", lines)
+}
+
```
cjyetman, line 57: I don't understand what these are for

R/utils.R
```diff
@@ -13,6 +13,48 @@ setup_project <- function() {
...
+this_repo <- function() {
+  ifelse(in_transitionmonitor(), "bound", "PACTA_analysis")
```
cjyetman, line 59: also don't understand what this is

bin/remove-outputs
```diff
@@ -0,0 +1,4 @@
+#! /bin/bash
+
+sudo rm working_dir/50_Outputs/TestPortfolio_Input/ -rf
```
cjyetman, line 3: maybe `TestPortfolio_Input` should be parameterized somehow since that's an arbitrary name?

bin/run-r-scripts
```diff
@@ -1,15 +1,14 @@
...
-# Rscript -e "update.packages(ask = FALSE)"
-Rscript -e "install.packages(c('here','writexl','countrycode', 'renv', 'tibble', 'stringr', 'zoo', 'data.table'))"
-# FIXME: TestPortfolio_Input should be a parameter
-Rscript --vanilla web_tool_script_1.R TestPortfolio_Input
-Rscript --vanilla web_tool_script_2.R TestPortfolio_Input
-Rscript --vanilla web_tool_script_3.R TestPortfolio_Input
+# FIXME: Eventually we want to run the Rscripts directly, but it's easier
+# to detect problems if you run them via integration-test.Rmd because some of
+# its tests should allow you to narrow down a bit the cause of the problem
+Rscript -e "rmarkdown::render('preconditions.Rmd', output_file = tempfile())"
```
cjyetman, line 9: If this is what's meant to run on the web server, I strongly believe that it should not waste any time nor risk potential problems by rendering a markdown file

bin/update-repos
```diff
@@ -0,0 +1,20 @@
...
+cd ../create_interactive_report
+git checkout master
+git pull origin master
+
+cd ../StressTestingModelDev
+git checkout master
+git pull origin master
+
+cd ../PACTA_analysis
+git checkout "$current_branch"
```
cjyetman, line 20: not really sure what all this is for, but I hope this is not intended to run on the web server every time a report generation is triggered because I imagine this will be a large bottleneck

docker/2diirunner-with-packages/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+MAINTAINER "Mauro Lepore" maurolepore@gmail.com
+
+RUN apt-get update && apt-get install -y pandoc
+
+# This repository has binaries not just for Windows and Mac, but also for Linux
```
cjyetman, line 6: actually, as far as I understand, this repo has binaries *only* for Linux... i.e. this will not benefit you if you try to use this on macOS or Windows

docker/2diirunner-with-packages/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+RUN echo "options(repos = c(CRAN = 'https://packagemanager.rstudio.com/all/__linux__/bionic/latest'))" >> /root/.Rprofile \
```
cjyetman, line 7: In my experience, just setting the default repo in options isn't enough... you must also configure the HTTP UserAgent properly. You can do that with something like
```
echo 'options(HTTPUserAgent = sprintf("R/%s R (%s)", getRversion(), paste(getRversion(), R.version$platform, R.version$arch, R.version$os)))' >> ~/.Rprofile
```
You can also run RStudio's test script to tell if it is setup properly, like...
```
Rscript -e 'source("https://docs.rstudio.com/rspm/admin/check-user-agent.R")'
```
if it's setup properly, you'll get a message like `Success! Your user agent is correctly configured.`

docker/2diirunner-with-packages/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+    && Rscript -e "install.packages('fst')" \
+    && Rscript -e "install.packages('ggforce')" \
+    && Rscript -e "install.packages('ggmap')" \
+    && Rscript -e "install.packages('ggplot2')" \
+    && Rscript -e "install.packages('ggrepel')" \
+    && Rscript -e "install.packages('ggthemes')" \
+    && Rscript -e "install.packages('glue')" \
+    && Rscript -e "install.packages('grid')" \
+    && Rscript -e "install.packages('gridExtra')" \
+    && Rscript -e "install.packages('here')" \
```
cjyetman, line 27: Not sure, but it might be more efficient to install these packages all in one command. I assume that every time you use `Rscript` a new R console is spun up, which probably has some non-negligible start-up and clean-up overhead

bin/update-repos
cjyetman: I think I understand what this is for now, and in principle, I get it. But it seems a bit dangerous to me to jump into 4 other directories/repos and perform git commands automatically. For example... what if one of these local repos is on a different branch and has uncommitted changes?

docker/README.md
```diff
@@ -0,0 +1,23 @@
...
+* Update local repos to the latest commit on the main branch:
+
+```bash
+./bin/update-repos
+```
+
+* Run the web tool in a single step
+
+```bash
+./bin/run-web-tool
```
cjyetman, line 16: is this an existing file, because I don't see it in this PR?

docker/README.md
```diff
@@ -0,0 +1,23 @@
...
+```
+
+* Or only start the docker container (e.g. to debug)
+
+```bash
+./bin/start-container
```
cjyetman, line 22: likewise, is this an existing file?

integration-test.Rmd
```diff
@@ -23,224 +13,16 @@ knitr::opts_chunk$set(
...
-
-extract_from <- function(x, pattern) {
-  line <- grep(pattern, x, value = TRUE)
-  sub(glue(".*({pattern}.*)"), "\\1", line)
-}
+## Introduction
 
-make_config_portable(config_2)
-```
+This document provides a reproducible example of how to run the so called "web tool". It is based on [these instructions](https://bit.ly/2RCRJn7). It is useful in three ways:
```
cjyetman, line 18: I get "Access Denied" at this bit.ly URL

parameter_files/WebParameters_2dii.yml
```diff
@@ -1,9 +1,9 @@
 default:
     paths:
-        project_location_ext: /Users/jacobkastl/Documents/git/PACTA_analysis/
-        data_location_ext: /Users/jacobkastl/Documents/git/pacta-data/2019Q4/
-        template_location: /Users/jacobkastl/Documents/git/create_interactive_report/
-        stress_test_location: /Users/jacobkastl/Documents/git/StressTestingModelDev/
+        project_location_ext: /home/mauro/git/PACTA_analysis/
+        data_location_ext: /home/mauro/git/pacta-data/2019Q4/
+        template_location: /home/mauro/git/create_interactive_report/
+        stress_test_location: /home/mauro/git/StressTestingModelDev/
```
cjyetman, line 6: there's got to be a better way than repeatedly switching between user specific paths

preconditions.Rmd
```diff
@@ -0,0 +1,228 @@
...
+is_sibling <- function(x) {
+  parent <- path_dir(here())
+  dir_exists(path(parent, x))
+}
+all_siblings <- all(map_lgl(repos, is_sibling))
+
+expect_true(all_siblings)
+```
+
+**NOTE: As of this writing the main line of development is not the standard branch `master` -- it is the branch `current_web_functionality`.**
```
cjyetman, line 119: this is not true anymore, correct?

cjyetman:
FYI... I ran the commands exactly as they are in https://asciinema.org/a/364647 and I ended up with different results and `Execution halted`

docker/2diirunner-with-packages/Dockerfile
AlexAxthelm: Agree with that. My suggestion would be to write a script (`install_packages.R`) that has you dependencies, and call that.

AlexAxthelm:
@maurolepore, I agree with @cjyetman's comment above. I think we should be cautious about how this  image is constructed, and ensure that our contractors are using the same images we are. Maybe we can chat  today about your deployment process and environment, and I can advise once I have a better understanding of how everything fits together?

jacobvjk:
@cjyetman @maurolepore I wrote another email to contructiva this morning, to get more info on the state of the dockerfile, but haven't heard from them yet. Will keep pestering them.

R/utils.R
maurolepore: These are markers that the function `update_dockerfile_packages()` looks for in the Dockerfile -- so it knows what to update.
Thanks for noting it's unclear.

bin/remove-outputs
maurolepore: Agree (https://github.com/2DegreesInvesting/PACTA_analysis/issues/176)

bin/run-r-scripts
maurolepore: I'm not sure what the problem is with rendering markdown, but it it is an issue we could `knitr::purl()` the .Rmd file to convert it to an .R file, then `source()` it.

In any case, this is likely temporary, just to help us debug.

bin/update-repos
maurolepore: Yeah, it's probabbly going to cause more problems than solutions. And a better approach I think would be to create a temporary directory, and clone the repo -- but that would take to long for the quick iteration I like for testing.

I'll just remove this script.

bin/run-r-scripts
cjyetman: running just that command on my machine takes 15 second or so to run, so it's not conducive to make the web server feel quick and responsive

docker/README.md
maurolepore: See https://github.com/2DegreesInvesting/PACTA_analysis/blob/docker/docker/README.md

docker/README.md
maurolepore: See https://github.com/2DegreesInvesting/PACTA_analysis/blob/docker/docker/2diirunner-with-packages/README.md

integration-test.Rmd
maurolepore: @jacobvjk, can you grant access to @cjyetman to the google doc with the instructions? Maybe make it public?

docker/README.md
cjyetman: `./bin/start-container` is not mentioned in the README file you point to

in any case, I see that file exists in master now, so it must have been added at some point before I had last updated

integration-test.Rmd
jacobvjk: @cjyetman @maurolepore you should now both have access

parameter_files/WebParameters_2dii.yml
maurolepore: I agree.

(BTW, What you see here happens automatically, but it shouldn't be here -- I forgot to add this file to my list of `--assume-uncahged` files.)

preconditions.Rmd
maurolepore: Correct. I'll fix it.

docker/2diirunner-with-packages/Dockerfile
maurolepore: I think the container runs on Linux, even if you run docker on a mac or windows.

docker/2diirunner-with-packages/Dockerfile
maurolepore: @AlexAxthelm, what's a good way to pass such a file install_packages.R to the container? For now I fixed this in an inelegant way like so:

```r
  ...
  && Rscript -e "install.packages(c("pkg1", "pkg2", ...))"

**Comment by maurolepore** [comment]
Thanks a lot @cjyetman and @AlexAxthelm!

@AlexAxthelm, I would take your offer for a chat but later. I believe @jacobvjk got access to the upstream Dockerfile and that should make this PR mostly needless. In the meantime, I learn myself how the pieces of this project fit together. I'm pretty new to this project and it's still a bit fuzzy.

File: docker/2diirunner-with-packages/Dockerfile (hunk 14 above)
Comment (cjyetman) [reply docker/2diirunner-with-packages/Dockerfile:6]: I was talking about the comment about RStudio's repository... their repository serves Linux binaries, but as far as I can tell it does not server Mac or Windows binaries.

File: parameter_files/WebParameters_2dii.yml (hunk 20 above)
Comment (cjyetman) [reply parameter_files/WebParameters_2dii.yml:6]: I managed to work around this problem by using relative paths on my local clone, for instance...
```
        project_location_ext: ../PACTA_analysis/
        data_location_ext: ../pacta-data/2019Q4/
        template_location: ../create_interactive_report/
        stress_test_location: ../StressTestingModelDev/
```

File: R/read_file.R (hunk 7 above)
Comment (maurolepore) [reply R/read_file.R:42]: https://github.com/2DegreesInvesting/PACTA_analysis/pull/178

File: R/imports.R (hunk 6 above)
Comment (maurolepore) [reply R/imports.R:2]: https://github.com/2DegreesInvesting/PACTA_analysis/pull/178

**Comment by maurolepore** [comment]
Most of the reviewers's comments translated to some change in this PR. I take it out of draft mode, but before we merge I would like to consider the Dockerfile from transition monitor.

File: parameter_files/WebParameters_2dii.yml (hunk 20 above)
Comment (maurolepore) [reply parameter_files/WebParameters_2dii.yml:6]: That's a great idea.

File: parameter_files/WebParameters_2dii.yml [hunk 22]
```diff
@@ -1,9 +1,9 @@
 default:
     paths:
-        project_location_ext: /Users/jacobkastl/Documents/git/PACTA_analysis/
-        data_location_ext: /Users/jacobkastl/Documents/git/pacta-data/2019Q4/
-        template_location: /Users/jacobkastl/Documents/git/create_interactive_report/
-        stress_test_location: /Users/jacobkastl/Documents/git/StressTestingModelDev/
+        project_location_ext: ../PACTA_analysis/
+        data_location_ext: ../pacta-data/2019Q4/
+        template_location: ../create_interactive_report/
+        stress_test_location: ../StressTestingModelDev/
```
maurolepore, line 6: Thanks @cjyetman

maurolepore:
I'm moving back to draft. I need to sync against master and develop a way to use the upstream Dockerfile from transition monitor.

maurolepore:
Follow up at #198

PR #166, opened 2020-10-12
don't change encoding option because it's dangerous and unpredictable

cjyetman:


cjyetman:
seems like something is wrong with the run-integration-test action... it always fails at:
```
    GITHUB_PAT: ***
Error: Input required and not supplied: token
```

maurolepore:
Yeah, the secrets are configured in such a way that the integration test fails if your PR comes from a fork. I couldn't find a solution so far. Until then, you may submit directly to 2DII -- not from your fork.

PR #171, opened 2020-10-12
squelch another summarise warning

cjyetman:


jacobvjk (approved):
lgtm

maurolepore (approved):
I'm not familiar with this argument but it seems you do :)

cjyetman:
`.groups` is a new, experimental argument for `dplyr::summarise` introduced in v1.0.0. It enables control of what happens to the tibble's groups after a summarise command does its magic. Currently (at least v1.0.0-v1.0.2), the default is `.groups = NULL` which changes its behavior based on the number of rows in each group after summarise, plus it always prints a warning message. If you set `.groups` explicitly, the warning message is suppressed. `.groups = "drop_last"` is described as "dropping the last level of grouping. This was the only supported option before version 1.0.0.", so it is equivalent to the behavior from versions of `dplyr` prior to v1.0.0, which is what/when most of these calls were written, so I presume it's safest to maintain the prior behavior (though I support in the future taking into account how this new argument can/should be used properly to our advantage.) 

https://dplyr.tidyverse.org/reference/summarise.html

PR #173, opened 2020-10-12
don't return unnecessary list

cjyetman:


maurolepore (approved):
LGTM

maurolepore (changes requested):
Oh, locally I see:

``` r
  devtools::load_all()
#> Loading PACTA.analysis
test_integration()
#> processing file: integration-test.Rmd
#> Quitting from lines 205-233 (integration-test.Rmd)
#> Error in sub(glue("({field}:[ ]?).*"), glue("\\1{value}/"), x): invalid 'replacement' argument
```

<sup>Created on 2020-10-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


--

I'll dig a bit to see if I figure out the issue.

parameter_files/WebParameters_2dii.yml
```diff
@@ -1,6 +1,6 @@
 default:
     paths:
-        project_location_ext: ./
+        project_location_ext: ../PACTA_analysis/
```
maurolepore, line 3: Is there a particular reason to use the form `./` instead of the form `../PACTA_analysis/`?
The form "./" breakes the integration test. I could make it work but I'm trying to figure out 
if it's worth the effort.

cjyetman: no, I think either should work.... `../PACTA_analysis/` is just more specific

cjyetman: but, this is in the wrong PR, or?

parameter_files/WebParameters_2dii.yml
```diff
@@ -1,6 +1,7 @@
 default:
     paths:
-        project_location_ext: ./
+        # Prefer "../PACTA_analysis/" -- not "./" nor "/path/to/PACTA_analysis"
+        project_location_ext: ../PACTA_analysis/
```
maurolepore, line 4: The form "./" seems to work on docker but not locally. The form "../PACTA_analysis" works everywhere.

parameter_files/WebParameters_2dii.yml
maurolepore: I don't know how this made it here but it is :)

maurolepore (approved):

maurolepore (approved):
I think we had enough discussion. Thanks!

PR #174, opened 2020-10-12
return the path if it's not found

cjyetman:


maurolepore (approved):
LGTM. Thanks!

maurolepore (changes requested):
Oh, wait -- if we are going to return, should we still `warn()` or should be `abort()`?

cjyetman:
> Oh, wait -- if we are going to return, should we still `warn()` or should be `abort()`?

I'll leave that decision up to you. My intention here was just to get a more useful warning/error message than ``The `path` does not exist: {path}``. I assumed that was the original intention based on the `{variable}` syntax from `glue`? Now you'll get something like ``The `path` does not exist: ../pacta-data/2019Q4/cleand_files/fin_data.fst``

R/read_file.R
```diff
@@ -26,7 +26,7 @@
 read_file <- function(path) {
   # FIXME: A warning seems like a bad idea. It should likely be an error
   if (!fs::file_exists(path)) {
-    rlang::warn("The `path` does not exist: {path}")
+    rlang::warn(glue::glue("The `path` does not exist: {path}"))
```
maurolepore, line 29: > I'll leave that decision up to you. My intention here was just to get a more useful warning/error message than The `path` does not exist: {path}. I assumed that was the original intention based on the {variable} syntax from glue? Now you'll get something like The `path` does not exist: ../pacta-data/2019Q4/cleand_files/fin_data.fst

Ok, then I think it's good and safest as is. Changing this to an error may brake other code, and the warning now is better than before -- meaning that this PR merits to be merged.

maurolepore (approved):

PR #177, opened 2020-10-12
stop using paste0 to manually build file paths

cjyetman:
FYI... there are changes that need to be made upstream in the stress testing and create interactive report repos before this will fully pass

### DIFF
--- a/0_web_functions.R
+++ b/0_web_functions.R
@@ -56,14 +56,14 @@ save_if_exists <- function(df, portfolio_name_, save_name, csv_or_rds = "rds") {
 }
 
 set_webtool_paths <- function() {
-  project_location <<- paste0(working_location, "working_dir/")
+  project_location <<- file.path(working_location, "working_dir")
 
-  log_path <<- paste0(project_location, "00_Log_Files")
-  par_file_path <<- paste0(project_location, "10_Parameter_File")
-  raw_input_path <<- paste0(project_location, "20_Raw_Inputs")
-  proc_input_path <<- paste0(project_location, "30_Processed_Inputs")
-  results_path <<- paste0(project_location, "40_Results")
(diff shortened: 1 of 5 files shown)


cjyetman:
this is also failing the integration test with `Error: Input required and not supplied: token` as in #166

jacobvjk:
> FYI... there are changes that need to be made upstream in the stress testing and create interactive report repos before this will fully pass

Okay, will take a look at it

0_web_functions.R
```diff
@@ -56,14 +56,14 @@ save_if_exists <- function(df, portfolio_name_, save_name, csv_or_rds = "rds") {
...
 set_webtool_paths <- function() {
-  project_location <<- paste0(working_location, "working_dir/")
-
-  log_path <<- paste0(project_location, "00_Log_Files")
-  par_file_path <<- paste0(project_location, "10_Parameter_File")
-  raw_input_path <<- paste0(project_location, "20_Raw_Inputs")
-  proc_input_path <<- paste0(project_location, "30_Processed_Inputs")
-  results_path <<- paste0(project_location, "40_Results")
-  outputs_path <<- paste0(project_location, "50_Outputs")
+  project_location <<- file.path(working_location, "working_dir")
```
maurolepore, line 59: Nit: You may also use `fs::path()`. I don't mind depending on fs -- it's well developed an has 0 dependencies

<https://cran.r-project.org/web/packages/fs/index.html>

maurolepore (approved):
LGTM. Thanks!

0_web_functions.R
cjyetman: I don't know if there's really any advantage to that, but I could redo it tomorrow with `fs::path()`

maurolepore: I think there is no advantage, other than the shorter name.

PR #183, opened 2020-10-12
137 prune dependencies

maurolepore:
@jacobvjk, this removes the need of packages from GitHub -- it may fix the problem 
by which the docker image failed to build (https://github.com/con-wbsch/2diidockerrunner).

I'm not expecting a comprehensive review. For now, removing GitHub packages seems like
a good-enough win.


0_graphing_functions.R
```diff
@@ -3346,7 +3346,7 @@ CarstenMetricChart <- function(plotnumber, chart_type, explicit_filename = "") {
     port <- bind_rows(combin, market)
 
     Type <- "Corporate bond"
-
+    # FIXME: Replace with dplyr::recode() or other options
     port$portfolio_name <- plyr::mapvalues(port$portfolio_name, c(portfolio_name_select, cb_market_ref), c("Portfolio", cb_market_ref))
 
     port$portfolio_name <- factor(port$portfolio_name, levels = c("Portfolio", cb_market_ref), ordered = TRUE)
```
maurolepore, line 3352: This is for another PR-- see  #184

0_sda_approach.R
```diff
@@ -1,6 +1,4 @@
-# sda_approach
-# @author Vincent
-
```
maurolepore, line 3: Git tracks authorship, and the file name says the same than the first comment.

integration-test.Rmd
```diff
@@ -26,7 +26,7 @@ knitr::opts_chunk$set(
 Packages used in this file:
 
 ```{r}
-library(tidyverse)
```
maurolepore, line 29: Remove the tidyverse and instead use each of the required packages explicitely

parameter_files/WebParameters_2dii.yml
```diff
@@ -1,6 +1,6 @@
 default:
     paths:
-        # Prefer "../PACTA_analysis/" -- not "./" nor "/path/to/PACTA_analysis"
+       # Prefer "../this_repo/" -- not "./" nor "/path/to/this_repo"
```
maurolepore, line 3: Using "PACTA_analysis" in the comment was a bad idea because causes an error 
in the integration-test.Rmd. Maybe I should do something about that -- but I don't 
want to over engineer this issue. I'll see if this bites again. For now I leave it at that.

0_graphing_functions.R
Clare2D: Anything in 0_graphing_functions can be ignored as this will be supersceded

jacobvjk (approved):
lgtm

jacobvjk:
@maurolepore to be fair, the stress testing code still loads r2dii.utils and tidyverse. My guess is this should be pruned as well?

PR #188, opened 2020-10-13
only install packages that are not already installed

cjyetman:
in the long run, nothing will need to be installed, but in the short term this saves a bunch of time by not installing packages on top of already existing packages


maurolepore (approved):
LGTM. 
Nit: Why not reusing the container? That should save even more time.?

PR #192, opened 2020-10-13
don't continue when there's a failure

cjyetman:
This will make the docker halt when one of the scripts fails, rather than continuing on with the next one


maurolepore (approved):
Great idea!
LGTM

PR #195, opened 2020-10-13
suppress pkg startup messages

### LINKED ISSUE #187: suppress package startup messages
Can we go back to suppressing package startup messages, please? Especially now that the packages are all loaded in one place, and we're using `conflicted` to sort out function masking problems.

I find it very difficult to sort through the logs when it's filled with messages from so many packages, many of which are not used/needed for a given script, and they all get printed every time `use_r_packages()` is called, so it happens multiple times when you do the web build run.

cjyetman:
this will resolve #187


maurolepore (approved):
LGTM

PR #196, opened 2020-10-13
trim down packages

cjyetman:
only load packages used in this repo, the stress test repo, and the interactive report repo


maurolepore (approved):
Yay!
LGTM

maurolepore (commented):
@cjyetman, I had to solve a merge conflict following your other PR that touched this file.
It was an easy fix but just in case please check this still looks as you expect.

cjyetman:
yup, I made 2 PRs on the same file a couple minutes apart. I felt they were conceptually different, so better to keep separate.

Your merge fixes look about right to me.

PR #197, opened 2020-10-13
Remove read_file()

maurolepore:
This PR removes `read_file()`, which added a lot of code with little value. 
Its most useful feature was the ability to read both .rds and .rda files. 
This feature is preserved in `read_rda()` until we are confident we no
longer have .rds file saved with the misleading extension .rda.

Overall, this PR fixes a but and makes the code base smaller.


R/read_file.R
```diff
@@ -1,64 +1,26 @@
...
 #' data1 <- data.frame(x = 1)
 #' saveRDS(data1, path)
-#' read_file(path)
+#' read_rda(path)
 #'
 #' # More natural case: .RData saved as .rda
 #' data2 <- data.frame(x = 1)
 #' save(data2, file = path)
-#' read_file(path)
+#' read_rda(path)
```
maurolepore, line 16: We keep `read_rda()` -- which is more focused than `read_file()` -- until we ensure no .rds file is saved with the extension .rda.

web_tool_script_1.R
```diff
@@ -57,22 +57,29 @@ if (new_data == TRUE) {
...
+    if (!file.exists(fst_file)) {
+      return(NULL)
+    }
 
-  fin_data <- read_file(file.path(file_location, "fin_data.fst"))
+    fst::read_fst(fst_file)
+  }
+
+  fund_data_path <- file.path(file_location, "fund_data.fst")
+  fund_data <- read_fst_or_return_null(fund_data_path)
```
maurolepore, line 71: I need eyes here. 

In general it is a bad idea for a function to be type-unstable -- sometimes returning one thing (e.g. a dataframe) and sometimes another thing (e.g. `NULL`). But I tolerate this here in the interest of keeping the code running. If instead of `NULL` I throw an error, I brake stuff -- and for now there are other priorities.

maurolepore (commented):
Here is the most interesting stuff.

web_tool_script_1.R
cjyetman: I agree, in both senses.... it's probably not a great idea, but until broader changes are made this is the expected behavior.

R/read_file.R
cjyetman: Nit: Conceptually I  wish that it tried the "right" thing first (read it as an RDA file), and only try the weird thing (read it as an RDS file) just in case that failed, but in this case I guess it makes sense to try to read as RDS first because that's a clear pass or fail operation and there's no way for both reading as RDA and RDS to work on the same file, so it should be safe.

maurolepore: TODO: Happy path first.

R/read_file.R
```diff
@@ -1,64 +1,33 @@
...
-    "rda" = read_rda(path),
-    "rds" = readr::read_rds(path),
-    abort("Extension is supported but something unexpected happened.")
+read_rda2 <- function(file_name) {
+  tryCatch(
+    read_rdata(file_name),
+    # Unfortunately, legacy code saved .rds files with the extension .rda
+    error = function(e) readr::read_rds(file_name)
   )
 }
```
maurolepore, line 30: Happy  path first

R/read_file.R
maurolepore: DONE

maurolepore:
Thanks CJ -- I'll go ahead and merge. @Clare2D, this PR removes `read_ifile()`. I see all tests pass but let me know if you find any issue.

PR #203, opened 2020-10-14
add language and peer group option

Clare2D:


Clare2D:
Now I need to update the create_interactive_report again as I need the language variable as well. Will do so asap.

maurolepore (changes requested):
I solved merge conflicts and flaged FIXME. I think it's safar for the new arguments to go at the end of the signature -- not at the start.

cjyetman:
seems like we're going in circles now... I just merged a PR in the `create_interactive_report` repo to remove `peergroup` from the function's arguments?

cjyetman:
current version of `create_interactive_report ` does not have the `peergroup` nor the `language` argument, so I definitely do not approve of this change
https://github.com/2DegreesInvesting/create_interactive_report/blob/343c6bcee844a165c6ad220107ff9f2eba603d0c/create_interactive_report.R#L1-L29

maurolepore:
This seems incomplete so I changed it to draft to make it harder to merge by accident.

cjyetman:
New arguments should also preferably have a default value so that existing calls to the function do not fail because they don't explicitly set a value for the new argument, e.g. `peergroup = NULL`

Clare2D:
replace with new one

PR #204, opened 2020-10-14
Trigger gh-actions

maurolepore:


maurolepore:
I'm surprised to see that the integration test passes on GH-actions. I expected them to brake because right now the `master` branch of `create_interactive_report` is broken. Locally the test errors.

PR #205, opened 2020-10-14
explicitly name function arguments

cjyetman:
explicitly name function arguments to avoid problems if/when the function signature changes

however, if someone adds an argument to the function without a default value, this will not work around that problem I suppose


maurolepore (approved):
LGTM

PR #206, opened 2020-10-14
New test_r_scripts() replaces test_integration()

maurolepore:


.github/workflows/test-r-scripts.yaml
```diff
@@ -51,20 +51,16 @@ jobs:
           r-version: ${{ matrix.config.r }}
           http-user-agent: ${{ matrix.config.http-user-agent }}
 
-      - uses: r-lib/actions/setup-pandoc@v1
-      - name: Install rmarkdown
-        run: |
-          install.packages("remotes")
-          remotes::install_cran("rmarkdown")
-        shell: Rscript {0}
-
```
maurolepore, line 60: We no longer need pandoc nor rmarkdown

.github/workflows/test-r-scripts.yaml
```diff
@@ -51,20 +51,16 @@ jobs:
...
-      - name: Run integration test
+
+      - name: Test R scripts
         run: |
-          cd PACTA_analysis
-          Rscript -e "knitr::knit('integration-test.Rmd')"
+          setwd("PACTA_analysis")
+          source("R/test_r_scripts.R")
+          test_r_scripts()
+        shell: Rscript {0}
```
maurolepore, line 66: We now reuse the funciton `test_r_scripts()`

R/test_r_scripts.R
```diff
@@ -0,0 +1,14 @@
...
+    message("Testing: ", command[[i]])
+    system(command[[i]])
+  }
+
+  invisible(x)
+}
+
+r_script_command <- function(x) {
+  sprintf("Rscript --vanilla web_tool_script_%s.R TestPortfolio_Input", x)
+}
```
maurolepore, line 14: These are all valid calls:

```r
test_r_scripts()
# Same
test_r_scripts(1:3)

# Just one
test_r_scripts(1)
```

R/utils.R
```diff
@@ -1,10 +1,3 @@
-test_integration <- function(input = "integration-test.Rmd") {
-  needless <- tempfile(fileext = ".html")
-  rmarkdown::render(input, output_file = needless)
-
-  invisible(input)
-}
-
```
maurolepore, line 7: We no longer need this.

maurolepore (commented):
This PR improves the code base by testing the web-tool in a way closer to the one it runs in production. It should also remove the system dependency on pandoc, and the R dependency on rmarkdown.

cjyetman:
This all **looks** good to me, but honestly I'm not up to speed on how to test from a branch form a PR, so I haven't actually tested this myself. I guess I could/should try to switch to this branch and test it, but I'm afraid of mucking something else up.

cjyetman (approved):
awesome sauce! works for me

maurolepore:
For the record, there are many options to checkout a pr. 

With the `gh` CLI

```bash
gh pr checkout 206
```

With usethis

```r
pr_fetch(206)
```

Or the uglier options linked at the end of every PR

image

cjyetman:
for the record, I just switched to the `simpler-integration-test` branch and tested it, which was super easy for me and seemed to work, and then switched back 😄

PR #209, opened 2020-10-15
Pass Stress Test data to report

jacobvjk:
Depends on  2DegreesInvesting/create_interactive_report#44

This PR:

- passes stress test result data for Bonds and/or Equity to the interactive report
- passes the portfolio_overview to the interactive report.
- passes an empty tibble with the same columns to the interactive report in case either of the tables is missing

This is required in order to prepare the stress test data for visualisation

PR #210, opened 2020-10-15
Update project params

jacobvjk:
This PR:

- introduces two new parameters, `peergroup` and `user_id`
- receives these via the project config when running the web tool
- passes peergroup to create_interactive_report() as named argument, to provide information about comparison groups

**OPEN:**

- do we pass user_id to the report as is? @Clare2D 
- should I pass the peergroup name AND the corresponding file to the report or just one of them? @Clare2D


web_tool_script_3.R
```diff
@@ -182,6 +182,7 @@ create_interactive_report(
   template_dir = template_dir,
   company_charts_dir = company_charts_dir,
   output_dir = output_dir,
+  peergroup = peergroup,
```
cjyetman, line 185: please be careful here.... [see here](https://github.com/2DegreesInvesting/PACTA_analysis/pull/209#issuecomment-708988926) for some pointers on how best to make changes like this

cjyetman: i.e. as far as I can tell, this PR **will** break things since the current `create_interactive_report()` function signature will not know how to handle this new argument

jacobvjk: Yes, I guess this would pretty much just re-introduce the bug from yesterday. Will not open the PR, until we at least do something with these arguments in create_interactive_report()

Clare2D:
I'm working with these variables as well as a language input. Working out
how to merge over multiple repositories at the same time isn't simple. I
will just work on this during the day today and discuss merging this
evening.

cjyetman:
I put some ways to deal with this over here [PACTA_analysis/pull/209#issuecomment-708988926](https://github.com/2DegreesInvesting/PACTA_analysis/pull/209#issuecomment-708988926)
> OK, clearly this is going to be on an on-going issue, which is the result of keeping these repositories separate.
> 
> These arguments don't exist in the current `create_interactive_report()` function signature, so this PR will break things.
> 
> 1. PRs like this need to be coordinated somehow with PRs in the other repo so that function signature changes like this can occur (more or less) simultaneously in both repos... and then we should probably start a habit of notifying everyone (like in the #pacta2020 Slack channel?) that a breaking change has just occurred and everyone must update to the current origin master.
> 2. There are a few best practices we can use to reduce the problems that these types of changes cause...
>    a. **always** put new function arguments at the **end** of the argument list
>    b. **always** add a default value for new function arguments, e.g. `new_argument = NULL`
>    c. **always** explicitly name arguments when calling the function from somewhere else, e.g. `my_func(param1 = my_value)`
>    d. **maybe** add the `...` argument to the end of functions in heavy development so that they can capture any new/other/undefined arguments being passed to it (as is the case in this PR) and still function properly just ignoring the unknown arguments
> 
> @maurolepore do you agree and/or have any other suggestions?
> 
> https://github.com/2DegreesInvesting/create_interactive_report/blob/343c6bcee844a165c6ad220107ff9f2eba603d0c/create_interactive_report.R#L1-L29
> 
> ```r
> create_interactive_report <-
>   function(repo_path,
>            template_dir,
>            company_charts_dir,
>            output_dir,
>            project_name,
>            investor_name,
>            portfolio_name,
>            start_year,
>            select_scenario,
>            portfolio_allocation_method,
>            scenario_geography,
>            twodi_sectors = c('Power', 'Automotive', 'Shipping', 'Oil&Gas', 'Coal', 'Steel', 'Cement', 'Aviation'),
>            gbtech_sectors = c('Power', 'Automotive', 'Oil&Gas', 'Coal'),
>            green_techs = c('RenewablesCap', 'HydroCap', 'NuclearCap', 'Hybrid', 'Electric'),
>            tech_roadmap_sectors = c('Automotive', 'Power', 'Oil&Gas', 'Coal'),
>            audit_file,
>            emissions,
>            equity_results_portfolio,
>            bonds_results_portfolio,
>            equity_results_company,
>            bonds_results_company,
>            equity_results_map,
>            bonds_results_map,
>            indicies_equity_results_portfolio,
>            indicies_bonds_results_portfolio,
>            peers_equity_results_portfolio,
>            peers_bonds_results_portfolio
>   ) {
> ```

jacobvjk:
@Clare2D is going to add these same changes as part of her PR, so I will close this PR as soon as hers is merged

cjyetman:
please link to the PR in 2DegreesInvesting/create_interactive_report (when it's available) that enables these arguments... and please leave this in draft mode until that PR lands

otherwise, this seems like a no-brainer

jacobvjk:
sure, I think it's this one, but maybe there will be a new one. either way, will update once there is more clarity

Clare2D:
These changes are being merged atm

cjyetman:
this was just closed... the same changes are being merged in a different PR?

btw, 2DegreesInvesting/create_interactive_report/pull/49 added the `peer_group` argument to the `create_interactive_report` function, but this PR uses `peergroup`

Clare2D:
Yes the same changes are being made. peer_group > peergroup. @jacobvjk  is this what you're using? or can you?!

jacobvjk:
sure, peer_group is fine, although I am not personally working with it right now anyway.

PR #211, opened 2020-10-15
update list of needed pkgs

cjyetman:
Now that we have a docker image that has all of the necessary pkgs pre-installed **and** that I previously modified the `install.packages` command in this file, I think it's best to also have here the same complete list of necessary pkgs to install. Which leads to two possible scenarios....

1. case: you use the default `2dii/r-packages:latest` docker image, or another docker image that already has all of the pkgs needed installed; result: this command will skip installing anything because all of these will already be installed, so this will remain a quick start-up

2. case: you use some alternate docker image that does *not* already have all of the needed pkgs installed; result this command will force the docker instance to install any (but only) pkgs that are need and are not already installed so that the other scripts can finish successfully


jacobvjk (approved):
lgtm
the options you outlined in your comment seem to be very reasonable, good to go from my side

cjyetman:
I'm gonna force this one through because it's more likely to prevent problems than to create new ones.

maurolepore (commented):
Thanks

PR #215, opened 2020-10-15
Use testtaht ed3

maurolepore:
This PR activates testthat edition 3, which adds cool new features.

https://testthat.r-lib.org/articles/third-edition.html#activating


cjyetman (approved):
well, if you like living on the edge, this seems to be the way to go 😉

maurolepore:
Yeah, fair enough. Will come back to this after testthat 3 is on CRAN

PR #216, opened 2020-10-15
TODO: SPLIT TOO LARGE. New functions to update R packages and dockerfile

maurolepore:
This is a massive PR that I plan to split and resubmit.


.Rbuildignore
```diff
@@ -46,3 +46,4 @@
 ^deduplicate$
 ^bin$
 ^docker$
+^\.renvignore$
```
maurolepore, line 49: Put here the paths you want renv to ignore when it detects packages.

inst/extdata/r_packages.R
```diff
@@ -0,0 +1,40 @@
+library(assertthat)
+library(bookdown)
+library(config)
+library(conflicted)
```
maurolepore, line 4: This file is in a weird place just to make it accessible via `system.file()` in tests -- but I'm now I need that so I we may eventually move it to a more natural place.

tests/testthat/test-find_dependencies.R
```diff
@@ -0,0 +1,41 @@
+test_that("outputs the expected tibble", {
```
maurolepore, line 1: @jacobvjk, play with this test code to try figure out how it works. 
You can run it interactively and also via `devtools::test()` (or control + shift + T)

tests/testthat/test-update_dockerfile_packages.R
```diff
@@ -0,0 +1,10 @@
+test_that("updates a dockerfile with the given R packages file", {
```
maurolepore, line 1: @jacobvjk, play with this test code to try figure out how it works. 
You can run it interactively and also via `devtools::test()` (or control + shift + T)

tests/testthat/test-update_r_packages.R
```diff
@@ -0,0 +1,11 @@
+test_that("updates a file as epected", {
```
maurolepore, line 1: @jacobvjk, play with this test code to try figure out how it works. 
You can run it interactively and also via `devtools::test()` (or control + shift + T)

maurolepore (commented):
Here are the highlights of this PR. In short, it adds some features to detect packages automatically via renv, then to use those packages to write a file with calls to `library()`, and a dockerfile.

R/update_dockerfile_packages.R
```diff
@@ -0,0 +1,82 @@
...
+#' writeLines(readLines(some_dockerfile))
+#' update_dockerfile_packages(path = some_dockerfile)
+#' # After
+#' writeLines(readLines(some_dockerfile))
+#'
+#' # By defaul it writes to a temporary file
+#' tmp <- update_dockerfile_packages()
+#' writeLines(readLines(tmp))
+#' @noRd
+update_dockerfile_packages <- function(dockerfile = NULL,
```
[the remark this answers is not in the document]
maurolepore: UPS! FIXME should be install.packages — not library !!

maurolepore:
Flollow up at #220 #221 #222 #223

PR #218, opened 2020-10-16
adapt to new variable names in recently updated pacta-data repo

cjyetman:


Clare2D (approved):

cjyetman:
FYI... this is to adapt to upstream changes that have already merged here:
https://github.com/2DegreesInvesting/pacta-data/commit/f7b658c3639953fefc48ad3a9c6892a17a33a2b1
https://github.com/2DegreesInvesting/pacta-data/commit/ef182e359b540ffc285e70154a357ce81acb9b82

0_portfolio_input_check_functions.R
```diff
@@ -995,14 +995,14 @@ get_and_clean_company_fin_data <- function() {
 
   # comp_fin_data_raw <- comp_fin_data_raw %>% select(
   #   company_id, company_name, bloomberg_id, country_of_domicile, corporate_bond_ticker, bics_sector, bics_subgroup,
-  #   icb_subgroup, mapped_sector, has_asset_level_data, has_assets_in_matched_sector, sectors_with_assets,
+  #   icb_subgroup, financial_sector, has_asset_level_data, has_assets_in_matched_sector, sectors_with_assets,
   #   current_shares_outstanding_all_classes, company_status, market_cap, bond_debt_out,
   #   financial_timestamp
   # )
```
maurolepore, line 1001: Nit: Are we following up with a PR to remove this dead code?

0_portfolio_input_check_functions.R
```diff
@@ -1531,7 +1531,7 @@ calculate_portfolio_emissions <- function(
 add_other_to_sector_classifications <- function(audit) {
   # fix sector classifications
   audit <- audit %>%
-    mutate(ald_sector = ifelse(mapped_sector != "Other" & is.na(ald_sector), mapped_sector, ald_sector))
+    mutate(ald_sector = ifelse(financial_sector != "Other" & is.na(ald_sector), financial_sector, ald_sector))
```
maurolepore, line 1534: I don't remember why but I think `dplyr::if_else()` has some advantage over `ifelse()`.

web_tool_script_3.R
```diff
@@ -216,10 +216,10 @@ if (file.exists(file.path(proc_input_path, portfolio_name, "overview_portfolio.r
...
 
 
-indicies_equity_results_portfolio <- read_rds(file.path(data_location_ext, "0_Indices_equity_portfolio.rda"))
-indicies_bonds_results_portfolio <- read_rds(file.path(data_location_ext, "0_Indices_bonds_portfolio.rda"))
-peers_equity_results_portfolio <- read_rds(file.path(data_location_ext, "0_Indices_equity_portfolio.rda"))
-peers_bonds_results_portfolio <- read_rds(file.path(data_location_ext, "0_Indices_bonds_portfolio.rda"))
+indicies_equity_results_portfolio <- read_rds(file.path(data_location_ext, "Indices_equity_portfolio.rda"))
+indicies_bonds_results_portfolio <- read_rds(file.path(data_location_ext, "Indices_bonds_portfolio.rda"))
+peers_equity_results_portfolio <- read_rds(file.path(data_location_ext, "Indices_equity_portfolio.rda"))
+peers_bonds_results_portfolio <- read_rds(file.path(data_location_ext, "Indices_bonds_portfolio.rda"))
```
maurolepore, line 222: Can we follow up with a PR to save these files with the extension .rds?

maurolepore (approved):
LGTM. I added a few comments and ideas to follow up.
Merge whenver you are ready.

0_portfolio_input_check_functions.R
Clare2D: if_else expects that the types are going to be the same ie it's replacing a character with a character so it throws more errors, which probably isn't a bad thing, but I sometimes ignore this. I think there are issues also with replacing NAs with characters, but I'm not sure.

web_tool_script_3.R
Clare2D: I've been thinking about a "data prep script", that does some of the tasks done when "new_data == TRUE". this could also include changing these file formats, and doing some additional renaming.

0_portfolio_input_check_functions.R
cjyetman: while philosophically I agree with cleaning up and removing "dead" code, I'm not currently in a position to determine whether or not this code is truly "dead" or if it's commented out for a good reason

web_tool_script_3.R
cjyetman: @maurolepore maybe create issues for these follow-up PR requests rather than comments that will probably be forgotten once this is merged?

0_portfolio_input_check_functions.R
cjyetman: yes, @Clare2D is exactly right... `dplyr::if_else()` is "safer" because it type checks everything... which can lead to more errors (which some may say is a good thing), and which may require you to return type specific `NA`s... like `NA_character_` instead of just `NA`

0_portfolio_input_check_functions.R
maurolepore: I know, that's precisely why it's a bad idea to comment code out in the first place -- nobody wants to take responsability for deleting it. I say we delete it. If we need it, we can find it with something like

```bash
git log -S "comp_fin_data_raw"
```

cjyetman: this very PR, for instance, has what I consider to be "legitimate" use of commented code...

https://github.com/2DegreesInvesting/PACTA_analysis/blob/ac4802218c57fab978e04d521376afe21ea2a05e/web_tool_script_3.R#L251-L259

those arguments for the stress testing stuff are removed from the current master of `create_iteractive_report` so this PR will fail/break if I leave them in, but they will very likely need to be added back in once the stress test stuff settles, and i.m.o. it's easier and safer to uncomment rather than manually re-adding them.

maurolepore: Good, for every rule there is always a good reason to not follow it :)

PR #220, opened 2020-10-16
Setup R packages for later PRs

maurolepore:
Follows #216 
Merging sequence #220 > #222 > #221 > #223

This PR doesn't do much other than setting the R project for other PRs to come.

PR #221, opened 2020-10-16
New update_r_packages()

maurolepore:
Depends on #222

Follows #216
Merging sequence #220 > #222 > #221 > #223


maurolepore:
CJ suggested to now merge this until after the upcoming release. I'll close now to avoid confusion and direct attention to where we need it most. I may reopen eventually.

PR #226, opened 2020-10-17
Inc project code

Clare2D:
This is necessary to choose the correct output template and possibly as a test for other decisions later


Clare2D:
This should be a very small PR and should come after the original one is merged

cjyetman (approved):
This is part of a big merge across multiple repos. @Clare2D and @jacobvjk and I all understand that there is follow-up needed in the StressTest and create_interactive_report repos before everything is working again.

PR #230, opened 2020-10-19
use a universally acceptable font family spec

### LINKED ISSUE #229: font specification must be universally workable, or at least on Linux
https://github.com/2DegreesInvesting/PACTA_analysis/blob/4715968160a7175ca5812e192ee7d3519207ac13/0_portfolio_input_check_functions.R#L1165

This font family seems to have changed a few times. If this is meant to work on Linux/docker so that the whole system can run on the web server, we need to choose a font family that will work on Linux. Ideally, we choose a font family that works universally across systems (according to [this](https://ggplot2.tidyverse.org/articles/ggplot2-specs.html#font-face), one of "sans", "serif", or "mono")

this blocks #225 because that reintroduces `create_audit_chart(audit_file, proc_input_path)` which depends on this working properly

cjyetman:
closes #229

### DIFF
--- a/0_portfolio_input_check_functions.R
+++ b/0_portfolio_input_check_functions.R
@@ -1162,7 +1162,7 @@ create_audit_chart <- function(audit_file, proc_input_path) {
 
   ## GRAPH VALUES
   base_size <- 12
-  base_family <- "Clear Sans"
+  base_family <- "sans"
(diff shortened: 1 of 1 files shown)


Clare2D (approved):
Fine for me

PR #240, opened 2020-10-20
Port weight calc

Clare2D:
Still a draft while @2diiKlaus confirms what the scope of the calculation is.  
@cjyetman the main concern I've got is what happens if there is no data. Can you check this code for smart alternatives here?

closes #79

### DIFF
--- a/0_portfolio_input_check_functions.R
+++ b/0_portfolio_input_check_functions.R
@@ -1556,7 +1556,7 @@ pw_calculations <- function(eq_portfolio, cb_portfolio){
 
   if(data_check(port_all)){
 
-  port_all <- port_all %>%  select(grouping_variables,company_id, value_usd)
+  port_all <- port_all %>%  select(!!!rlang::syms(grouping_variables),company_id, value_usd)
 
   port_all <- calculate_port_weight(port_all, grouping_variables)
 
--- a/0_portfolio_input_check_functions.R
+++ b/0_portfolio_input_check_functions.R
@@ -1134,6 +1134,7 @@ create_merged_portfolio <- function(eq_portfolio, cb_portfolio) {
 }
 
 create_portfolio_subset <- function(portfolio, portfolio_type, relevant_fin_data) {
+
   if (portfolio_type %in% unique(portfolio$asset_type)) {
     portfolio_subset <- portfolio %>%
       ungroup() %>%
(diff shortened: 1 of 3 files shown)


web_tool_script_1.R
```diff
@@ -7,6 +7,7 @@ source("0_portfolio_input_check_functions.R")
 source("0_global_functions.R")
 source("0_web_functions.R")
 source("0_json_functions.R")
+source("0_portfolio_test.R")
```
cjyetman, line 10: should this be `source("0_portfolio_input_check_functions.R")` since that's where you added the `pw_calculations()` function?

0_portfolio_input_check_functions.R
```diff
@@ -1538,16 +1539,37 @@ add_other_to_sector_classifications <- function(audit) {
...
+
+  if(data_check(cb_portfolio)){
+
+    port_all <- bind_rows(port_all, cb_portfolio)
+
+  }
+
+  if(data_check(port_all)){
+
+  port_all <- port_all %>%  select(grouping_variables,company_id, value_usd)
```
cjyetman, line 1559: is `grouping_variables` the name of a column that is expected to be in `eq_portfolio` or `cb_portfolio`, or is it meant to be a character vector containing the names of columns to group by? If the latter, it should probably be unquoted, no?

0_portfolio_input_check_functions.R
```diff
@@ -1538,16 +1539,37 @@ add_other_to_sector_classifications <- function(audit) {
...
+
+  port_all <- calculate_port_weight(port_all, grouping_variables)
```
cjyetman, line 1561: same here

cjyetman: actually, no.... it's getting passed as a function arg here

0_portfolio_input_check_functions.R
```diff
@@ -1538,16 +1539,37 @@ add_other_to_sector_classifications <- function(audit) {
...
+
+
+  pw <- port_all %>%
+    group_by(!!!rlang::syms(grouping_variables), company_id) %>%
```
cjyetman, line 1565: so yeah, this seems more appropriate way to use `grouping_variables`.... though I'm not sure if the syntax here is the most appropriate/efficient

cjyetman (commented):
needs a fix to `grouping_variables`, then it should be ready to go

cjyetman:
is this related to #79?

0_portfolio_input_check_functions.R
Clare2D: It is grouping_variables <- c("investor_name", "portfolio_name") is the latter right. What do you mean by unquoted? It is unquoted right?

0_portfolio_input_check_functions.R
Clare2D: Yeah this is what was working a while ago re syntax so copied that. can change at some point if there is something better.

Clare2D:
Yes this is about #79

0_portfolio_input_check_functions.R
cjyetman: it should probably have the same syntax as [L1565](https://github.com/2DegreesInvesting/PACTA_analysis/pull/240/files#diff-9a3348aa2255d360e2e12fe80be7719715a083dc6e42d5af91fe5d53b4315b3dR1565), no?
`!!!rlang::syms(grouping_variables)`

web_tool_script_1.R
Clare2D: no it also sources now a file from port_test ; portfolio_input_check_functions is already sourced!

Clare2D:
Made that change now. I think it's good to go.

cjyetman (approved):
approved... I leave it to @Clare2D or @2diiKlaus to merge

2diiKlaus (commented):
Can we change `port_weight ` to `portfolio_weight`

the webdeveloper needs a different output than rda. json would be ideal

PR #241, opened 2020-10-21
parameterize the portfolio name when using the web tool

### LINKED ISSUE #176: TestPortfolio_Input should be parameterized
https://github.com/2DegreesInvesting/PACTA_analysis/pull/164/files#r503134832

We should be able to do something like this:

```bash
portfolio=${1:-"TestPortfolio_Input"}
Rscript --vanilla web_tool_script_1.R "$portfolio"
Rscript --vanilla web_tool_script_2.R "$portfolio"
Rscript --vanilla web_tool_script_3.R "$portfolio"
```

But notice that "TestPortfolio_Input" is still hard-coded elsewhere.

cjyetman:
This PR allows the `bin/run-web-tool` shell script to take a/the portfolio name as an argument and passes it through the scripts/process so that it is read in by the web tool scripts.

e.g. `./bin/run-web-tool TestPortfolio_Input` and/or `./bin/run-web-tool MyCustomPortfolio` will/should work

closes #176


cjyetman:
testing this on current master, everything seems to work as expected until the very end, just the latex rendering of the executive summary fails for some very odd reason...

```
output file: /create_interactive_report/swiss_en_exec_summary/template.knit.md

/usr/bin/pandoc +RTS -K512m -RTS /create_interactive_report/swiss_en_exec_summary/template.utf8.md --to latex --from markdown+autolink_bare_uris+tex_math_single_backslash --output /bound/working_dir/50_Outputs/TestPortfolio_Input/template.tex --self-contained --highlight-style tango --latex-engine xelatex --include-in-header preamble.tex --variable graphics 
pandoc: preamble.tex: openFile: does not exist (No such file or directory)
Error: pandoc document conversion failed with error 1
Execution halted
```

but other than that, I'm pretty sure this is the simplest way to achieve what was intended

cjyetman:
ignore last comment... it appears I'm now getting the same error on master, so something else must have changed on my system... but it doesn't appear to be caused by this PR

cjyetman:
@maurolepore I believe this PR is ready to go, but I will leave it in draft mode until you're able to review it and determine whether or not it fully, properly achieves your goal here #176

cjyetman:
after discussion with @maurolepore today, I think we both agreed that this PR is ready to merge, but...

because it requires users to change their current testing process from using `./bin/run-web-tool` to `./bin/run-web-tool TestPortfolio_Input`, we're going to wait till things settle down to impose this change (don't want to cause any additional
problems while we're ironing things out)

however, I believe this will be necessary to enable the web tools to work properly on the web server when it goes live, so we will need to merge this within the next week

bin/README.md
```diff
@@ -1,22 +1,25 @@
 ## Running the web tool on transitionmonitor's docker-container
 
-Import the base images with:
-
-```bash
-docker image import https://tool.transitionmonitor.com/downloads/2diirunner_1.3.0.tar.xz 2diirunner:1.3.0
-```
-
```
maurolepore, line 8: This is no longer relevant. We now use images from dockerhub.

bin/run-web-tool
```diff
@@ -1,3 +1,3 @@
 #! /bin/bash
 
-"$(pwd)"/bin/start-containter-with /bound/bin/run-r-scripts
+"$(pwd)"/bin/start-containter-with /bound/bin/run-r-scripts "${1:-TestPortfolio_Input}"
```
maurolepore, line 3: This is the high level script what I expect users to call. It now has a sensible default to simplify testing.

bin/README.md
```diff
@@ -1,22 +1,25 @@
...
 Ensure your working directory is PACTA_analysis, then run one of these workflows:
 
 * Run the web tool in a single step
 
 ```bash
+# Use the default portfolio name: TestPortfolio_Input
 ./bin/run-web-tool
```
maurolepore, line 9: This is the easiest way to run all scripts in the web tool. Maybe the name `run-web-tool` is missleading.  The most common use I expect is testing, so maybe we should rename to `test-web-tool` or `test-r-scripts`?

maurolepore (commented):
> because it requires users to change their current testing process from using ./bin/run-web-tool to ./bin/run-web-tool TestPortfolio_Input, we're going to wait till things settle down to impose this change.

@CJ, I added a default so users can continue to use the short form `./bin/run-web-tool` for quick tests, while also allowing custom portolio names with the form `./bin/run-web-tool some-portfolio-name`.

See what you think.

bin/README.md
cjyetman: 👍

bin/README.md
cjyetman: I think the reference to "web tool" is important here... there are many R script in this repo, some of which are never touched when running all there of the web tool scripts

bin/run-web-tool
cjyetman: haven't test this myself yet, but if this works, great! 😄

bin/run-r-scripts
```diff
@@ -10,7 +10,6 @@ echo 'options(repos = c(CRAN = "https://packagemanager.rstudio.com/all/__linux__
 echo 'options(HTTPUserAgent = sprintf("R/%s R (%s)", getRversion(), paste(getRversion(), R.version$platform, R.version$arch, R.version$os)))' >> ~/.Rprofile
 # Rscript -e "update.packages(ask = FALSE)"
 Rscript -e 'pkgs_to_install <- c("assertthat", "bookdown", "config", "conflicted", "countrycode", "data.table", "devtools", "dplyr", "forcats", "fs", "fst", "ggplot2", "glue", "here", "highcharter", "janitor", "jsonlite", "knitr", "purrr", "readr", "readxl", "renv", "reshape2", "rlang", "rmarkdown", "rstudioapi", "scales", "stringr", "testthat", "tibble", "tidyr", "tidyselect", "usethis", "withr", "writexl", "zoo"); install.packages(setdiff(pkgs_to_install, installed.packages()))'
```
maurolepore, line 12: @cjyetman do we need to install packages? The packages we need might already be installed in the image 2dii/r-packages -- or we should extend it so they are.

cjyetman: That line shouldn't be unnecessary. I left it as a fail safe in case the docker image hadn't been updated yet.

maurolepore: Good idea. Should I ensure all packages are includes in 2dii/r-packages and then remove this line?

cjyetman: I like the sound of that

cjyetman:
given @maurolepore's recent improvements, particularly the ability to still run the command without an explicit portfolio name argument so that it doesn't break existing usage, I've taken this out of draft mode and give my approval for it

maurolepore (commented):
Thanks!

PR #255, opened 2020-10-25
Updates revent changes

Clare2D:
This PR needs to go along with the PR in create_interactive_report and should be merged before too much more work is done. 

That being said - the scenario issue needs to be resolved. Ie that raised in issue https://github.com/2DegreesInvesting/pacta-data/issues/5 

This PR includes the scenario source in the output that is passed to the results, so it should be considered whether this happens already.

depends on 2DegreesInvesting/create_interactive_report/pull/75


cjyetman (approved):
🤞

Issue #69, opened 2020-08-13
webtool: issues with file paths


jacobvjk:
https://github.com/2DegreesInvesting/PACTA_analysis/blob/6894e33dc71920aeec535b036e1126a44f10950c/web_tool_script_2.R#L56

the definition of equity_inputs_file is different than that of bonds_inputs_file.. I am fairly sure this should not be like that..

in general, we should discuss again, how multiple investors/portfolios within an uploaded file are handled in terms of paths. 
My current understanding is that in the directory working_dir/30_Processed_Inputs/, there should be one folder per each portfolio. This would mean though, that the code needs to write into these folders and, in web_tool_script_2.R, it needs to. read from these folders one at a time.
To my understanding, the code would then loop over all portfolios, when running the analysis. Within the loop, it runs the calculation for one pf at a time and saves the outcome into the correspondingly named folder. Then onto the next pf in the next iteration of the loop. The reason for me to write this, although it ofc sounds trivial, is that I don't see this happening in web_tool_script_2.R. Probably because the way portfolios get passed to the tool and/or split up is different now?
@Clare2D

Clare2D:
Indeed this needs to be changed. This strange reference in the EQ name was me trying to do this before. The real name here will be an input passed from the platform ie. the unique code per portfolio. 

Yep indeed. The tool currently loops over investor name rather than portfolio name. This should be changed for the webtool. It will just increase the time of running, because as you suggest, the script will need to open the portfolio file, then save it in the correct location before moving to the next portfolio. 

I imagine for one investor a sample file path being:

web_dir/30_processed_inputs/XJISOAMBWE/Total_portfolio.csv
web_dir/40_results/XJISOAMBWE/equity_results_portfolio.csv

Without the "project name" and with this "Portfolio folder" saved there. Is this similar to what you thought?

jacobvjk:
Yes, I think that would be a reasonable way to set paths here. I will implement that, including looping over portfolio, in case there is more than one being passed then.

cjyetman:
seems like this was fixed in 273d8d35803aec8f72377c87d94e7b211f2db2cf  and merged in #64 

and then changed numerous time since... current state is...
https://github.com/2DegreesInvesting/PACTA_analysis/blob/9978f00a3e871571d8e37b71d9e7bee848a38f24/web_tool_script_2.R#L41

can we close this @jacobvjk or @Clare2D?

PR #256, opened 2020-10-26
sync up arguments

jdhoffa:
Sorry guys, the arguments between PACTA_analysis and create_interactive_report fell out of sync, I think when I reverted 2DegreesInvesting/PACTA_analysis#69

With this PR, and it's counterpart in create_interactive_report, we should be running as normal again.


jdhoffa:
Locally, `./bin/run-web-tool` now works without error for me:

cjyetman:
@jdhoffa can you revisit this please, given the context of massive PRs just merged to master here and in 2DegreesInvesting/create_interactive_report

Clare2D:
I think these are now superceded.

jdhoffa:
yup looks good, thanks!

Issue #262, opened 2020-10-26
web process fails without isins_missing_bbg_data.csv 


cjyetman:
#260 removed these files that are in gitignore and/but get changed every time the the web tools process is run, so I assumed they were not necessary

turns out `data/isins_missing_bbg_data.csv` is needed because this function reads it and writes it every time the process runs... so it must be in git, it must change every time, and we must just ignore that it's changing all the time...
https://github.com/2DegreesInvesting/PACTA_analysis/blob/9671739c0ffeda205100e79ca2bab09879643ee3/0_portfolio_input_check_functions.R#L775-L789

so, I'm going to re-add `data/isins_missing_bbg_data.csv` in a following PR, but I think this issue should stay open... something doesn't feel right about this process... maybe if we need to read in that file and make changes to it, maybe the changes should be written to a new file that can be safely ignored, or something like that, e.g. `data/isins_missing_bbg_data_updated.csv`

PR #260, opened 2020-10-26
stop tracking these files that have been ignored

cjyetman:
these files were ignored a long time ago in this commit https://github.com/2DegreesInvesting/PACTA_analysis/commit/81f6fb3caa4b1ef6dff252f56cd883d2ed70d34e which was part of this PR #54 

data/fund_isins_without_bbg_data.csv
data/isins_missing_bbg_data.csv

PR #263, opened 2020-10-26
put back data/isins_missing_bbg_data.csv

cjyetman:
#260 removed it from the repo, this puts it back, and removes the unnecessarily duplicated .gitignore command/file

this addresses #262 in the near term, but doesn't completely resolve it


maurolepore (approved):
LGTM -- the changes reflect the intent.

Issue #271, opened 2020-10-27
webtool: remove auditchart and websitetext


jacobvjk:
`AuditChart.png` and `websitetext.txt` are both not required by the reports that are generated by the webtool.

They are written to a directory that is not portfolio specific. This means that when running the tool on diff portfolios in production, they would be overwritten every time.

Since there is no use in writng them anywhere else, they should be removed instead.

fyi @Clare2D

PR #272, opened 2020-10-27
Rm unnecessary files

jacobvjk:
This PR:
- corrects file paths for `create_audit_chart()` and `website_text()`
- comments out function calls in webtool script 1 that write files we do not need in the web tool
- this ensure we do not overwrite these outputs for diff portfolios in production

closes: https://github.com/2DegreesInvesting/PACTA_analysis/issues/271

### DIFF
--- a/0_web_functions.R
+++ b/0_web_functions.R
@@ -246,7 +246,7 @@ check_input_file_contents <- function(portfolio_, portfolio_name, investor_name)
   return(portfolio_clean)
 }
 
-website_text <- function(audit_file) {
+website_text <- function(audit_file, proc_input_path) {
   PortValues <- audit_file %>%
     ungroup() %>%
     filter(valid_input == TRUE) %>%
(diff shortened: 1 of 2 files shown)


cjyetman (approved):
as long as the only call of `website_text()` in this repo is the one you've removed here, this lgtm.... otherwise, as always, it's dangerous to change function signatures

cjyetman:
though, this doesn't do what its title suggests "Rm unnecessary files"... is that purposeful or is there more commits coming?

jacobvjk:
> though, this doesn't do what its title suggests "Rm unnecessary files"... is that purposeful or is there more commits coming?

Bad naming of the branch. It comments the functions out, so the files won't be created anymore.
That's the main idea here. I did change the functions a bit, as they are handling the paths wrong either way.

jacobvjk:
> as long as the only call of `website_text()` in this repo is the one you've removed here, this lgtm.... otherwise, as always, it's dangerous to change function signatures

It's the only place where this is used and I figured this signature makes more sense

PR #280, opened 2020-10-29
Scenario data updates

Clare2D:
This includes updates for the new scenario data as per the PR in create_interactive_report.

The aim here is to:
- update the parameter file 
- update the chosen scenario for the report
- cleaning previous approach which added the "scenario_source" column to the outputs (web_tool_script_2) and then checked and cleaned this afterwards (web_tool_script_3)

Possibly this could be changed to include more compatibility with either way of naming the scenarios, but this hasn't been implemented.

depends on 2DegreesInvesting/pacta-data/pull/6
related to 2DegreesInvesting/create_interactive_report/pull/118

### DIFF
--- a/web_tool_script_3.R
+++ b/web_tool_script_3.R
@@ -36,7 +36,7 @@ real_estate_dir <- path(user_results_path, project_code, "real_estate")
 
 output_dir <- file.path(outputs_path, portfolio_name_ref_all)
 
-scenario <- "SDS"
+scenario <- "WEO2019_SDS"
 portfolio_allocation_method <- "portfolio_weight"
 scenario_geography <- "Global"
 audit_file <- read_csv(file.path(proc_input_path, portfolio_name, "audit_file.csv"), col_types = cols())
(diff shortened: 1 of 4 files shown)


Clare2D:
These PRs will cause the tool to fail until @jacobvjk  make the missing updates to the Stresstest tool but they are important for progressing with relevant changes to charts

cjyetman:
does this depend on 2DegreesInvesting/pacta-data/pull/6?

Clare2D:
Should work with the updated create_interactive_report PR https://github.com/2DegreesInvesting/create_interactive_report/pull/118

web_tool_script_3.R
```diff
@@ -36,7 +36,10 @@ real_estate_dir <- path(user_results_path, project_code, "real_estate")
 
 output_dir <- file.path(outputs_path, portfolio_name_ref_all)
 
-scenario <- "SDS"
+scenario <- "WEO2019_SDS"
+scenario_auto <- "ETP2017_B2DS"
```
jacobvjk, line 40: shouldn't this be `"ETP2017_SDS"`?

web_tool_script_3.R
```diff
@@ -36,7 +36,10 @@ real_estate_dir <- path(user_results_path, project_code, "real_estate")
...
+scenario_other <- "ETP2017_B2DS"
```
jacobvjk, line 41: shouldn't this be `"ETP2017_SDS"`?

jacobvjk (commented):
see comments, just validating

web_tool_script_3.R
jacobvjk: @Clare2D

web_tool_script_3.R
jacobvjk: @Clare2D

web_tool_script_3.R
jacobvjk: https://github.com/2DegreesInvesting/PACTA_analysis/issues/289

web_tool_script_3.R
jacobvjk: https://github.com/2DegreesInvesting/PACTA_analysis/issues/289

jacobvjk (approved):
Comments have been outsourced to new tasks, so this
LGTM

PR #282, opened 2020-10-30
Fix how sed strips whitespace

cjyetman:
this version works for me
```
sed 's/  */ /g'
```

this version does not
```
sed s/"  \+"/" "/
```

any ideas why/which is better?

### DIFF
--- a/bin/show-upstream-and-status-at
+++ b/bin/show-upstream-and-status-at
@@ -8,5 +8,5 @@ git branch -vv | \
   # Remove leading "* " in "* <current branch> ..."
   sed s/"^* "/""/ | \
   # Replace duplicated white space with a single white space
-  sed s/"  \+"/" "/
+  sed 's/  */ /g'
 git status


AlexAxthelm:
`sed` doesn't understand `+` or other PCRE symbols by default. You can enable them with the `-E` flag. If you do that, I'd suggest using `\s+` for good measure.

maurolepore (approved):
LGTM. Your solution seems to do the job.
Thanks @AlexAxthelm for pointing out `-E`.

For the record, here are some examples:

```bash
➜  ~ echo "abc          def          ghi" | sed 's/  */ /'
abc def          ghi
➜  ~ echo "abc          def          ghi" | sed 's/  +/ /'
abc          def          ghi
➜  ~ echo "abc          def          ghi" | sed 's/  \+/ /'
abc def          ghi
➜  ~ echo "abc          def          ghi" | sed -E 's/  +/ /' 
abc def          ghi

```

Issue #109, opened 2020-10-02
Webtool: log files should be written into sub directory (pf_name) of 00_Log_Files


jacobvjk:
If we don't do this, it will not be possible to keep error messages for specific portfolios or users so displaying them at the right point will be difficult.

First rough overview of what needs to happen:

Currently the write_log function will write output directly into the 00_Log_Files folder. This would mean that it will be overwritten with every run of the tool.
Instead it should be written into a sub directory named after the portfolio that is being analysed. (similar structure as 30, 40, 50...)
This requires setting up an additional subfolder in the create_portfolio_subfolders() function and moving the function further up in the script, ideally close to set_webtool_paths()

additionally, the write_log() function in global_funtions.R needs to be able to get the path it writes to as an input, as the path now differs from the offline version.

The integration-test.Rmd file needs to be adapted as well to cover this change
probably this should be added to the ensure_empty_directory() section.

PR #283, opened 2020-10-30
Move path of logs

jacobvjk:
This PR:

- adds the creation of a subfolder for "00_Log_Files" to the function `create_portfolio_subfolders()`
- `working_dir/00_Log_Files/<pf_name>/`
- moves the `create_portfolio_subfolders` function call to the initialisation section of `web_tool_script_1.R` (required to have the log folder prepared before the first calls to `write_log()`)
- surfaces an implicit function argument in `create_portfolio_subfolders()` to increase control
- changes the `log_path` in the web tool to point to the newly created subfolder (log_path is used for nothing else, so this is safe)
- changes the `write_log()` function to write the logs to a file in `log_path`
- updates `.gitignore` to cover the "00_Log_Files" directory
- adds a `.gitkeep.txt` file to "00_Log_Files"

These changes are required to be able to provide error messages to the user on a portfolio specific level.

closes: https://github.com/2DegreesInvesting/PACTA_analysis/issues/109

### DIFF
--- a/0_global_functions.R
+++ b/0_global_functions.R
@@ -280,6 +280,6 @@ write_log <- function(msg, ...) {
     as.character(msg),
     ...
   )
-  write(composed, file = paste0(project_location, "/00_Log_Files/error_messages.txt"), append = TRUE)
+  write(composed, file = file.path(log_path,"error_messages.txt"), append = TRUE)
 }
 
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,8 @@ vignettes/*.pdf
 *.knit.md
 
 .Ruserdata
+**/working_dir/00_Log_Files/*
+!**/working_dir/00_Log_Files/.gitkeep.txt
 **/working_dir/30_Processed_Inputs/*
 !**/working_dir/30_Processed_Inputs/.gitkeep.txt
 **/working_dir/40_Results/*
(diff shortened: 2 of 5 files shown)


0_global_functions.R
```diff
@@ -280,6 +280,6 @@ write_log <- function(msg, ...) {
     as.character(msg),
     ...
   )
-  write(composed, file = paste0(project_location, "/00_Log_Files/error_messages.txt"), append = TRUE)
+  write(composed, file = file.path(log_path,"error_messages.txt"), append = TRUE)
```
cjyetman, line 283: I really don't like this stuff where we rely on some variable being available in some parent environment. Seems easy enough to pass `log_path` into the function if that's required. Otherwise, how do I know here that it's reasonable to expect that `log_path` will be available?

But, I realize this is what has been done all along, and the change you've made is low-risk because it follows the same strategy and is therefore rather easy to assess.

cjyetman (approved):

0_global_functions.R
jacobvjk: You're right though and it's an easy fix. Will open a follow up issue and get on it right away
https://github.com/2DegreesInvesting/PACTA_analysis/issues/286

Issue #269, opened 2020-10-27
create a GitHub Action that tests web tool scripts with custom docker image


cjyetman:
Apparently you can directly reference a docker container in a GitHub Action, which might be ideal for testing our web tools in our specially made docker image.... for example...
```
jobs:
  my_first_job:
    steps:
      - name: My first step
        uses: docker://alpine:3.8
```

That would allow us to run a test of our web tools scripts exactly how we test locally using the `2dii/r-packages` docker image and `./bin/run-web-tool`, which should be as close to possible to what will run on the web server.

_Originally posted by @cjyetman in https://github.com/2DegreesInvesting/PACTA_analysis/issues/264#issuecomment-716911643_

cjyetman:
I guess this is a duplicate of #268?

maurolepore:
I don't think so. The docker image does not deal with cloning repos -- which is the focus of #268

cjyetman:
Yes, you're correct. I was working late and thinks we're getting a bit fuzzy.

cjyetman:
trying to achieve this in #275 

happy to get any help, because I haven't figured it all out yet

cjyetman:
think it would be ideal if our docker repo would add n image to the GitHub Container Repository and the web tool test scripts would run on that.... see https://github.com/2DegreesInvesting/docker/issues/38

cjyetman:
closed by #444

Issue #267, opened 2020-10-27
install pandoc in .github/workflows/test-r-scripts.yaml


maurolepore:
Follow up https://github.com/2DegreesInvesting/PACTA_analysis/pull/261#issuecomment-716809041

cjyetman:
this could possibly be resolved by using the suggestion in #269 to import the [2DegreesInvesting/2diidockerrunner/r-packages/Dockerfile](https://github.com/2DegreesInvesting/2diidockerrunner/blob/master/r-packages/Dockerfile) image to run the action/workflow

maurolepore:
Closed by #299

PR #285, opened 2020-10-30
get rid of test that can't possibly pass

cjyetman:
remove the R scripts test since it can't possibly pass without pandoc and latex support

maybe this will be addressed eventually in #269 or #267, but for now I think it's best not to run a bunch of tests that we know will fail


maurolepore (commented):
I understand this test is no longer informative but instead of removing it (1) I would rather leave it and fix the code (2). 

1. Test passes > test brakes > remove tests
2. Test passes > test brakes > fix code > test passes

cjyetman:
There's nothing to fix in the code... this fails because the GitHub Action worker image does not have pandoc and latex dependencies installed which are required. If you know how to manage that, that would be preferable, but I've tried and have given up.

cjyetman:
i.e.
code works > add test that can't work > tests fail
code works

Issue #286, opened 2020-10-30
log_path as explicit funciton arg in write_log


jacobvjk:
currently write_log() gets log_path implicitly from the global environment.
surface the argument to the function signature to be more explicit about this

PR #287, opened 2020-10-30
Surface path write log

jacobvjk:
Following up on:

https://github.com/2DegreesInvesting/PACTA_analysis/pull/283

This PR:

- surfaces the file path in `write_log` as the argument `file_path` with default `log_path`, so it does not depend on `log_path` being given via the global environment
- changes all function calls of `write_log` to include the named argument `file_path`

closes: https://github.com/2DegreesInvesting/PACTA_analysis/issues/286

### DIFF
--- a/0_global_functions.R
+++ b/0_global_functions.R
@@ -274,12 +274,12 @@ first_char_up <- function(x) {
 
 
 # write error log for input portfolio - msg should be a string containing the error message
-write_log <- function(msg, ...) {
+write_log <- function(msg, file_path = log_path, ...) {
   composed <- paste(
     as.character(Sys.time()),
     as.character(msg),
     ...
   )
-  write(composed, file = file.path(log_path,"error_messages.txt"), append = TRUE)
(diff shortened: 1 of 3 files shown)


cjyetman (approved):
adding a variable as the default value for a parameter in a function signature is not a great idea (e.g. `file_path = log_path` versus `file_path = "/default/path/to/log"`)

but I see how that makes the most sense here to maintain backward compatibility, and it facilitates moving in the right direction as you have in the rest of this PR adding the file_path/log_path explicitly in the calls to the function

jacobvjk:
> adding a variable as the default value for a parameter in a function signature is not a great idea (e.g. `file_path = log_path` versus `file_path = "/default/path/to/log"`)
> 
> but I see how that makes the most sense here to maintain backward compatibility, and it facilitates moving in the right direction as you have in the rest of this PR adding the file_path/log_path explicitly in the calls to the function

noted for future endeavours

PR #279, opened 2020-10-28
Enhance run-web-tool to show upstream-branch and status

cjyetman:
This should add some output at the top of running the web tool scripts to verify the status of all the need repos. This should help debugging, and help determine differences between runs at different time and/or on different machines.

Does require that the system running it already has `git2r` installed.

PR #275, opened 2020-10-28
setup action to use custom docker image

cjyetman:
first attempt at doing #269

PR #297, opened 2020-10-30
Test R scripts on 2dii runners

maurolepore:
Closes #275 
Closes #279 

This PR adds a workflow to test the web-tool on 2dii runners, i.e. the docker containers we maintain and expect to use in production.

Right now I see the tests fail because of problems with LaTeX dependencies. But the fix should be on the docker containers, not here. 

But life is too short to debug LaTeX. Can we avoid it?


cjyetman:
very strange... I would guess due to the following error messages that `tinytex` the R package is installed, but latex is not installed on the system... but how can that be the case when we built that docker image ourselves and know what's in it and use it successfully elsewhere? very strange

```
tlmgr search --file --global '/xelatex'
181
Cannot determine type of tlpdb from /github/home/texmf!
182
cannot setup TLPDB in /github/home/texmf at /usr/bin/tlmgr line 6424.
183
! sh: 1: xelatex: not found
```

maurolepore:
Yes, very strange. Does this make sense to you?
https://github.com/rstudio/bookdown/issues/474#issuecomment-336494457
I feel it might be useful but I don't understand it.

cjyetman:
I don't understand the beginning of that thread (sounds mostly like user error), but the last few comments seem to be about needing to install the `tinytex` package and then also subsequently running `tinytex::install_tinytex()`, which is true and which we're already doing here in the docker image builder [2DegreesInvesting/2diidockerrunner/blob/master/r-packages/Dockerfile#L51-L52](https://github.com/2DegreesInvesting/2diidockerrunner/blob/master/r-packages/Dockerfile#L51-L52)

Also, the interactive report is rendered with bookdown>rmarkdown>markdown>pandoc>HTML, while the executive summary is rendered with rmarkdown>markdown>pandoc>PDF... so the latex warning is from the executive summary part which does not use bookdown.

maurolepore:
I'll close this to keep the open PR focused on critical stuff. This PR is really-nice-to-have PR but it's stuck in unfortunate LaTeX isues which solution may take more than we can afford.  LaTeX dependencies seem like too much trouble for what we use them to -- i.e. to print a single document.  It'll be great to not print a PDF at all and instead print a lighter format like .html.

PR #292, opened 2020-10-30
Setup LaTeX with tinytex

maurolepore:
Closes #285 (if this pr follows #291 -- so merge #291 first)

Setup LaTeX with tinytex
https://github.com/r-lib/actions

Thanks @cjyetman

PR #291, opened 2020-10-30
Setup pandoc

maurolepore:
Closes #267
Addresses a part of #285 

Thanks @cjyetman

PR #299, opened 2020-10-31
Setup environment to test-r-scripts

maurolepore:
Closes #285
Replaces #291 #292 and #293 combined.All three PRs touched the
same file and are related in that together they create the
environment for test-r-scripts. This combined PR makes it
easier to review the relationship between all changes, and
avoids merge conflicts. All in all, I expect the reiew to
go more smoothly.

### DIFF
--- a/.github/workflows/test-r-scripts.yaml
+++ b/.github/workflows/test-r-scripts.yaml
@@ -23,32 +23,32 @@ jobs:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
 
     steps:
-
-      - uses: r-lib/actions/setup-pandoc@master
-      - uses: r-lib/actions/setup-tinytex@v1
-
       - name: Checkout PACTA_analysis
         uses: actions/checkout@v2
         with:
           path: PACTA_analysis
+
       - name: Checkout pacta-data
         uses: actions/checkout@v2
         with:
           repository: 2DegreesInvesting/pacta-data
           token: ${{ secrets.MAURO_PAT_FOR_2DII }}
           path: pacta-data
+
       - name: Checkout create_interactive_report
         uses: actions/checkout@v2
         with:
           repository: 2DegreesInvesting/create_interactive_report
           token: ${{ secrets.MAURO_PAT_FOR_2DII }}
           path: create_interactive_report
+
       - name: Checkout StressTestingModelDev
         uses: actions/checkout@v2
         with:
           repository: 2DegreesInvesting/StressTestingModelDev
           token: ${{ secrets.MAURO_PAT_FOR_2DII }}
           path: StressTestingModelDev
+
       - name: Checkout user_results
         uses: actions/checkout@v2
         with:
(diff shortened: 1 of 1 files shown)


cjyetman:
amazing, so this seems to work! The strategy here is to use a "normal" worker, not our custom docker image, set it up with pandoc and latex, checkout the necessary repos in it, and then run the tests. I guess that's slightly less than ideal because there are significant differences between that and our intended use, but it's certainly better than nothing and better than anything we've had so far.

One thing I'm curious/worried about. In our existing scripts the this repo gets cloned locally to a directory named "bound", which I've never been a fan of, but I don't see that happening in this script anywhere, so I'm a bit surprised that doesn't cause any problems. Any thoughts on that?

Otherwise, I approve :-)

cjyetman (approved):

maurolepore:
A note about this workflow: Even if we get the container working on gh-actions we'll still need this workflow because container events seem to be supported only for Linux runners -- here we can use not only Linux but also macOS and windows.

maurolepore:
RE /bound, I don't know why transitionmonitor designed the container to monunt the contents of PACTA_analysis/ into /bound/ (maybe for robustness -- as they control the name /bound but not PACTA_analysis?). I guess we could replace `path: PACTA_analysis` with `path: bound`:

https://github.com/2DegreesInvesting/PACTA_analysis/blob/db9b76765d3e20e45bafd3825c79888dc0a77ee4/.github/workflows/test-r-scripts.yaml#L33

But that seems to complicate things more than we need (and like you I also dislike bound it makes the code harder to understand).

Notice we specify `path` for the gh-runner to work from the parent directory of all the sibling repos we clone. So later we `cd` into PACTA_analysis (but -- again -- we could rename it to "bound") via `setwd()`:

https://github.com/2DegreesInvesting/PACTA_analysis/blob/db9b76765d3e20e45bafd3825c79888dc0a77ee4/.github/workflows/test-r-scripts.yaml#L75

.github/workflows/test-r-scripts.yaml
```diff
@@ -27,38 +27,46 @@ jobs:
...
           r-version: ${{ matrix.config.r }}
           http-user-agent: ${{ matrix.config.http-user-agent }}
 
       - name: Install dependencies from CRAN
         run: |
+          install.packages("rmarkdown")
+          install.packages("knitr")
           install.packages("renv")
           dependencies <- renv::dependencies()$Package
           install.packages(dependencies)
```
maurolepore, line 72: Since @cjyetman's review I rearranged the steps to group all R-related stuff. This change of style does not affect behaviour so I'm not asking to review again.

maurolepore:
In the last run I saw an error but seemed temporary, only on macOS, and only on the R CMD check workflow -- not the workflow this PR touches. So I'll go ahead and merge.

cjyetman:
I've seen code spread throughout this network of repos that has stuff like
```r
setwd("../create_interactive_report")
```
based on the assumption of the sibling directory names

which makes me fear that maybe there is/was/will be something like
```r
setwd("../bound")
```
but for now it seems to be working, so that's great

maurolepore:
Yeah, that's a reasonable fear. I wish not to see `setwd()`  nor the hard-coded path to "bound" -- which we might replace with a more flexible/relative approach:

``` r
here::here()
#> [1] "/home/mauro/git/PACTA_analysis"

you_are_here <- fs::path_file(here::here())
you_are_here
#> [1] "PACTA_analysis"

dot_dot_here <- fs::path("..", you_are_here)
dot_dot_here
#> ../PACTA_analysis

setwd(dot_dot_here)
```

<sup>Created on 2020-11-01 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

PR #300, opened 2020-11-01
new action checkout

maurolepore:


maurolepore:
What I'm trying to do here seems to be not supported -- i.e. to call an actions from another action.

https://github.community/t/call-an-action-from-another-action/17524/2

PR #303, opened 2020-11-03
removed reference to old translation file

Clare2D:
This dataframe has been replaced with @jdhoffa's newer translation method and can be deleted. 

This PR removes the reading the of the file and the reference in the create_interactive_report function.


jdhoffa (commented):
I tidy this up already in 2DegreesInvesting/create_interactive_report#115
I would prefer to wait until this gets merged, since I will be adding that argument back in

Clare2D:
Excellent. I've then removed this in https://github.com/2DegreesInvesting/create_interactive_report/pull/128 so this should probably be removed from that PR

PR #305, opened 2020-11-03
fix the path which got messed up in the PR somehow

cjyetman:


cjyetman:
duplicate of #304 

closing

PR #306, opened 2020-11-03
fs "equivalents" are not always equivalent

cjyetman:
`fs::dir_delete` always errors out if it can't find the directory whereas `unlink` does not, so in order to use the `fs` "equivalent" of the base function here, it will have to be wrapped in a `try` since the intended expectation is that the directory may or may not be there


jacobvjk (approved):
lgtm

PR #307, opened 2020-11-03
add a check of current repos' status to test

cjyetman:
just before running all the tests, also print out the current status of all the relevant repos in the output to make it easier to verify what is actually being tested


jacobvjk (approved):
good idea!
LGTM

PR #308, opened 2020-11-03
check for changes on remote origin first

cjyetman:


cjyetman:
had to set the directory for the git fetch, but now it should work as expected

test_test_cases.R
```diff
@@ -59,6 +59,7 @@ report_git_status <-
   function(repo_root = ".") {
     for (i in  repo_root) {
       cli::cli_h1(paste0("repo status for: ", i))
+      system2("git", c("-C", i, "fetch"), stdout = FALSE)
```
maurolepore, line 62: Can you explain a bit what this does? I'm scratching my head :)

maurolepore: Running it in my head give me something that makes little sense:


```bash
git -C <some path> fetch
```

cjyetman: yes, exactly.... and so it runs the git command as if it were run from `<some path>`

https://git-scm.com/docs/git#Documentation/git.txt--Cltpathgt

maurolepore: Amazing!

maurolepore (approved):

Issue #309, opened 2020-11-03
Remove filter for small EQ holdings


jacobvjk:
https://github.com/2DegreesInvesting/PACTA_analysis/blob/0e03b064dba89bbb69cd636b75aec1a35eaee41a/3_run_analysis.R#L37

We introduced this filter as a hack to get rid of some duplicate values, which are now handled more appropriately via a dedicated list of duplicate ISINs.

This filter should be removed, as it can in specific cases lead to valid equity holdings being excluded from the analysis. This was the case for EIOPA, when splitting the portfolio up into a large number of subgroups. While that is not a very common case, there is no reason for this to stay in the code and be a hazard for other future applications

fyi @Clare2D

jacobvjk:
also: this is not an issue in P2020, as the filter is not set there

Clare2D:
Just want to clarify that this was removed to allow the code to run faster and not hit memory issues rather than to remove the duplicates!

jacobvjk:
Okay, good you pointed that out. We can discuss this tomorrow after the deadline if you want

PR #310, opened 2020-11-03
remove obsolete/hazardous filter on eq portfolio

jacobvjk:
This PR:

- removes the filter on the equity part of the portfolio that removes holdings with a very small weight
- this was originally a hack to get rid of duplicates but is long superseded by explicitly handling duplicates in a dedicated list of duplicate ISINs
- in rare cases, this filter can erroneously remove valid equity holdings, so it should be removed
- this change does not touch pacta 2020 in any way, as the filter was only set in the offline version

closes: https://github.com/2DegreesInvesting/PACTA_analysis/issues/309

### DIFF
--- a/3_run_analysis.R
+++ b/3_run_analysis.R
@@ -34,8 +34,6 @@ if (file.exists(equity_input_file)) {
 
     port_eq <- calculate_weights(port_raw_eq, "Equity", grouping_variables)
 
-    port_eq <- port_eq %>% filter(port_weight > 1e-6)
-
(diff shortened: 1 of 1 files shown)


maurolepore (approved):
LGTM. The change reflects the intent, it's explained as safe, and @Clare2D participated in the issue. I won't merge but @jacobvjk you seems confident enough to do so.

Clare2D:
@cjyetman this could be a solution

cjyetman:
this was merged 14 days ago, so whatever's here is already in there

PR #311, opened 2020-11-03
New arg --fetch (needs access to private repos)

maurolepore:


cjyetman (approved):
I don't understand under what conditions one would want to not fetch or at least attempt to, but technically this look ok.

maurolepore:
Thanks for your review. 
So far I did all what we discussed on the call with one big exception: The repos are still copied from the local system -- not cloned from GitHub. (I hope to implement that but this seems like the first step).

So, the container does not have any credentials to clone private repos. If we fetch form the container, it fails.
We can still ask git or a status of each repo at it is locally.

PR #312, opened 2020-11-03
Mount and run web tool

maurolepore:


cjyetman (approved):
nit: I wonder if it's possible to achieve this with `./bin/show-upstream-and-status-at-pacta-repos $1`

otherwise, LGTM

maurolepore:
Yeah, the names of the commands got a lil crazy. One alternative would be to remove the old `run-web-tool` (assuming we no longer want to mount) and replace it with what is no named `run-web-tool-from-local-copy`. 

But you suggest to call `./bin/run-r-scripts` from inside `./bin/show-upstream-and-repos-at-pacta-repos`?

If that's the case, then I think the name of the command would not reflect the "run r scripts" part.

I'd like to discuss a bit more because surely I didn't get the names right yet.

cjyetman:
no, I'm just wondering if you can use `./` in place of `"$(pwd)"/`

`"$(pwd)"/` seems more error prone because it's running a command (`pwd`) using that as an in place environmental variable (`$(...)`) and then quoting it in case there's whitespace in whatever it evaluates to (`"..."`)... seems like just `./` would be easier, safer, and get the job done

maurolepore:
Oh, ok -- thanks! But that won't work if we run on our own systems, right? I'm trying to write a command that works both on our host machines and on the containers we run.

maurolepore:
Oh, yes, it should work! Let me try

cjyetman:
as far as I know `./` refers to your current working directory on just about every computer system

cjyetman:
actually, sorry, to be more precise...

`./` is a **relative** path to your current working directory
`pwd` probably is more likely to give you an absolute path to your current working directory

I guess it's possible under some very specific situations where that might make a difference, but for the most part they should be equivalent (unless/until you change your current working directory... obviously?).

maurolepore:
Thanks! My experiments suggest it does work as you say. I pushed a last commit that does that. Still I'm not very happy with the names but it may be best to follow up on a separate PR -- for example `start-conainer` is now ambiguous as we used to have only one container (the one that mounted the pacta family) but right now we have two: the old one plus the one that copies the repos (not mount).

maurolepore:
CJ, please have another look.

cjyetman:
is this change desired?
`"$(pwd)"/bin/show-upstream-and-status-at ../PACTA_analysis $1`
to
` ./bin/show-upstream-and-status-at . $1`

i.e. `../PACTA_analysis` to `.`

I'm not in a position to actually test this right now, but that seems a bit weird to me

maurolepore:
Yes, I think "../PACTA_analysis" won't work in the container -- where it's instead it should be ../bound.
But since we expect to run these scripts form inside PACTA_analysis(/bound), then we can just refer to the current working directory with ".". This change allows the script to work on both the container and local host.

```bash
➜  PACTA_analysis git:(master) ./bin/show-upstream-and-status-at .  
PACTA_analysis
Mauro Lepore commited 4 hours ago:
master 3b819f0 [origin/master] Mount and run web tool (#312)
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
```

PR #315, opened 2020-11-04
change na to 0

Clare2D:
This deals with an issue that arises when the plan_emission_factor is NA and the scen_emission_factor == 0. This is aggregated to scenario level where it then sums to 0 because na.rm == T. When this is summed, the NA is ignored, where as the 0 is used in a weighted averaged. And the plan and scenario values differ for year == start_year which is not the accurate case. 

This resolves that issue. Could be solved somewhere earlier however, this is a reasonable solution. If the scenario emission factor is 0 then the plan emission factor can also be 0.


jacobvjk (approved):
trusting you there content wise, code LGTM

PR #317, opened 2020-11-04
Assert input to green_brown() has length 1

maurolepore:
Relates to https://github.com/2DegreesInvesting/create_interactive_report/pull/177

This seems not urgent. 

I see two functions with the name `green_brown()` -- one here and the other one in create_interactive_report -- and they are not identical. The version in create_interactive_report seems more developed but I can't tell which version should win. Right now the web tool seems to be using the version from create_interactive_report.

Just in case the version here gets called somewhere, I added an assertion to detect the problem that https://github.com/2DegreesInvesting/create_interactive_report/pull/177 fixed.

### DIFF
--- a/0_graphing_functions.R
+++ b/0_graphing_functions.R
@@ -757,9 +757,11 @@ graph_name <- function(plotnumber, ParameterFile, explicit_filename = "") {
   return(graphname)
 }
 
+# FIXME: There is another green_brown() in the repo create_interactive_report and
+# they are not identical. Which one should win?
 green_brown <- function(Tech) {
   GreenTechs <- c("Electric", "Hybrid", "RenewablesCap", "HydroCap", "NuclearCap")
-
+  stopifnot(length(Tech) == 1L)
   if (Tech %in% GreenTechs) {
     TechIs <- "green"
   } else {


Clare2D:
What is the status on this PR?

maurolepore:
It's ready but in draft because it's not crucial. I think if could be reviewed and merged as soon as the web tool goes live.

maurolepore:
Now that we are past the initial release, I think this is a good time to make this function a little more strict and to ask around if anyone needs this copy -- before I remove it because it looks like the version we actually use is not this one but the one in the repo create_interactive_report.

Clare2D:
Yes this is a common check that we need to do. Perhaps we need to move this to a more Global location? What I also wanted to do/did at some stage was move the list of "green techs" to be an input variable rather than being defined within the function. Setting this list as a default list however making it more explicit which values are being considered there.

maurolepore:
> Perhaps we need to move this to a more Global location? 

Good point. The limits of PACTA_analysis  are quite blurry. I hope we can narrow its scope and then we'll more clearly see where `green_brown()` should live. 

> What I also wanted to do/did at some stage was move the list of "green techs" to be an input variable rather than being defined within the function. Setting this list as a default list however making it more explicit which values are being considered there.

It seems that's done in the version that we actually use:

https://github.com/2DegreesInvesting/create_interactive_report/blob/06c6f72136d47f74c3eecc51841b5fa2747557f8/create_interactive_report.R#L2206

(I remind you that the version in this PR seems to be used nowhere, so I would merge this PR just to get it out of the way. Once we have a few more tests, I suggest we just delete this funcition here as it seems to be dead code replaced by the version in create_interactive_report.)

maurolepore:
@cjyetman, can you review so we decide if we merge, close, or edit this PR?

cjyetman (approved):
sorry, this fell off my radar

this is a good check, so let's merge it

that question of whether or not this function is needed at all will need to be addressed eventually, but might as well make it safer as is for the moment

PR #318, opened 2020-11-05
Improve emissions data

Clare2D:
This changes the way of calculating the emissions data, ignoring 2dii values and rather using the average values we have available. 

This function
- merges in the emissions factors based on the company bics_sector
- where this is in the financial sector of a pact sector, this is identified
- where there is missing bics_sector info, but there is pacta sector, a mean sector value is applied
- the mean value is used to calculate an average financed emissions

This is late for this change - but the data is now just being used with a different priority given the quality of the emissions factors at aggregate were not sufficient.


jdhoffa (approved):
hard for me to know if the maths are correct, but it looks reasonable, and I see no new warnings/ errors when running the docker so ¯\_(ツ)_/¯

PR #320, opened 2020-11-05
enable running as non-user when doing latex

cjyetman:
this is required so that when the docker is run with this mysterious non-user 1000 it can still load its .Rprofile and get the right settings to enable rendering through latex

### DIFF
--- a/bin/run-r-scripts
+++ b/bin/run-r-scripts
@@ -5,11 +5,6 @@ umask 000
 
 cd /bound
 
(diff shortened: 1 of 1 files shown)


Clare2D:
I'll let Mauro review this one..

Clare2D (approved):
Trusting CJ on this one-

maurolepore (commented):
LGTM

PR #323, opened 2020-11-07
New file .pacta helps locates pacta repos

maurolepore:
Inspired by the package 'here', this PR helps define membership to the pacta ecosystem via a root file '.pacta' -- discoverable with 'locate' (and our new wrapper [`pacta-find`](https://github.com/2DegreesInvesting/pacta-cli/blob/master/pacta-find)). 

This aliviates the problem of having not just one but an ecosystem of related packages, because we can now discover automatically where in a system the members live -- regarless of their name. This is fundamental to building tools that work on all repositories at once (command-line applications, R functions, docker images).

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -47,3 +47,4 @@
 ^bin$
 ^docker$
 ^\.renvignore$
+^\.pacta$


cjyetman:
FYI... the `locate` options you use in `pacta-locate` are GNU (Linux) specific and do not work on BSD (macOS)... not sure what Bash shells on Windows are based on.

cjyetman:
Also, while I like this idea, I wonder if it needs to be more nuanced then simply searching your entire drive for `.pacta` files... like maybe only sub directories of a given directory? I currently have more than one clone of a few of these repos on my local drive, for one reason or another, and I can imagine that whatever I'm doing with this (family of?) tools, I might want that to be specific to a set of repos on my drive, but not every copy it can possibly find on my drive.

maurolepore:
Thanks @cjyetman, you idea is a much better one. I now use `pacta-find`. As you suggest, it takes the path to the parent of pacta repos  (or defaults to the current working directory). It still looks for an empty file called ".pacta" in the root of every directory under the given path (parent of pacta siblings) -- meaning that this PR is still necessary.

Let me know what you think and if you can try [pacta-find](https://github.com/maurolepore/bin/blob/master/pacta-find) on macos -- I think it should work because I use commands I believe come by default on all Unix systems.

```bash
➜  ~ pacta-find ~/git                                   
/home/mauro/git/create_interactive_report
/home/mauro/git/PACTA_analysis
/home/mauro/git/pacta-data
/home/mauro/git/StressTestingModelDev
/home/mauro/git/user_results

➜  ~ cd git  

➜  git pacta-find
/home/mauro/git/create_interactive_report
/home/mauro/git/PACTA_analysis
/home/mauro/git/pacta-data
/home/mauro/git/StressTestingModelDev
/home/mauro/git/user_results
```

cjyetman (approved):

maurolepore:
Thanks! I'll work in batch to celebrate:

image

PR #327, opened 2020-11-09
better checks on the needed repos

cjyetman:
improvements to the checks that check the status of all the needed repos before running the tests

### DIFF
--- a/test_test_cases.R
+++ b/test_test_cases.R
@@ -56,20 +56,52 @@ for (i in seq_along(test_cases_csvs)) {
 
 
 report_git_status <-
-  function(repo_root = ".") {
-    for (i in  repo_root) {
-      cli::cli_h1(paste0("repo status for: ", i))
-      system2("git", c("-C", i, "fetch"), stdout = FALSE)
-      print(git2r::branch_get_upstream(git2r::repository_head(i)))
-      print(git2r::status(i))
+  function(repo_roots = ".") {
+    for (repo_root in  repo_roots) {
+      cli::cli_h1(paste0("repo status for: ", repo_root))
+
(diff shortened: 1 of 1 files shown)


test_test_cases.R
```diff
@@ -56,20 +56,52 @@ for (i in seq_along(test_cases_csvs)) {
...
+      repo <- git2r::repository(repo_root)
+      local_head <- git2r::revparse_single(repo, "HEAD")
+      upstream_head <- git2r::revparse_single(repo, "origin/HEAD")
+      local_repo_head <- git2r::repository_head(repo)
+      upstream_branch <- git2r::branch_get_upstream(local_repo_head)
+      remote_url <- git2r::branch_remote_url(upstream_branch)
+      repo_ahead_behind <- git2r::ahead_behind(local_head, upstream_head)
+
+      cat(paste0("Upstream branch: ", upstream_branch$name, " from ", remote_url), "\n")
+      if (!all(repo_ahead_behind == 0)) {
```
maurolepore, line 74: Can you confirm you want all()? In my head I need any().

maurolepore (commented):
Please check one if statement and confirm you want all() and not any().
Except for that LGTM

test_test_cases.R
cjyetman: yes... `repo_ahead_behind` will be a 2 element vector, and I want *both/all* to be 0, otherwise that means something is not in sync and I should report what the difference is

maurolepore: Thanks

maurolepore (approved):

maurolepore:
LGTM — approved but please merge on your end. My brain is still off

PR #328, opened 2020-11-09
create empty audit & emissions file

Clare2D:
This will allow real estate portfolios to print even when there is no audit /emissions info available

### DIFF
--- a/web_tool_script_3.R
+++ b/web_tool_script_3.R
@@ -26,10 +26,10 @@ source(file.path(template_path, "create_executive_summary.R"))
 source(file.path(template_path, "useful_functions.R"))
 
 report_name = select_report_template(project_code = project_code,
-                       language_select = language_select)
+                                     language_select = language_select)
 
 exec_summary_name = select_exec_summary_template(project_code = project_code,
-                                           language_select = language_select)
+                                                 language_select = language_select)
 
 template_dir <- paste0(template_path, report_name,"/_book/")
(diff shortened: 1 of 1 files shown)


web_tool_script_3.R
```diff
@@ -45,8 +45,22 @@ scenario_other <- "ETP2017_B2DS"
...
 scenario_geography <- "Global"
-audit_file <- read_csv(file.path(proc_input_path, portfolio_name_ref_all, "audit_file.csv"), col_types = cols())
-emissions <- read_rds(file.path(proc_input_path, portfolio_name_ref_all, "emissions.rda"))
+
+if (file.exists(file.path(proc_input_path, portfolio_name_ref_all, "audit_file.csv"))){
+  audit_file <- read_csv(file.path(proc_input_path, portfolio_name_ref_all, "audit_file.csv"), col_types = cols())
+}else{
+  audit_file <- tibble("investor_name" = NA_character_, "portfolio_name" = NA_character_,
+                       "asset_type" = NA_character_, "valid_input" = NA, "isin" = NA_character_,
+                       "direct_holding" = NA, "value_usd" = NA_real_)
```
cjyetman, line 54: probably safest to use an explicitly typed `NA` value for `valid_input` and `direct_holding`

cjyetman (approved):
you may want to use explicitly typed NAs here (see my comment), but otherwise this lgtm

PR #332, opened 2020-11-16
removes loop that read in all files

Clare2D:
This simplifies the read in to remove the chance of reading in other files in the folder. Wim passes a csv file renamed to be called portfolio_name_ref_all.csv which is the value passed to the command line.


jacobvjk (approved):
Tested by placing a full portfolio and a subset of it with diff filenames and param files in `working_dir/10...` and `working_dir/20...` and adapting the default portfolio_name_ref_all according to the case.
The output behaved as expected producing results for each variant, including the right coverage stats

--> LGTM

PR #334, opened 2020-11-17
Fix original scripts

Clare2D:
This PR makes changes to the original scripts to ensure that these can be run for offline projects. 

There is an open question as to whether we switch to allow the html reports also to be created, or whether we stick with the pdfs. For this PR I suggest we stick with PDF. 

@jacobvjk I've got some open questions as to changes in the fin data and sector classifications and whether we can easily incorporate the stress testing or not.

### DIFF
--- a/0_global_functions.R
+++ b/0_global_functions.R
@@ -82,10 +82,10 @@ set_global_parameters <- function(file_path) {
 
   equity_market_list <<- cfg$Lists$Equity.Market.List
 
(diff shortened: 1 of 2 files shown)


jacobvjk:
@Clare2D do you want to discuss this next week in the bubble? Otherwise happy to go into details e.g. tomorrow

Clare2D:
Next week works for me.

Clare2D:
Finally pushed a different version of this

PR #339, opened 2020-11-20
REVERTME: Test CI detects error in web_tool_script_1.R

maurolepore:


maurolepore:
Good. My experiment suggests CI is still sensitive to errors in script 1-3. I was hesitant because I haven't seen that fail in a while.

PR #340, opened 2020-11-20
Add more tests

maurolepore:
Recently I discussed with CJ a few tests we could add to our integration pipeline. I thought about that quite a bit while developing the now gone [integration-test](https://github.com/2DegreesInvesting/PACTA_analysis/blob/add-tests/integration-test.md), but that file may have some ideas I may be able to reuse.

### DIFF
--- a/tests/testthat/test-test_r_scripts.R
+++ b/tests/testthat/test-test_r_scripts.R
@@ -19,13 +19,7 @@ test_that("example data is available under sample_data/", {
   expect_true(fs::file_exists(example_data))
 })
 
-test_that("example data is available under sample_data/", {
-  # FIXME: When this file if unavailable the error message is unhelpful
-  # Error in get_input_files(portfolio_name_ref_all) :
-  #   Difference in input files and input argument portfolio names.
-  # Execution halted
-  # Error: This script threw an error.
-  # TODO: Fail fast with a more informative error message.
+test_that("example data is available under working_dir/", {
   example_data <- here::here(
     "working_dir", "20_Raw_Inputs", "TestPortfolio_Input.csv"
   )
--- a/tests/testthat/test-test_r_scripts.R
+++ b/tests/testthat/test-test_r_scripts.R
@@ -10,7 +10,7 @@ test_that("example data is available under sample_data/", {
   # At https://bit.ly/35Xypsw I read:
   # > copy TestPortfolio_Input.csv from sample_files/20_input_files/ to
   # > working_dir/20_Raw_Inputs/
-  # Do we copy that file manually or programatically? If so, where is the code?
+  # Do we copy that file manually or programmatically? If so, where is the code?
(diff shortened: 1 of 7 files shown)


tests/testthat/test-test_r_scripts.R
```diff
@@ -4,3 +4,25 @@ test_that("stop_on_error stops on error", {
...
+  # > copy TestPortfolio_Input.csv from sample_files/20_input_files/ to
+  # > working_dir/20_Raw_Inputs/
+  # Do we copy that file manually or programmatically? If so, where is the code?
+  # My experiments suggest web_tool_script_1.R runs to the end without it, and
+  # I need to know it that's okay of if I should throw an error ASAP.
+  example_data <- here::here(
+    "sample_files", "20_input_files", "TestPortfolio_Input.csv"
+  )
+  expect_true(fs::file_exists(example_data))
+})
```
maurolepore, line 20: @jacobvjk, this PR is still WIP but this section is ready for your feedback.

tests/testthat/test-test_r_scripts.R
```diff
@@ -4,3 +4,25 @@ test_that("stop_on_error stops on error", {
...
+
+test_that("example data is available under working_dir/", {
+  example_data <- here::here(
+    "working_dir", "20_Raw_Inputs", "TestPortfolio_Input.csv"
+  )
+  expect_true(fs::file_exists(example_data))
+})
```
maurolepore, line 27: @cjyetman, today we discussed ways to access the file system from tests; one thing we didn't consider is `here::here()` and it looks like it might help!

cjyetman: Thanks. I am using `here::here()` already. I think that's the only way it possibly works. I have been using it at the top of every test script inside a `setwd()` though, and this usage is a lot better, so thanks for that!

tests/testthat/test-test_r_scripts.R
cjyetman: manually.... `TestPortfolio_Input.csv` is an example of what a user's portfolio input file would look like... they are expected to put that file into the`20_Raw_Inputs/` directory under their "working directory" (which could be named different than "working_dir")

I would expect `web_tool_script_1.R` to work as expected without `sample_files/20_input_files/TestPortfolio_Input.csv`

I would expect `web_tool_script_1.R` to NOT work as expected without `working_dir/20_Raw_Inputs/TestPortfolio_Input.csv` OR another csv file in the same directory with a filename that matches a YML file's filename in `10_Parameter_File` and has `portfolio_name_in` inside the YML set to something that matches, e.g. `working_dir/10_Parameter_File/TestPortfolio_Input_PortfolioParameters.yml` contains `portfolio_name_in: TestPortfolio_Input`

maurolepore:
I'm splitting this PR.

PR #342, opened 2020-11-24
add na.rm to port weight calc

Clare2D:
Noticed a missing na.rm on the calculation of the port weight for the CAG. This should fix errors if nas come through the data.


cjyetman (approved):
Nit: prefer `TRUE` instead of `T`

comment: we should probably check universally across the repo for stuff like this... it's come up enough times

otherwise, lgtm

PR #347, opened 2020-11-24
New function to skip check() but run test()

maurolepore:
https://github.com/2DegreesInvesting/PACTA_analysis/pull/340#discussion_r529172294 wasn't exactly right. My experiments suggest that we can use `here::here()` to reach the source only during `devtools::test()` but not during `devtools::check()`. Here I add a new test helper to skip `check()` but run `test()`; while it's not ideal to skip `check()`, being able to at least test seems useful.

### DIFF
--- a/tests/testthat/test-utils.R
+++ b/tests/testthat/test-utils.R
@@ -15,3 +15,14 @@ test_that("use_r_packages() puts listed packages on search() path", {
     expect_true(is_attached(package))
   }
 })
+
+test_that("helper skip_check_but_run_test() works as expected", {
+  # This test should not be skipped during devtools::test()
+  # This test should be skipped during devtools::check()
+  # If you remove `skip_check_but_run_test()`, devtools::check() should error
+  skip_check_but_run_test()
+
+  # This file should be available only when the source code is accessible, i.e.
+  # not during R CMD check, when the code runs from the system installation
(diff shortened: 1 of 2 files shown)


cjyetman (approved):
don't really understand how it works, but it looks cool.... maybe you can show me how to use it because it sounds useful on my PR branch

tests/testthat/test-utils.R
```diff
@@ -15,3 +15,14 @@ test_that("use_r_packages() puts listed packages on search() path", {
     expect_true(is_attached(package))
   }
 })
+
+test_that("helper skip_check_but_run_test() works as expected", {
+  # This test should not be skipped during devtools::test()
+  # This test should be skipped during devtools::check()
+  # If you remove `skip_check_but_run_test()`, devtools::check() should error
+  skip_check_but_run_test()
```
maurolepore, line 23: Any test that depends on `here::here()` should pass tests but would throw an error during `devtools::check()`. That means that the checks will fail on CI. To avoid that error you can skip the problematic test by adding `skip_check_but_run_test()` inside the call to `test_that()` -- as I do here.

Happy to run a little demo.

cjyetman: 9accae89df39d3768067b089d296f4bb436605ad
thanks!

PR #348, opened 2020-11-24
Test test_r_scripts: example data exists

maurolepore:
This PR tests that the expected input data exists. The test
is quite abstract because I tried to avoid writing the
actual path to the file and instead get it via an interface.
My intent is to have a single place where to update the
path when it changes. Right now the value is unclear because
the path is used only once, but I it would pay off if we
do reuse the interface in a few places. I should probabbly
follow up with a PR replacing hard-coded paths with calls
to this interface.﻿

### DIFF
--- a/R/path_from.R
+++ b/R/path_from.R
@@ -1,0 +1,12 @@
+path_from <- function(from = NULL, ...) {
+  if (is.null(from)) return(here::here(...))
+  here::here(from, ...)
+}
+
+raw_inputs <- function() {
+  list("working_dir", "20_Raw_Inputs")
+}
+
+parameter_file <- function() {
+  list("working_dir", "10_Parameter_File")
+}
--- a/tests/testthat/test-test_r_scripts.R
+++ b/tests/testthat/test-test_r_scripts.R
@@ -4,3 +4,14 @@ test_that("stop_on_error stops on error", {
   expect_error(stop_on_error(exit_code = 1), "error")
   expect_error(stop_on_error(exit_code = 99), "error")
 })
+
+test_that("example data is available under working_dir/", {
+  # https://git.io/JkX5O
+  config <-
+    path_from(parameter_file(), "TestPortfolio_Input_PortfolioParameters.yml")
+  portfolio_name <- get_param("parameters", "portfolio_name_in")(config)
+
+  example_data <- path_from(raw_inputs(), glue::glue("{portfolio_name}.csv"))
+  expect_true(fs::file_exists(example_data))
+})
+
--- a/tests/testthat/test-path_from.R
+++ b/tests/testthat/test-path_from.R
@@ -1,0 +1,11 @@
+test_that("is sensitive to `from`", {
+  expect_equal(
+    path_from("a", "b", "c"),
+    here::here("a", "b", "c")
+  )
+
+  expect_equal(
+    path_from(),
+    here::here()
+  )
+})


R/path_from.R
```diff
@@ -0,0 +1,12 @@
+path_from <- function(from = NULL, ...) {
+  if (is.null(from)) return(here::here(...))
+  here::here(from, ...)
+}
+
+raw_inputs <- function() {
+  list("working_dir", "20_Raw_Inputs")
```
maurolepore, line 7: My intent is for this to be the only place that knows the exact path to this weird directory.

R/path_from.R
```diff
@@ -0,0 +1,12 @@
...
+}
+
+parameter_file <- function() {
+  list("working_dir", "10_Parameter_File")
```
maurolepore, line 11: My intent is for this to be the only place that knows the exact path to this weird directory.

tests/testthat/test-test_r_scripts.R
```diff
@@ -4,3 +4,14 @@ test_that("stop_on_error stops on error", {
...
+
+test_that("example data is available under working_dir/", {
+  # https://git.io/JkX5O
+  config <-
+    path_from(parameter_file(), "TestPortfolio_Input_PortfolioParameters.yml")
+  portfolio_name <- get_param("parameters", "portfolio_name_in")(config)
+
+  example_data <- path_from(raw_inputs(), glue::glue("{portfolio_name}.csv"))
+  expect_true(fs::file_exists(example_data))
+})
```
maurolepore, line 16: Here I sacrificed readability in order to try avoid hard-coding paths. I'm not 100% sure this is the right decision and criticism is welcome.

maurolepore: My intent here is to test something that the web-tool needs without actually touching the web-tool script. This would be probably be best replaced (or complemented) with an assertion in the web-tool script itself (and I'll follow up with a PR) but this seems the safest move until we have more tests covering this code-base.

maurolepore: I'm also considering testing not that the file exists at a given path but instead that the data is available. That way the test should better resist a refactoring by which we become independent of the file system and instead use a more portable way to interface between the different processes.

cjyetman:
honestly, I'm really confused about this PR, how it works, and what it's intended to improve... and that makes me hesitate. I can understand abstracting hard coded paths into a single utility function somewhere, but, for instance, why would they return a list of directory names instead of returning a valid path to the thing they are meant to point to?

At a quick glance, it seems to me that the `path_from()` function's first argument is set to `NULL` by default and then that's tested for within the function, but `from` will never equal `NULL` inside the function when it runs because the first passed argument will be assigned to `from` based on argument order, unless the call is...
1. `path_from()` which would seem to have no purpose
2. `path_from(from = NULL)` or `path_from(from = NULL, "some", "other", "values")` which seems unlikely and ugly

following from that, it doesn't seem to do anything useful beyond what `here::here()` already does on its own.

I also question the usage of `here::here()` in the context of a package. As I understand it, [it's designed for "project-oriented workflows"](https://here.r-lib.org). I know this repo is a mix of project and pkg, but scripts that live in `R` are clearly intended to work as if part of a pkg, and should probably do so.

I worry about adding code to these repos that's too fancy, too sophisticated, too complicated for the average coder at 2dii to easily understand, maintain, and use. If it takes me 20 minutes to wrap my head around what's going on with a few lines of code, I really worry that it's going to add more code in these repos that most people are afraid to touch, which is hopefully the situation we're trying to get away from.

maurolepore:
Thanks for your thoughtful review. Fair enough. This is hard to understand and thus hard to maintain. The more direct approach of #349 seems better.

PR #349, opened 2020-11-24
Assert: fail fast if example data doesn't exist

maurolepore:
Relates to #348

This PR adds an assertion to the web-tool script 1.﻿ This tests the
same as #348 but this is more intrusive and has the potential to
disrupt existing code. Instead, #348 tests the same condition but
not on the production environment but only on the test environment.

It would be nice to discuss this different strategies to testing and
decide if we now want one or both.


web_tool_script_1.R
```diff
@@ -95,6 +95,11 @@ if (new_data == TRUE) {
 ####################
 #### PORTFOLIOS ####
 ####################
+abort_if_file_doesnt_exist(
+  here::here(
+    "working_dir", "20_Raw_Inputs", glue::glue("{portfolio_name_ref_all}.csv")
```
maurolepore, line 100: Here I see an opportunity to reuse `path_from(raw_inputs(), ...)` from https://github.com/2DegreesInvesting/PACTA_analysis/pull/348

cjyetman: why does `abort_if_file_doesnt_exist()` get defined twice in two different scripts?

cjyetman: never mind... haven't had my coffee yet ☕

web_tool_script_1.R
```diff
@@ -95,6 +95,11 @@ if (new_data == TRUE) {
...
+  )
+)
 portfolio_raw <- get_input_files(portfolio_name_ref_all)
```
cjyetman, line 103: I do think it might be better if `abort_if_file_doesnt_exist()` were integrated into `get_input_files()` rather than being run separately... but that opens a can of worms because `get_input_files()` is currently defined in `0_web_functions.R`

R/abort_if_file_doesnt_exist.R
```diff
@@ -0,0 +1,17 @@
...
+#' # Cleanup
+#' fs::file_delete(path)
+#' @noRd
+abort_if_file_doesnt_exist <- function(path) {
+  if (!fs::file_exists(path)) {
+    rlang::abort(glue::glue("This file must exist but it doesn't:\n{path}"))
+  }
+
+  invisible(path)
+}
```
cjyetman, line 17: honestly, this seems overly dependent heavy... 3 dependent packages to achieve `stopifnot(file.exists(path))`?

cjyetman (approved):

R/abort_if_file_doesnt_exist.R
```diff
@@ -0,0 +1,18 @@
...
+#' fs::file_delete(path)
+#' @noRd
+abort_if_file_doesnt_exist <- function(path) {
+  if (!file.exists(path)) {
+    msg <- sprintf("This file must exist but it doesn't:\ns%", path)
+    stop(mdg, call. = TRUE)
+  }
+
+  invisible(path)
+}
```
maurolepore, line 18: This is now refactored to use only base R.

R/abort_if_file_doesnt_exist.R
maurolepore: Good point. I refactored it to use only base R. See below.

R/abort_if_file_doesnt_exist.R
```diff
@@ -0,0 +1,18 @@
...
+#' fs::file_delete(path)
+#' @noRd
+abort_if_file_doesnt_exist <- function(path) {
+  if (!file.exists(path)) {
+    msg <- sprintf("This file must exist but it doesn't:\n%s", path)
+    stop(msg, call. = TRUE)
+  }
+
+  invisible(path)
+}
```
maurolepore, line 18: Here is the pure base-r implementation

R/abort_if_file_doesnt_exist.R
```diff
@@ -0,0 +1,18 @@
...
+#' abort_if_file_doesnt_exist(path)
+#'
+#' # Cleanup
+#' fs::file_delete(path)
+#' @noRd
+abort_if_file_doesnt_exist <- function(path) {
+  if (!file.exists(path)) {
+    msg <- sprintf("This file must exist but it doesn't:\n%s", path)
+    stop(msg, call. = FALSE)
+  }
```
maurolepore, line 15: Here is now the base-R implementation.

web_tool_script_1.R
maurolepore: I agree that in the long run `get_input_files()` could itself assert inputs. But because we have so few tests, right now I prefer to add changes in a way that minimize their effects on the system.

maurolepore:
I see one run fails about latex. This seems unrelated to this PR so I'll go ahead and merge

```
Error: Error: LaTeX failed to compile /home/runner/work/PACTA_analysis/PACTA_analysis/PACTA_analysis/working_dir/50_Outputs/TestPortfolio_Input/executive_summary/template.tex. See https://yihui.org/tinytex/r/#debugging for debugging tips. See template.log for more info.
```

cjyetman:
This LaTeX error is strange and only happens on one virtual machine, but let's keep an eye on that because any LaTeX mess makes me nervous. It seems to have been triggered by this `! Package xcolor Error: No driver specified.`. No idea what that means. If we see that again, I think we better take care of that sooner rather than later.

Issue #331, opened 2020-11-16
Clean up creating empty data frames web_tool_script_3


Clare2D:
In the web_tool_script_3 we create empty data frames if there is nothing to read in. There could be a simpler and shorter method for doing this.

Issue #214, opened 2020-10-15
Propagate ST params from stress test directly


jacobvjk:
Nice! Magic number are ugly.

_Originally posted by @maurolepore in https://github.com/2DegreesInvesting/PACTA_analysis/pull/209#discussion_r505515025_

such variables should not be hard-coded in the web tool but rather be read in some way fromthe ST output

PR #350, opened 2020-11-25
Clean web tool script 3

Clare2D:
This is a significant PR with a few different functions. @jacobvjk this will allow us to have stress test functions across the scripts. 

Aims:
- Puts project parameters into a separate file so that we can have different variables for Swiss, General, other countries etc. (#214 
- Add empty results functions moving tibbles into functions (#331 )
- Creates function to read in project parameters 
- specifies which report should be read in (https://github.com/2DegreesInvesting/create_interactive_report/issues/271)
- Changes Peer File name reference 

Needs to be run in conjunction with updates to pacta-data that will be turned into a PR soon. 
https://github.com/2DegreesInvesting/pacta-data/pull/19
https://github.com/2DegreesInvesting/create_interactive_report/pull/276
https://github.com/2DegreesInvesting/StressTestingModelDev/pull/121


jacobvjk (commented):
Really happy to see this. lgtm, but I won't approve, as I think CJ has a better understanding of the project parameters.

Clare2D:
This also merges the AnalysisParameters file into the new Projectparameters_xxxx fil This was discussed with @cjyetman and will hopefully remove confusion. For each PACTA run what needs to be defined:
- Project Parameters (apply for all participants)
- Portfolio Parameters (unique to a single participants - defines only which project they are participating in)

cjyetman:
this works for me locally with everything as is

1. didn't we discuss removing completely `AnalysisParameters.yml` from the root level?

2. when I change the project code in `working_dir/10_Parameter_File/TestPortfolio_Input_PortfolioParameters.yml` to "PA2020FL", everything seems to work as expected except that the Executive Summary is still triggered, for which there is currently no template for in `create_interactive_report`, and thus triggers an error... is that expected, or do we need to prevent that from happening (i.e. will there eventually be a template for that in `create_interactive_report`)?

Clare2D:
Thanks CJ. I've now removed those remaining AnalysisParameter files and added in some initial exec summary files for Liechtenstein. Otherwise the error is to be expected.

cjyetman:
cool... then I'm gonna approve these PRs... up to you to merge...
2DegreesInvesting/PACTA_analysis/pull/350
2DegreesInvesting/pacta-data/pull/19
2DegreesInvesting/create_interactive_report/pull/276
2DegreesInvesting/StressTestingModelDev/pull/121

cjyetman (approved):

Clare2D:
Woohoo! Thanks.

PR #352, opened 2020-12-02
improve the performance of running the 3 web tool scripts in succession

cjyetman:
these fixes are intended to improve the performance of running the 3 web tool scripts in succession, offline to manually generate reports (which unfortunately I've had to do many times, and more unfortunately looks like I will have to continue doing in the near future)...

1. uses `base::saveRDS` instead of `read::write_rds`, because the tidyfolk made this opinionated decision for all of us that storage space is cheaper than the time it takes to uncompress a file
2. remove objects at the end of the first 2 scripts that can be VERY large but are not needed anymore
3. don't write or read in the audit object as a CSV because it can be MASSIVE


web_tool_script_3.R
```diff
@@ -37,8 +37,8 @@ real_estate_dir <- path(user_results_path, project_code, "real_estate")
 
 output_dir <- file.path(outputs_path, portfolio_name_ref_all)
 
-if (file.exists(file.path(proc_input_path, portfolio_name_ref_all, "audit_file.csv"))){
-  audit_file <- read_csv(file.path(proc_input_path, portfolio_name_ref_all, "audit_file.csv"), col_types = cols())
+if (file.exists(file.path(proc_input_path, portfolio_name_ref_all, "audit_file.rda"))){
```
maurolepore, line 40: Nit: Maybe

```r
audit_path <- file.path(proc_input_path, portfolio_name_ref_all, "audit_file.rda")
if (file.exists(audit_path)){
  audit_file <- readRDS(audit_path)
```

maurolepore (approved):
Great to see that some needless code and processes are now gone.
I haven't run this locally but I see the CI worklow passing -- so LGTM.

Clare2D (approved):
lgtm

Issue #353, opened 2020-12-09
Write_log: add explicit named argument to last case where this is missing


jacobvjk:
explicit specification of file path missing here:
https://github.com/2DegreesInvesting/PACTA_analysis/blob/926d584eafd2d07c32017cf580dede5195775773/0_web_functions.R#L254
default is working fine, but this should be made more robust

fyi @Clare2D

PR #354, opened 2020-12-09
make function arg explicit

jacobvjk:
closes https://github.com/2DegreesInvesting/PACTA_analysis/issues/353

This PR:
- adds the named argument to the function call write_log() in one instance where the explicit argument was missing
- does not touch the functionality of the code, as the default value is the same as the argument that is now being explicitly passed
- makes the code a bit more robust to future changes


jacobvjk:
ugh, I forgot I need to push to upstream in this repo in order for the gh actions to work.. will close and re-open new PR

PR #355, opened 2020-12-09
make function arg explicit

jacobvjk:
closes https://github.com/2DegreesInvesting/PACTA_analysis/issues/353

This PR:

- adds the named argument to the function call write_log() in one instance where the explicit argument was missing
- does not touch the functionality of the code, as the default value is the same as the argument that is now being explicitly passed
- makes the code a bit more robust to future changes


jacobvjk:
unsure why the R CMD check passes when the test fails. In either case there is a bug in stress testing, which still cals source("utils.R"). This is wring since moving utils into the R directory and exporting it's functions. instead, these functions should be covered by load_all. will adapt https://github.com/2DegreesInvesting/StressTestingModelDev/issues/146

cjyetman (approved):
seems ok to me

PR #356, opened 2020-12-09
Add sector order to create_interactive_report input

MonikaFu:
Relates to https://github.com/2DegreesInvesting/create_interactive_report/pull/313

For this PR to work properly above PR from `create_interactive_report` is needed.

In this PR I modify `web_tool_script_3.R` to read `sector_order` data from `create_interactive_report` directory and add it as an input to the `create_interactive_report` function call.

---
depends on 2DegreesInvesting/create_interactive_report/pull/313


maurolepore:
@MonikaFu, I made this a draft because I see it depends on https://github.com/2DegreesInvesting/create_interactive_report/pull/313 -- so I interpret that this PR is not ready to merge.

web_tool_script_3.R
```diff
@@ -163,6 +163,11 @@ js_translations <- jsonlite::fromJSON(
   txt = path(template_path, "data/translation/js_labels.json")
 )
 
+sector_order <- readr::read_csv(
+  path(template_path, "data/sector_order/sector_order.csv"),
```
maurolepore, line 167: Best to use `path("to", "somewhere")` instead of `path("to/somewhere")`. Different platforms use different path separators -- avoid "/".

maurolepore: Our tests on GitHub actions show [this fails](https://github.com/2DegreesInvesting/PACTA_analysis/runs/1523994553?check_suite_focus=true). As you disclose, you can indeed wait for https://github.com/2DegreesInvesting/create_interactive_report/pull/313 or you may rewrite this in such a way that you keep the ecosystem working at all times. For example, you may now add something like 

```r
if (fs::file_exists("...") {
  # read file
} else {
  # do nothing
}
```

maurolepore: You will have to come back after https://github.com/2DegreesInvesting/create_interactive_report/pull/313 and remove this little logic, but at lease you move in the direction you want, and always in the green test-zone.

web_tool_script_3.R
```diff
@@ -218,7 +223,8 @@ create_interactive_report(
   ipr_results_stress_test = ipr_results_stress_test,
   display_currency = display_currency,
   currency_exchange_value = currency_exchange_value,
-  header_dictionary = header_dictionary
+  header_dictionary = header_dictionary,
+ sector_order = sector_order
```
maurolepore, line 227: My suggestion above adds logic that you can't inject in a variable like `sector_order()`, but you can inject it in a function `sector_order()`.

maurolepore (commented):
Thanks! See my comments.

MonikaFu:
@maurolepore - indeed. Both PRs are needed for the code to work.

web_tool_script_3.R
maurolepore: Let me try something out. An example may be clearer than my words above.

web_tool_script_3.R
MonikaFu: > Best to use `path("to", "somewhere")` instead of `path("to/somewhere")`. Different platforms use different path separators -- avoid "/".

I'm not very familiar with the usage of the `path()` function but it sounds reasonable - will modify according to your suggestion.

MonikaFu: @maurolepore see 5e13306

maurolepore (approved):
I made minor changes to break the dependency with create_interactive_report#313. All workflows on GitHub actions now pass, and you can merge this PR before create_interactive_report#313.

It now LGTM. Feel free to edit or merge.

web_tool_script_3.R
```diff
@@ -163,6 +163,22 @@ js_translations <- jsonlite::fromJSON(
...
+      Have you already merged create_interactive_report#313?"
+    ))
+    return(NULL)
+  }
+
+  readr::read_csv(
+    path(template_path, "data", "sector_order", "sector_order.csv"),
+    col_types = cols()
+  )
+}
```
maurolepore, line 180: This will now work even if the data is unavailable. This is an intermediate refactoring step to break the dependency on create_interactive_report#313.

web_tool_script_3.R
```diff
@@ -221,6 +237,12 @@ create_interactive_report(
   header_dictionary = header_dictionary
 )
 
+has_arg <- utils::hasName(formals(create_executive_summary), "sector_order")
+if (has_arg) {
+  arg <- list(sector_order = sector_order())
+  arguments <- append(arguments, arg)
+}
+purrr::invoke(create_interactive_report, arguments)
```
maurolepore, line 245: This will now work even if `create_interactive_report()` has no formal argument `sector_order`. 

(This is an intermediate refactoring step to break the dependency on create_interactive_report#313.)

maurolepore:
I see one test fails (R-release on Windows) do to problems with latex. That seems unrelated to this PR so I would approve it anyway. 

```r
Error: Error: LaTeX failed to compile D:/a/PACTA_analysis/PACTA_analysis/PACTA_analysis/working_dir/50_Outputs/TestPortfolio_Input/executive_summary/template.tex. See https://yihui.org/tinytex/r/#debugging for debugging tips. See template.log for more info.
```

Also, the same environment that fails `on: push` passes `on: pull_request`, so I suspect the problem may go away by itself -- I'll re-run the workflows and see what happens.

image

MonikaFu:
Thanks @maurolepore for making this PR deployable without the dependency on create_interactive_report#313. I will try to follow your steps in the future!

maurolepore:
The scafolding I added here is removed via https://github.com/2DegreesInvesting/PACTA_analysis/pull/360

PR #357, opened 2020-12-10
Changes the peergroup for Liechtenstein

Clare2D:
Changes the peer_group for Liechtenstein. Unfortunately this has to happen here and can't be done by Constructiva.


cjyetman (approved):
I knew this was potentially coming, and this looks like the easiest most appropriate solution. 👍🏻 

Is it intended for there to be extra spaces in the new group names though?

Clare2D:
Yep. This is the format of the peers files so simpler just to keep them both with spaces.

PR #358, opened 2020-12-11
Update gh-actions workflows

maurolepore:
This PR updates gh-aciotns workflows.

* To run R CMD check: The goal is to keep it close to 
  latest changes in r-lib/actions/examples.

* To test the web tool: The goal is to generalize it so
  we can reuse an identical workflow in all .pacta siblings.
  This is part of the long-term goal of automating tests in 
  .pacta siblings so we can refactor them safely and quickly.

Both workflows have in common that they need to adapt to
a new release of usethis, which requires a system dependency.
I first worked each workflow independently but I found myself
duplicating commits. So I bring here the work for both workflows
with the single goal of  making out CI pipeline more up to date.


maurolepore:
I finally got this to a good state. I'll close this messy PR and open a cleaner one.

PR #362, opened 2020-12-12
Update CI

maurolepore:
This PR removes a dependency on usethis and
makes our CI workflows more efficient, robust
and reusable. 

Remove dependency on usethis:
This PR removes a thin dependency on usethis. This
is motivated by a new release of usethis, wihch
need a system dependency that caused gh-actions
workflows to break. Trying to solve that problem
was taking to long and it seems a good oportunity
to get one dependency down. 

Update CI workflows:
On the CI front, I took the change to update the
R CMD check workflow with that where it comes from, 
r-lib/actions/examples, which adds suppor for the
"main" branch, and uses less runners (triggs only
on push to master or main, and on any pull request
to those branches).


.github/workflows/R-CMD-check.yaml
```diff
@@ -7,9 +7,11 @@
 on:
   push:
     branches:
+      - main
       - master
   pull_request:
     branches:
+      - main
       - master
```
maurolepore, line 15: * We are now specific about what branches trigger gh-actions. This saves runners and time when,  
  for example, we push a branch but don't yet care about running CI. If we do want to trigger CI we
  can now more intentionally make it a PR -- maybe as a draft.

* GitHub is pushing to rename the default branch from master to main. As we transition, we need
  to watch both.

.github/workflows/R-CMD-check.yaml
```diff
@@ -40,12 +42,12 @@ jobs:
...
       - uses: actions/checkout@v2
 
-      - uses: r-lib/actions/setup-r@master
+      - uses: r-lib/actions/setup-r@v1
         with:
           r-version: ${{ matrix.config.r }}
           http-user-agent: ${{ matrix.config.http-user-agent }}
 
-      - uses: r-lib/actions/setup-pandoc@master
+      - uses: r-lib/actions/setup-pandoc@v1
```
maurolepore, line 50: The workflows we use come from r-lib/actions/examples. It's nice to keep up with 
changes there.

.github/workflows/test-r-scripts.yaml
```diff
@@ -1,4 +1,12 @@
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+      - master
+  pull_request:
+    branches:
+      - main
+      - master
```
maurolepore, line 9: As for the workflow to run R CMD check, also here we are more intentional about what branches trigger gh-actions.

.github/workflows/test-r-scripts.yaml
```diff
@@ -23,9 +31,14 @@ jobs:
...
     steps:
+      # When checking out the repository that triggered a workflow, `ref:`
+      # defaults to the reference or SHA for that event. Otherwise, uses the
+      # default branch --https://github.com/actions/checkout
       - name: Checkout PACTA_analysis
         uses: actions/checkout@v2
         with:
+          repository: 2DegreesInvesting/PACTA_analysis
+          token: ${{ secrets.MAURO_PAT_FOR_2DII }}
           path: PACTA_analysis
```
maurolepore, line 42: This should make this workflow reusable in other pacta repos.

.github/workflows/test-r-scripts.yaml
```diff
@@ -63,13 +76,27 @@ jobs:
           r-version: ${{ matrix.config.r }}
           http-user-agent: ${{ matrix.config.http-user-agent }}
 
+      - name: Install system dependency of usethis >= 2.0.0
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt install libgit2-dev
```
maurolepore, line 82: From usethis 2.0.0 there is a new system dependency.  usethis may not a dependency directly, but it is indirectly via devtools. So we need to install the system dependency or we'll fail to run the web tool scripts. 

Relates to https://github.com/2DegreesInvesting/docker/issues/28

.github/workflows/test-r-scripts.yaml
```diff
@@ -63,13 +76,27 @@ jobs:
...
+          deps <- unique(renv::dependencies("../PACTA_analysis")$Package)
+          exclude <- c(
+            "R",
+            "utils",
+            "base",
+            "tools",
+            "create_interactive_report",
+            "PACTA.analysis",
+            "StressTestingModelDev"
+          )
```
maurolepore, line 98: We are now installing only what we need, and avoiding avoidable installation problems.

.renvignore
```diff
@@ -1 +1,2 @@
 inst/extdata/r_packages.R
+data-raw
```
maurolepore, line 2: Ask renv to ignore packages we use in data-raw/ -- as it is a non-standard directory that faces only developers. Developers are expected to have usethis as part of their devleopment environment -- but we do not need to force users to do so.

DESCRIPTION
```diff
@@ -39,7 +39,6 @@ Imports:
     rlang,
     rmarkdown,
     tibble,
-    usethis,
```
maurolepore, line 42: PACTA.analysis -- the package -- no longer depends on usethis.

NAMESPACE
```diff
@@ -19,4 +19,3 @@ importFrom(rlang,"%||%")
 importFrom(rlang,abort)
 importFrom(rlang,inform)
 importFrom(rlang,warn)
-importFrom(usethis,ui_field)
```
maurolepore, line 22: This is the only convenience function we used from usethis. Easy to remove.

R/get_param.R
```diff
@@ -18,7 +18,7 @@ START.YEAR <- get_param(
 
 pluck_param <- function(x, file, if_null) {
   out <- purrr::pluck(config::get(file = file), !!!x)
-  param <- ui_field(dplyr::last(unlist(x)))
```
maurolepore, line 21: In addition to removing `ui_field`, there is a little refactoring here. Sorry for the confusion. Let me explain:

The only meaningful change is removing `usethis::ui_field()`. Then I'm removing some intermediate variables that seem complicated -- the code without them seems clearer.

tests/testthat/test-get_param.R
```diff
@@ -0,0 +1,48 @@
+test_that("can use configuration file via options", {
```
maurolepore, line 1: I realized we had no tests for`get_param()` so I borrowed some from r2dii.utils and adapted them a bit.

tests/testthat/test-get_param.R
```diff
@@ -0,0 +1,48 @@
...
+  expect_equal(get_param("a_field")(), "a_value")
+})
+
+test_that("outputs the expected value", {
+  config_file <- glue::glue("
+    default:
+      a_field: a_value
+  ")
+  config_path <- tempfile()
+  readr::write_lines(config_file, config_path)
```
maurolepore, line 19: This may be refactored to remove duplication. I'll leave it as is for now, to maybe refactor is in a peer programming session with @MonikaFu

PR #363, opened 2020-12-12
Remove dependency on usethis

maurolepore:


maurolepore:
This is fixed by #364 

```
── web_tool_script_1.R ─────────────────────────────────────────────────────────
Error in loadNamespace(j <- i[[1L]], c(lib.loc, .libPaths()), versionCheck = vI[[j]]) : 
  there is no package called ‘usethis’
Calls: loadNamespace ... loadNamespace -> withRestarts -> withOneRestart -> doWithOneRestart
Execution halted
Error: Error: This script threw an error.
Execution halted
Error: Process completed with exit code 1.
```

If you prefer to keep CI green at all times, instead of reviewing this PR then #364 you can instead reopen and review https://github.com/2DegreesInvesting/PACTA_analysis/pull/362 -- maybe exploring file changes commit by commit to replicate the experience you would get in two separate PRs.

R/get_param.R
```diff
@@ -28,11 +28,9 @@ pluck_param <- function(x, file, if_null) {
   }
 
   if (is.null(out) && !is.null(if_null)) {
-    param_ <- ui_field(param)
     warn(glue("On config.yml, {param} is undefined."))
 
-    if_null_ <- ui_field(if_null)
-    inform(glue("Setting {param_} to default value: {if_null_}."))
+    inform(glue("Setting {param} to default value: {if_null}."))
```
maurolepore, line 33: Sorry for mixing the removal of `ui_field()` with the refactoring of variables (here removing the intermediate variable `if_null`).

tests/testthat/test-get_param.R
```diff
@@ -0,0 +1,48 @@
+test_that("can use configuration file via options", {
```
cjyetman, line 1: I don't really know what any of this stuff us either, so I guess it's fine

cjyetman (approved):

R/get_param.R
maurolepore: The `ui_*` helpers are tagged as internal, but the helpfile should pop up with `?usethis::ui`. You like colour messages, so you may like these helpers

image

tests/testthat/test-get_param.R
maurolepore: `get_param()` gets values from specific fields in a configuration file. It allows you to specify the behavior your want if the field is NULL (e.g. return NULL, throw an error, etc.) and to define the configuration file explicitely or implicitely via `options()`. The interface is awkward because it is a function factory -- it made sense to me in the context of r2dii.utils, but here I feel like I should make it simpler and likely to move it to a thinner version of r2dii.utils.

PR #365, opened 2020-12-14
Rename workflow to source web-tool scripts

maurolepore:
PACTA_analysis has more than the web-tool. This PR
renames files and functions so that it's clearer 
that what they do is specific to a part and not all
of PACTA_analysis.


maurolepore:
The name is now clearer

image

cjyetman (approved):
❤️

PR #366, opened 2020-12-15
Don't checkout user_results

maurolepore:
A conversation with CJ suggests I can remove user_results from the CI workflow where we source the web-tool scripts.


cjyetman:
I think it needs to be replaced by something more appropriate... but in the current state of things, the web tool scripts would probably fit without it.

cjyetman:
I guess this is ok considering the test passes. I thought that the current test parameters are for a test user in an initiative that has survey results, so I feared that this would break things, but I guess it looks alright.

cjyetman (approved):

maurolepore:
user_results/ is unused in the current CI workflow where we source the web-tool scripts.

PR #370, opened 2021-01-07
Small changes input check

Clare2D:
Most of this is white space formatting, however I also remove reference to the "fin_data" from a function where it is not used at all. this is in Line 1145. 

Not very important just to be slightly more precise.


cjyetman (approved):
👍🏻

PR #371, opened 2021-01-07
Should be CHF not EUR

Clare2D:
Closes https://github.com/2DegreesInvesting/create_interactive_report/issues/350
@catarinabrg 

They like CHF not EURs - who knew?!


MonikaFu (approved):
LGTM

PR #373, opened 2021-01-08
Add fund coverage summary

andybue:
Added a functions that
 
- analyze fund coverage 
get_fund_coverage
- summarize a fund coverage table that might be used as a basis for investor information on the fund coverage
summarize_fund_coverage
- list all unknown funds in funds
list_unknown_funds_in_funds 

Added calling these functions during web_tool_script_1

and added new exports in the export section


Clare2D (commented):
I think there is an additional function in there. Reflecting now, I'm not really a fan of the duplication - but we can improve on that later.

andybue:
Closed this and will add a new, cleaned PR with the same title in few minutes

PR #383, opened 2021-01-28
add lines in web_tool_script_1 L83 to ensure both isin columns contai…

### LINKED ISSUE #382: ensure that fund data isins are characters not factors
[Recent improvements to the fund data](https://github.com/2DegreesInvesting/pacta-data/pull/24) made the isin columns factors instead of characters to save on memory. Code here needs to ensure that it converts these factors to characters when/if needed, or needs to handle factors appropriately.

andybue:
close #382


cjyetman (approved):
lgtm

PR #386, opened 2021-02-09
adapt to changes in create_portfolio_subset

cjyetman:
`create_portfolio_subset()` signature was changed in #370
web_tool_script_1.R was adapted simultaneously, but this "offline" script was not


Clare2D (approved):
Thanks

PR #387, opened 2021-02-09
WAITING FOR #388. Add rmd workflow

Clare2D:
Depends on #388. Changes from here moved to there so that this PR stays focused
on only documenting (not editing) existing code.

----

This is an initial outline of an rmd document that notes some assumptions included in the code but also shows the steps for running a project offline. It copies primarily from the "offline" scripts but includes some components of the online scripts. 

For discussion.

### DIFF
--- a/workflow.Rmd
+++ b/workflow.Rmd
@@ -1,0 +1,473 @@
+---
+title: "PACTA_analysis workflow"
+author: "Clare"
+date: "08/02/2021"
+output: html_document
+---
+
+## Aim
+
+The aim of this document is to outline the workflow for running a PACTA project. 
+
+## PACTA_analysis Repo
+
+The PACTA_analysis repo contains the primary PACTA functions required to run a portfolio alignment assessment for investors. The current repository is built in a way that the functions work for both an online application - ie for the P2020 projects and platform; as well as the original application which is running a project or generating results offline.
+
+Whether online or offline, running an analysis has three components
+
+1. Portfolio Cleaning: Clean the input portfolio and merge in financial data, to categorise each holding and identify whether it's equity (EQ) or corporate bonds (CB)
+2. Run Analysis: Merge in ald + scenarios, and group results at company, portfolio and regional level
+3. Results: Present the results in a clear output format
+
+Both the online and offline scripts can be used to generate reports offline. The difference is that the offline set of scripts is set up to allow the generation of results for multiple portfolios and investors in one run, whereas online is focussed on quickly generating a report for one portfolio at once. While all steps maintain the grouping by investor and portfolio, to allow for this, it hasn't been optimised for running multiple portflios as once. The online tool however has been designed for the interactive reports, whereas the offline scripts have primarily been used to generate the static pdf results.  
+
+
+### Portfolio Cleaning
+
+This occurs in web_tool_script_1.R (online) or 2_project_input_analysis.R (offline). 
+
+Both start with reading in the necessary files. This includes some file cleaning, that should be shifting to the data_preprataion repository.  
+
+## Difference between online and offline
+
+There are two sets of scripts that call the functions required for both offline or online functionality. 
+
+Online: The "web_tool_script" set from 1 to 3 runs through the components of the analysis
+Offline: The 1,2,3,4 scripts run similarly, however have an initialisation script that should be run before each online in order to set up project pathways and naming. This allows for added flexibility as you may online want to be focussing on one of the components of the analysis. Ie you've already generated results and just want to create output files. 
+
+
+# Running a project
+
+
+## Offline 
+This steps through the functions required to run a project using the original scripts. The aim of this script is to define project paths and set parameters used in the following scripts. 
+
+### Set up parameters and paths
+```{r 1_portfolio_check_initialisation, echo = FALSE}
+
+devtools::load_all()
+use_r_packages()
+
+## Project Initialisation
+
+working_location <- paste0(getwd(), "/")
+
+source("0_portfolio_test.R")
+source("0_graphing_functions.R")
+source("0_reporting_functions.R")
+source("0_portfolio_input_check_functions.R")
+source("0_global_functions.R")
+source("0_sda_approach.R")
+
+project_name <- "TestProject"
+twodii_internal <- FALSE
+# TRUE or FALSE: TRUE means that the code is running on a 2dii laptop with dropbox connection
+
+#####################################################################
+### ONLY FOR EXTERNAL PROJECTS (twodii_internal <- FALSE):
+# Variables must exist for internal projects
+project_location_ext <- "C:/Users/clare/Desktop/ExternalTest"
+data_location_ext <- "C:/Users/clare/Desktop/PACTA/pacta-data/2019Q4"
+#####################################################################
+
+create_project_folder(project_name, twodii_internal, project_location_ext)
+
+set_project_paths(project_name, twodii_internal, project_location_ext)
+
+copy_files(project_name)
+
+options(r2dii_config = fs::path(par_file_path, "AnalysisParameters.yml"))
+
+
+# parameters
+project_code <- "GENERAL"
+
+set_global_parameters(fs::path(par_file_path, "AnalysisParameters.yml"))
+
+set_project_parameters(file.path(working_location, "parameter_files",paste0("ProjectParameters_", project_code, ".yml")))
+
+analysis_inputs_path <- set_analysis_inputs_path(twodii_internal, data_location_ext, dataprep_timestamp)
+
+```
+
+TO DO: 
+
+- Improve the sourcing of files. Setting the internal flag to false to allow for defining this is one solution here. 
+
+- Synchronise the parameter files with the online parameter files
+
+
+
+### Portfolio input analysis
+
+Aims:
+
+- Reads in necessary data files and cleans them 
+
+To Run: 
+
+- You need a portfolio to be placed in the folder ```{r} fs::path(project_location_ext, project_name, "20_Raw_Inputs")```. There are sample files found in this repository in the folder "sample_files"
+
+Note:
+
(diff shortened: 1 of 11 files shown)


workflow.Rmd
```diff
@@ -0,0 +1,473 @@
...
+Offline: The 1,2,3,4 scripts run similarly, however have an initialisation script that should be run before each online in order to set up project pathways and naming. This allows for added flexibility as you may online want to be focussing on one of the components of the analysis. Ie you've already generated results and just want to create output files. 
+
+
+# Running a project
+
+
+## Offline 
+This steps through the functions required to run a project using the original scripts. The aim of this script is to define project paths and set parameters used in the following scripts. 
+
+### Set up parameters and paths
```
maurolepore, line 45: Which exact file is this code coming from? Is it `1_portfolio_check_initialisation.R`? If so, can we run this instead of copy-pasting the code?:

```r
source("1_portfolio_check_initialisation.R")
```

workflow.Rmd
```diff
@@ -0,0 +1,473 @@
...
+```{r 1_portfolio_check_initialisation, echo = FALSE}
```
maurolepore, line 46: Knitting workflow.Rmd throws this error:

```r
processing file: workflow.Rmd

Quitting from lines 45-89 (workflow.Rmd) 
Error in config::get(file = file_path) : 
  Config file AnalysisParameters.yml not found in current working directory or parent directories
Calls: <Anonymous> ... eval -> eval -> set_global_parameters -> <Anonymous>
Execution halted
```

workflow.Rmd
```diff
@@ -0,0 +1,473 @@
...
+
+## Difference between online and offline
+
+There are two sets of scripts that call the functions required for both offline or online functionality. 
+
+Online: The "web_tool_script" set from 1 to 3 runs through the components of the analysis
+Offline: The 1,2,3,4 scripts run similarly, however have an initialisation script that should be run before each online in order to set up project pathways and naming. This allows for added flexibility as you may online want to be focussing on one of the components of the analysis. Ie you've already generated results and just want to create output files. 
+
+
+# Running a project
```
maurolepore, line 39: What are the inputs?

workflow.Rmd
```diff
@@ -0,0 +1,473 @@
...
+
+- Reads in necessary data files and cleans them 
+
+To Run: 
+
+- You need a portfolio to be placed in the folder ```{r} fs::path(project_location_ext, project_name, "20_Raw_Inputs")```. There are sample files found in this repository in the folder "sample_files"
+
+Note:
+
+- Fund file optional. This adds additional holdings to the portfolio (adds in the details of the companies owned within a fund). If the file for the current time stamp is there this should be read in. Often this file is project specific as the file size can be very large.
```
maurolepore, line 113: It would be nice to see a code chunk finding this file and printing a few lines to see what it looks like.

workflow.Rmd
```diff
@@ -0,0 +1,473 @@
...
+
+To Run: 
+
+- You need a portfolio to be placed in the folder ```{r} fs::path(project_location_ext, project_name, "20_Raw_Inputs")```. There are sample files found in this repository in the folder "sample_files"
+
+Note:
+
+- Fund file optional. This adds additional holdings to the portfolio (adds in the details of the companies owned within a fund). If the file for the current time stamp is there this should be read in. Often this file is project specific as the file size can be very large. 
+
+- this is currently reading in the "cleaned" files found in the pacta-data repo. This workflow needs to be cleaned up so that these "cleaned" files can be created for any project.
```
maurolepore, line 115: Here it would be nice to include a code chunk showing that from the working directory we must be able to find the external repository pacta-data, and presumably some specific file in there.

workflow.Rmd
```diff
@@ -0,0 +1,473 @@
...
+There are two sets of scripts that call the functions required for both offline or online functionality. 
+
+Online: The "web_tool_script" set from 1 to 3 runs through the components of the analysis
+Offline: The 1,2,3,4 scripts run similarly, however have an initialisation script that should be run before each online in order to set up project pathways and naming. This allows for added flexibility as you may online want to be focussing on one of the components of the analysis. Ie you've already generated results and just want to create output files. 
+
+
+# Running a project
+
+
+## Offline
```
maurolepore, line 42: At a high level, for the online tool we can do something like this:

```r
PACTA.analysis::source_web_tool_scripts()
```

Or like this:

```r
source("web_tool_script_1.R")
source("web_tool_script_2.R")
source("web_tool_script_3.R")
``` 

Is there some way to also run the "offline" workflow with one or a few calls to `source()` or other function?

workflow.Rmd
```diff
@@ -0,0 +1,473 @@
...
+# Running a project
+
+
+## Offline 
+This steps through the functions required to run a project using the original scripts. The aim of this script is to define project paths and set parameters used in the following scripts. 
+
+### Set up parameters and paths
+```{r 1_portfolio_check_initialisation, echo = FALSE}
+
+devtools::load_all()
```
maurolepore, line 48: This might be too much detail for now. I would start by documenting the workflows at the highest level, showing clearly what the inputs and outputs are, and making sure this rmarkdown document acutally runs in anyone's computer. Once we have that we could move deeper into details if necessary.

workflow.Rmd
```diff
@@ -0,0 +1,473 @@
...
+
+```
+
+There are options as to where to proceed from here. From this point there are options:
+/item Run the stress testing code
+/item Print static graphics and the pdf report (older code)
+/item Print the interactive report (html) and executive summary (pdf) 
+
+This is where there is a different in workflow between the online and offline case.
+The offline workflow directs a user to the pdf code that we've been using recently. There should be little stopping a user from printing a html report, however the script to do this may need to be created.
```
maurolepore, line 445: Maybe a flow chart could help.

workflow.Rmd
```diff
@@ -0,0 +1,473 @@
...
+
+This is where there is a different in workflow between the online and offline case.
+The offline workflow directs a user to the pdf code that we've been using recently. There should be little stopping a user from printing a html report, however the script to do this may need to be created. 
+
+
+#### SDA Approach
+
+The SDA approach is a methodology that is used for "Other" Sectors that do not have clear technology roadmaps, such as Steel, Cement, and Aviation but do have emission factor models. 
+
+This is just a test really to understand how this code works as we may want to incorporate and not lose this work that was done by Vincent.
```
maurolepore, line 452: I think Jackson developed Vincent's idea. Is it the same?

* https://2degreesinvesting.github.io/r2dii.analysis/articles/target-sda.html
* https://2degreesinvesting.github.io/r2dii.analysis/reference/target_sda.html

maurolepore (commented):
Thanks a lot Clare! I wrote some comments. 

I suggest to first focus on documenting the inputs, process, and outputs at a very high level. For example, we can run the so called "online" workflow with the right inputs and a single call to `PACTA.analysis::source_web_tool_scripts()`. 
It would be great for this rmarkdown document to include code to show the inputs of that function, then a call to `PACTA.analysis::source_web_tool_scripts()`, and some more code showing the outputs. That would prove that the "offline" workflow is reproducible, and anyone could track the source code. 

Do we have something similarly "automatic" for the so called "offline" workflow?

workflow.Rmd
```diff
@@ -0,0 +1,473 @@
...
+
+Whether online or offline, running an analysis has three components
+
+1. Portfolio Cleaning: Clean the input portfolio and merge in financial data, to categorise each holding and identify whether it's equity (EQ) or corporate bonds (CB)
+2. Run Analysis: Merge in ald + scenarios, and group results at company, portfolio and regional level
+3. Results: Present the results in a clear output format
+
+Both the online and offline scripts can be used to generate reports offline. The difference is that the offline set of scripts is set up to allow the generation of results for multiple portfolios and investors in one run, whereas online is focussed on quickly generating a report for one portfolio at once. While all steps maintain the grouping by investor and portfolio, to allow for this, it hasn't been optimised for running multiple portflios as once. The online tool however has been designed for the interactive reports, whereas the offline scripts have primarily been used to generate the static pdf results.
```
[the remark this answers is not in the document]
Clare2D: Let's discuss this!

workflow.Rmd
Clare2D: This is something that we can change to create this functionality. at the moment there are one or two parameters that do not allow this.

workflow.Rmd
```diff
@@ -0,0 +1,473 @@
...
+This occurs in web_tool_script_1.R (online) or 2_project_input_analysis.R (offline). 
+
+Both start with reading in the necessary files. This includes some file cleaning, that should be shifting to the data_preprataion repository.  
+
+## Difference between online and offline
+
+There are two sets of scripts that call the functions required for both offline or online functionality. 
+
+Online: The "web_tool_script" set from 1 to 3 runs through the components of the analysis
+Offline: The 1,2,3,4 scripts run similarly, however have an initialisation script that should be run before each online in order to set up project pathways and naming. This allows for added flexibility as you may online want to be focussing on one of the components of the analysis. Ie you've already generated results and just want to create output files.
```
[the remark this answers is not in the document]
maurolepore: More precisely, I do understand the sentence that starts with "Online: ", mostly because I worked on this workflow. Yet it would be useful to show code that complements this sentence. Running something like this should do:

```r
source("web_tool_script_1.R")
source("web_tool_script_2.R")
source("web_tool_script_3.R")
```

Regarding the paragraph starting with "Offline: ", I think my confusion is mostly caused by wording -- particularly the references to "online" in this "offline" section.

workflow.Rmd
maurolepore: Maybe we can write a little wrapper that sets the required parameters so we can source this script? I imagine something like this:

```r
initialize_portfolio <- function(parameters) {
  # We need to figure out this bit. But maybe this function
  # simply overwrites some .yml file on the fly.
  set_initialization_parameters(parameters)

  source("1_portfolio_check_initialization.R")
}
```

maurolepore:
Follow up in #389. 
I see this PR has more files than I expected and I can't figure out how to get them out. 
Locally, the diff against master shows only one file changed; yet the diff here shows 8.

PR #396, opened 2021-02-16
Add rmd workflow

Clare2D:
New PR that should allow you to run a new project with 2020Q4 data. There are multiple todos, which will be outlined in the Rmd in a different PR. 
The file cleaning should generally move to a script being written by @cjyetman , but this is working now for this purpose.


maurolepore:
Thanks! I see a merge conflict but seems straightforward to solve. It looks that we want to indeed add the call to `fund_coverage()` that does not exist in the master branch, right? 

```
<<<<<<< add-rmd-workflow
# information of coverage and coverage loses for all funds in raw_portfolio
fund_coverage <- get_fund_coverage(
  portfolio_raw,
  fin_data,
  fund_data,
  currencies,
  grouping_variables
)

=======
>>>>>>> master
```

maurolepore (commented):
LGTM. Once we fix the merge conflict we should be okay to merge.

maurolepore (approved):

Clare2D:
Yeah this is actually an older version that still doesn't work. I've got local changes that I've been trying to commit but with no luck unfortunately.

maurolepore:
Okay, let me know if I can help. I struggled with something similar yesterday.

Clare2D:
thanks. I just replicated the changes on a new version :( It seems to work again though and will create the PR now.

PR #398, opened 2021-02-16
Small changes to get PACTA_analysis working offline 

Clare2D:
Let's try this again. This contains the fixes required to get the PACTA code running offline. rmd to follow. Forming the starting point for changes to come through.


0_portfolio_input_check_functions.R
```diff
@@ -6,8 +6,8 @@ read_raw_portfolio_file <- function(project_name) {
 
   input_path <- paste0(project_location, "/20_Raw_Inputs/")
 
-  csv_to_read <- list.files(path = input_path, pattern = paste0(project_name, "_Input.csv"))
-  txt_to_read <- list.files(path = input_path, pattern = paste0(project_name, "_Input.txt"))
+  csv_to_read <- list.files(path = input_path, pattern = "_Input.csv")
```
maurolepore, line 9: A more robust pattern may be `"_Input[.]csv$"`

2_project_input_analysis.R
```diff
@@ -82,6 +84,8 @@ emissions_totals <- calculate_portfolio_emissions(
 ################
 #### SAVING ####
 ################
+file_format_list <- c("rda")
```
maurolepore, line 87: Maybe `"rda"` instead of `c("rda")`? It does the same but slightly more simply.

3_run_analysis.R
```diff
@@ -1,7 +1,7 @@
 # TODO:
 # Clean up sectors options
 
-port_col_types <- set_col_types(grouping_variables, "ddddccccddcl")
+# port_col_types <- set_col_types(grouping_variables, "ddddccccddcl")
```
maurolepore, line 4: Is "comment in/out" the way to determine whether a run of this file is for the online/offline tool?

maurolepore (approved):
LGTM

0_portfolio_input_check_functions.R
cjyetman: 👍🏻 those regexes can sneak up on you

maurolepore: I'll go ahead and do the change. Here is my modes test

``` r
pattern <- "_Input[.]csv"
filename <- "whatever_Input.csv"
grepl(pattern, filename)
#> [1] TRUE
```

<sup>Created on 2021-02-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

3_run_analysis.R
maurolepore: @Clare2D can you explain a bit why we are commenting a line out instead of removing it? 

(I suspect it's because it helps run a different workflow and if so I would think of a way to parametrize this line.)

0_portfolio_input_check_functions.R
```diff
@@ -864,16 +864,18 @@ create_id_columns <- function(portfolio, portfolio_type) {
...
 
 get_and_clean_currency_data <- function() {
-  currencies <- read_rda("data/currencies.rda")
-  
+  currencies <- readRDS("data/currencies.rds")
+
   currencies <- set_currency_timestamp(currencies)
-  
+
+  if (all(currencies$exchange_rate) == 1){print("currency data temporary. todo: update")}
```
cjyetman, line 877: I don't understand `all(currencies$exchange_rate)`.... presumably `currencies$exchange_rate` is a vector of values, so `all(currencies$exchange_rate)` evaluates to `TRUE` if all of the values in it are >0, and `TRUE` is automatically coerced to `1` when it's compared with ` == 1`?

maurolepore:
@Clare2D, I did only minor changes and pulled out workflow.Rmd and .md (I'll follow up in a separate PR). The checks now pass (following another PR plush some black magic) so I'm ready to merge. Now checking with @cjyetman to ensure I don't mess up his docker images.

0_portfolio_input_check_functions.R
```diff
@@ -1529,7 +1531,7 @@ get_average_emission_data <- function(inc_emission_factors) {
   average_sector_intensity <- data.frame()
 
   if (inc_emission_factors) {
-    average_sector_intensity <- read_rda(paste0(analysis_inputs_path, "average_sector_intensity.rda"))
+    average_sector_intensity <- read_rda(paste0(analysis_inputs_path, "/average_sector_intensity.rda"))
```
cjyetman, line 1534: might as well make this `read_rda(file.path(analysis_inputs_path, "average_sector_intensity.rda"))` while we're changing it anyway

0_portfolio_input_check_functions.R
```diff
@@ -1538,7 +1540,7 @@ get_company_emission_data <- function(inc_emission_factors) {
   company_emissions <- data.frame()
 
   if (inc_emission_factors) {
-    company_emissions <- read_rda(paste0(analysis_inputs_path, "company_emissions.rda"))
+    company_emissions <- read_rda(paste0(analysis_inputs_path, "/company_emissions.rda"))
```
cjyetman, line 1543: same here... use `file.path`

cjyetman (changes requested):
`data/currencies.rda` appears in at least the following three locations in master, so I would think it's not too safe to remove it in this PR...
https://github.com/2DegreesInvesting/PACTA_analysis/blob/e6ffb2c92e30d1373107a5e2d27751419dbcabf2/data-raw/data.R#L5
https://github.com/2DegreesInvesting/PACTA_analysis/blob/e6ffb2c92e30d1373107a5e2d27751419dbcabf2/.Rbuildignore#L37
https://github.com/2DegreesInvesting/PACTA_analysis/blob/c353c7fdd02cf9c7fa2972b98521b4f3f9208740/0_portfolio_input_check_functions.R#L833

3_run_analysis.R
Clare2D: I commented it out because it caused it to fail, and the test isn't very indicative of the issue. Could be rewritten as something more relevant

0_portfolio_input_check_functions.R
Clare2D: Hmm... basically I created a fake column in the df for 2020Q4 and filled it with 1s, so wanted to make sure that this eventually gets replaced with the actual values. Any good ideas for this?

Clare2D: So I struggled with this, and couldn't create an alternate file that included an extra column that could be opened by both calls to it (as discussed in Slack). We can either keep both files in here until this is resolved, or something else that I'm not sure about

maurolepore: 7d0f382 restores data/currencies.rda. Having it and not using it seems safer than removing it.

maurolepore: > I created a fake column in the df for 2020Q4 and filled it with 1s ...

I thinks it's more natural to use `NA` because it's R's representation of missing values. I would be more specific and fill it with the specific type of `NA` that fits the data you expect to eventually get in there:


``` r
data <- tibble::tibble(x = 1:3)
data$y <- NA_integer_
data
#> # A tibble: 3 x 2
#>       x     y
#>   <int> <int>
#> 1     1    NA
#> 2     2    NA
#> 3     3    NA

if (all(is.na(data$y))) {
  warning("TODO: update.")
}
#> Warning: TODO: update.
```

<sup>Created on 2021-02-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

cjyetman (approved):
lgtm

Issue #393, opened 2021-02-16
web tool scripts are failing on latex compilation


cjyetman:
@maurolepore all but the macOS instance are failing the web tool scripts tests with..
```
Error: Error: LaTeX failed to compile
D:/a/PACTA_analysis/PACTA_analysis/PACTA_analysis/working_dir/50_Outputs/TestPortfolio_Input/executive_summary/template.tex. See https://yihui.org/tinytex/r/#debugging for debugging tips.
```
https://github.com/2DegreesInvesting/PACTA_analysis/pull/391/checks?check_run_id=1908941675

I don't see any reason why this error would be caused by any recent changes in our code, so I assume it must be something with how the instance is being created.

maurolepore:
I think we are good now. I'll close before I forget as I did not use the right keyword on the related PR.

PR #399, opened 2021-02-16
Prune and update gh-actions workflows

maurolepore:
Relates to #393

This PR reduces two gh-action workflows, from being similar to
`usethis::use_github_action_check_full()` to being similar to
`usethis::use_github_action_standard()`. 

It not only reduced the number of checks (because it runs
in fewer platforms and versions of R) but also it updates
the workflows to more closely follow the workflows at 
r-lib/actions.

Notice that R 4.0.4 was released yesterday and gh-actions
may not reach this "too new" version of R. It is possible
that otherwise fine workflows now fail with R-release. If
the problem is indeed that R 4.0.4 is unavailable, then
we may just wait a few days in case the problem goes away.

The workflow that sources the web tool scripts is the 
skinniest, as the web tool runs only on one 
[platform](https://github.com/2DegreesInvesting/docker/blob/99e2844a31bcf97f6560245f32dff2b55cc3c746/system/Dockerfile#L1) and [version of R](https://github.com/2DegreesInvesting/docker/blob/99e2844a31bcf97f6560245f32dff2b55cc3c746/system/Dockerfile#L7).

* [Example run of source web tool scripts](https://github.com/2DegreesInvesting/PACTA_analysis/actions/runs/573053713)
* [Example run of R CMD check](https://github.com/2DegreesInvesting/PACTA_analysis/actions/runs/573053712)


maurolepore:
The error I see on CI seems unrelated with our code. Look how similar it is with the error I get on my local ubuntu 20.04 running R 4.0.3 (not yet 4.0.4):

CI

```r
Error: Error: unable to load shared object '/home/runner/work/_temp/Library/stringi/libs/stringi.so':
  libicui18n.so.55: cannot open shared object file: No such file or directory
```

Local

```
> library(stringr)
Error: package or namespace load failed for ‘stringr’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/home/mauro/R/x86_64-pc-linux-gnu-library/4.0/stringi/libs/stringi.so':
  libicui18n.so.60: cannot open shared object file: No such file or directory
```

.github/workflows/source-web-tool-scripts.yaml
```diff
@@ -11,20 +11,17 @@ on:
 name: Source web-tool scripts
 
 jobs:
-  Test-r-scripts:
+  R-CMD-check:
```
cjyetman, line 14: why is this changed to `R-CMD-CHECK` here in this action?

maurolepore: Thanks for noticing. I was sloppy when I copy-pasted from the R CMD check workflow. Fixed now.

cjyetman (approved):
cool, thanks! just for my information... do you have any idea what was crucial to getting the web tool scripts to run on the linux test?

maurolepore:
No, haha. I'm not sure if anything that I did mattered. It may well be that coincidentally someone made the latest release of R (4.0.4) available for gh-actions on ubuntu.

PR #402, opened 2021-02-17
Document Clare's workflow in workflow.Rmd(.md)

maurolepore:
Following #398 this pr aims to document the workflows that this repository is capable of. 

The goal is for a naive analyst to be able to generate outputs one click of a button or a call to a function. for the "web tool" this is already possible with `source_web_tool_scripts()` but for other workflows we don't have such a thing.


maurolepore (commented):
I'll go ahead and merge this PR. It touches no source code so it can't brake anything. And merging it will mark an interesting point in the git history of this repo -- that documents what Clare gets locally. On top of that I will submit a separate PR that aims to generalize what Clare got.

PR #403, opened 2021-02-18
Additional small changes

Clare2D:
A few additional changes that I thought of yesterday to improve a few things. 
- changing the parameter file name (dropping the _GENERAL) this is important so we don't end up changing the "ProjectParameters_GENERAL" that will be used online. 
- changed the other_sector_list parameter to refer to the name it has in the parameter file already

### DIFF
--- a/0_portfolio_input_check_functions.R
+++ b/0_portfolio_input_check_functions.R
@@ -5,9 +5,9 @@ read_raw_portfolio_file <- function(project_name) {
   portfolio <- NA
 
   input_path <- paste0(project_location, "/20_Raw_Inputs/")
-  # pattern matches e.g.: "whatever_Input.csv"
-  csv_to_read <- list.files(path = input_path, pattern = "_Input[.]csv$")
-  txt_to_read <- list.files(path = input_path, pattern = "_Input[.]txt$")
+
+  csv_to_read <- list.files(path = input_path, pattern = "_Input.csv")
+  txt_to_read <- list.files(path = input_path, pattern = "_Input.txt")
 
 
   if (length(csv_to_read) == 1) {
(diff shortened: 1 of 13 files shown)


maurolepore (commented):
@Clare2D, I removed all that deals with style and `paste0()` to make this PR easier to understand and review. I'll submit those changes in a follow up PR.

@cjyetman can you please see if this PR raises any concern?

cjyetman (approved):
I believe almost everything here is related only to the "offline" version of running PACTA, and does not affect the web tool scripts, so it seems pretty safe from that perspective. I don't know or understand the reason or context for most of these changes, so I can't really judge that much.

The one thing that does concern me a bit it is the change to `0_portfolio_input_check_functions.R` on L421 that changes `other_sector_list` to `pacta_sectors_not_analysed`... primarily because that could be something that affects the web tool scripts via `get_and_clean_fin_data()`, and I don't know anything about the purpose of that change, or if or where those objects do or do not exist.

0_portfolio_input_check_functions.R
```diff
@@ -418,7 +418,7 @@ check_fin_mapped_sectors <- function(fin_data) {
     ))
 
   actual_sectors <- unique(fin_data$security_mapped_sector)
-  if (any(!actual_sectors %in% c(sector_list, other_sector_list, "Other"))) {
+  if (any(!actual_sectors %in% c(sector_list, pacta_sectors_not_analysed, "Other"))) {
```
maurolepore, line 421: > [This] could ... affect the web tool scripts via `get_and_clean_fin_data()` -- CJ

Clare can you please explain a bit what is the purpose of this change and if/where these objects exist?

cjyetman: looking into this myself a bit... `pacta_sectors_not_analysed` appears to be something that gets set in the parameters yml file, and read in by `set_project_parameters()` in `0_global_functions.R`

https://github.com/2DegreesInvesting/PACTA_analysis/blob/af15e6c2b1d83cc1724d831139b7a7fec3101f8b/0_global_functions.R#L67

and it appears in the default parameter files for each of the projects, so I give a 👍🏻 
https://github.com/2DegreesInvesting/PACTA_analysis/search?q=pacta_sectors_not_analysed

cjyetman: additionally, `pacta_sectors_not_analysed` seems to have been introduced here (for context)
https://github.com/2DegreesInvesting/PACTA_analysis/pull/350

PR #404, opened 2021-02-18
Project Parameters File for Austria created

fiona511:
I created the file ProjectParameters_PA2020AT.yml for the Austrian project.

Changes in this parameter file compared to Liechtenstein include: 

        project_report_name: austria
        display_currency: EUR
        currency_exchange_value: 1.12

as well as the stress testing parameters

stress_test:
        shock_year: 2030
        price_data_version: 2021Q1

### DIFF
--- a/parameter_files/ProjectParameters_PA2020AT.yml
+++ b/parameter_files/ProjectParameters_PA2020AT.yml
@@ -1,0 +1,90 @@
+default:
+
+    reporting:
+        project_report_name: austria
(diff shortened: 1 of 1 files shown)


cjyetman (approved):
web tool scripts test is failing I believe because @fiona511's user doesn't have the appropriate permissions to trigger the test

contextually I don't really know if this is all correct, but technically speaking it seems safe and appropriate, so I'm going to merge

PR #405, opened 2021-02-23
change shipping scenario to one that exists

Clare2D:
The shipping scenario is incorrect, hence it is dropped from charts. This should resolve that for all projects.


cjyetman (approved):
makes sense to me

PR #406, opened 2021-02-23
add if statement for exec summary

Clare2D:
this ensure that there is no error if the directory with the required template doesn't exist


cjyetman (approved):
🙏🏻

PR #407, opened 2021-02-23
Add choose sectors

Clare2D:
In order to be able to remove shipping from the report, I've added filters so that you can select which sectors come through in the data. However this hasn't worked for all charts. I guess either we either have to change the chart code - or remove the shipping from the financial data as well. This is required for the general initiative as we are not allowed to use the shipping tech share. 
Perhaps there is an option in which we do not display the technologies in this case (we are allowed to show exposure). Or @MonikaFu is there a simple way for you to ensure Shipping remains on the right, even if it doesn't come through in the ALD? 
image

### DIFF
--- a/0_portfolio_test.R
+++ b/0_portfolio_test.R
@@ -23,7 +23,8 @@ get_ald_scen <- function(portfolio_type) {
   ald <- ald %>%
     filter(
       scenario_source %in% scenario_sources_list,
-      scenario_geography %in% scenario_geographies_list
+      scenario_geography %in% scenario_geographies_list,
+      ald_sector %in% sector_list
     ) %>%
     mutate(mapped_ald = 1)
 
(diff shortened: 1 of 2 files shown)


MonikaFu:
I'd say that from the chart perspective - shipping data should not be passed to the chart if it should not be displayed? I'd rather avoid filtering data in the plotting functions.

Clare2D:
Ok I worked out where that was happening and fixed it in a different PR in create-interactive-report

cjyetman:
so this is ready to go now that 2DegreesInvesting/create_interactive_report/pull/379 landed?

Clare2D:
Yep, it is!

cjyetman (approved):
again for my own understanding... `sector_list` in the "offline" code gets set to the `pacta_sectors_not_analysed` in the parameters file? and in the web tool scripts, `pacta_sectors_not_analysed` gets stored in `sector_list`, which gets passed on to `create_interactive_report()` as `twodi_sectors`, which then filters out Shipping for the general users?

if I understand that correctly, LGTM

PR #411, opened 2021-02-25
Add new fund processing offline

Clare2D:
Added Andy's functionality into the offline tool as well as the online. Primarily to ensure revision of the fund files. This doesn't touch anything live.


cjyetman (approved):
on the surface, this all makes perfect sense to me.... haven't tested myself though, just fyi

PR #415, opened 2021-03-03
Delete test_test_cases.R

### LINKED ISSUE #414: remove test_test_cases.R
I wrote `test_test_cases.R` last fall in the rush to get things working for the Swiss project. I think we should probably remove it because it has limited utility and may have fallen out of sync with the rest of the code anyway. Does anyone object to that?

@Clare2D @jacobvjk @maurolepore @MonikaFu @jdhoffa

cjyetman:
closes #414


Clare2D (approved):
Incredible work on this one!! ;)

PR #419, opened 2021-03-07
use portfolio_name_ref_all as fixed not regex, and some code cleanup

### LINKED ISSUE #418: Get input files fails with () in investor or portfolio name
https://github.com/2DegreesInvesting/PACTA_analysis/blob/42e53f8204dcf065cc9778f7671bc1de6ccf9c99/0_web_functions.R#L108

The line 117 in the 0_web_functions fails if there is a parenthesis in the investor name column. I assume this may also be the case for other special characters. Either we need to ensure that these don't get this far, or alter the function here.

cjyetman:
closes #418 

use `fixed = TRUE` to make sure that the string contained in `portfolio_name_ref_all` is matched as a literal string rather than interpreted as a regex command in `grep()`

@Clare2D can you test this in/with the situation that revealed the problem?

### DIFF
--- a/0_web_functions.R
+++ b/0_web_functions.R
@@ -114,12 +114,12 @@ get_input_files <- function(portfolio_name_ref_all) {
   # set_portfolio_parameters(file_path = file.path(par_file_path, "PortfolioParameters.yml"))
   # check that the provided reference names match to the input files
 
-  input_files <- list.files(path = input_path, full.names = F)
-  input_names <- gsub(".csv", "", input_files)
-  input_names <- gsub(".txt", "", input_names)
-  input_names <- gsub(".xlsx", "", input_names)
+  input_files <- list.files(path = input_path, full.names = FALSE)
+  input_names <- tools:::file_path_sans_ext(input_files)
+  portfolio_files <- grep(portfolio_name_ref_all, input_files,
+                          value = TRUE, fixed = TRUE)
+  input_file_type <- unique(tools::file_ext(portfolio_files))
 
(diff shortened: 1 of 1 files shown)


0_web_functions.R
```diff
@@ -114,12 +114,12 @@ get_input_files <- function(portfolio_name_ref_all) {
...
 
-  input_files <- list.files(path = input_path, full.names = F)
-  input_names <- gsub(".csv", "", input_files)
-  input_names <- gsub(".txt", "", input_names)
-  input_names <- gsub(".xlsx", "", input_names)
+  input_files <- list.files(path = input_path, full.names = FALSE)
+  input_names <- tools:::file_path_sans_ext(input_files)
+  portfolio_files <- grep(portfolio_name_ref_all, input_files,
+                          value = TRUE, fixed = TRUE)
+  input_file_type <- unique(tools::file_ext(portfolio_files))
```
Clare2D, line 121: I guess this will cause an error if a user uploads two different types of files right? I'm not sure if it's possible and what we should do about this though..

cjyetman: this PR should have the exact same behavior as before except for specifically treating `portfolio_name_ref_all` as a fixed string rather than a regex... other changes here were cosmetic/stylistic

but... it will likely give a warning (though not an error) on L123 if there is more than one file type because only the first element in what could be a multi-element result will be evaluated in the `if` statement.

however, as far as I can tell, that whole chunk testing the file extensions is irrelevant since L166 forces the filename to end in `.csv` anyway... so apparently this function is only able to read in CSV files, otherwise it will fail because it forces the file extension to be `.csv`

Clare2D:
Thanks for the clarification

Clare2D (approved):
LGTM

PR #420, opened 2021-03-08
dockerize

maurolepore:


.Rbuildignore
```diff
@@ -51,3 +51,8 @@
 ^test_cases$
 ^test_test_cases\.R$
 
+^Dockerfile$
+^docker-compose\.yml$
+^workflow-multiple-inputs_cache$
+^workflow\.Rmd$
+^workflow\.md$
```
maurolepore, line 58: This is unrelated but causes no harm and makes this PR compatible.

AlexAxthelm:
Let's talk later this week about docker, because I want to make sure that this is done in a controlled manner, with a plan. I'll close for now, but we can reopen once we have a plan.

maurolepore:
This or similar may continue in regards to AB#964

PR #424, opened 2021-03-09
Reuse image 2dii/pacta_analysis on CI

maurolepore:


cjyetman:
I think we need to take down https://hub.docker.com/r/2dii/pacta_analysis as soon as possible. My understanding is that it has private data in it. @Clare2D @AlexAxthelm

.github/workflows/source-web-tool-scripts.yaml
```diff
@@ -37,59 +21,24 @@ jobs:
...
-          install.packages("knitr")
-          install.packages("renv")
-          deps <- unique(renv::dependencies("PACTA_analysis")$Package)
-          exclude <- c(
-            "R",
-            "utils",
-            "base",
-            "tools",
-            "create_interactive_report",
-            "PACTA.analysis",
```
Clare2D, line 87: PACTA_analysis?

Clare2D:
The other repositories are private, so I don't think it's a problem no?

cjyetman:
> The other repositories are private, so I don't think it's a problem no?

The 2dii/pacta_analysis Docker image contains clones of our repositories that are private: StressTestingModelDev, create_interactive_report, and pacta-data

.github/workflows/source-web-tool-scripts.yaml
cjyetman: CRAN does not allow package names to contain `_`, so although the repo is named `PACTA_analysis`, the pkg infrastructure, e.g. the DESCRIPTION file, refer to it as `PACTA.analysis`

cjyetman:
I had @AlexAxthelm remove 2dii/pacta_analysis Docker Hub image out of caution.

To be fair, I might be wrong, because I'm not exactly sure how that docker image is/was created, but my assumption is that it uses our docker building scripts tailored for PACTA analysis, which do include pacta-data and other private repos inside of it.

@maurolepore let's talk about how that docker image is/was created when you get a chance

cjyetman:
ok, sorry.... I misunderstood the intention here and panicked. I had never heard of this `2dii/pacta_analysis` Docker image, and I assumed that it was the same as our `2dii/pacta` Docker image which includes clones of all the PACTA family of repos.

As I understand it now, @maurolepore is working on creating a new Docker image that will be included in this repo rather then in the `2DegreesInvesting/docker` repo (which I support, in the long run anyway). The `2dii/pacta_analysis` is actually much more similar (or the same) to our `2dii/r-packages` Docker image, which does not contain any private data. The docker file that creates this new docker is currently in a PR branch here https://github.com/2DegreesInvesting/PACTA_analysis/blob/dockerize/Dockerfile

cjyetman:
alright, I would suggest that this PR does **not** close #269 because the goal of that issue is to run the GH Action tests on the same environment as it runs on the transitionmonitor website, which this is not

additionally, I would suggest that this PR **not** modify the `.github/workflows/source-web-tool-scripts.yaml` unless/until transitionmonitor is actually using the same Docker image as here.... at the moment, `2dii/r-packages` would be a better choice since that is truly the base for the current docker image that runs on transition monitor

cjyetman:
also, I would say that this PR at least contextually depends on #420

.github/workflows/dev-source-web-tool-scripts.yaml
```diff
@@ -0,0 +1,48 @@
+name: Dev source web-tool scripts
```
maurolepore, line 1: @cjyetman, FYI this workflow is now in a different file. The one we've used so far remains intact.

cjyetman: thanks. don't kill me, but... I'd prefer two separate PRs, one with the addition of a new workflow file, and one with all of the changes to the existing workflow file. I think that makes it easier to separate them contextually, and would be easier to understand how and why changes were made.

cjyetman: actually, forget that... I see that is the case now... must've had an old diff cached in my browser or something and it looked like both were included here

maurolepore:
Thanks CJ for the input! I'll close until Alex helps us define a plan.

FYI I tried using 2dii/r-packages (instead of 2dii/pacta_analysis) but the latex configuration [failed](https://github.com/2DegreesInvesting/PACTA_analysis/pull/424/checks?check_run_id=2077749151) -- even when I copy-pasted the setup I see in 2dii/pacta. But I won't try to debug it as my experiments suggest we can avoid the problem if our base images was rocker/verse.

I'm very happy that Alex is now involved; it's a great opportunity to learn Docker.

BTW,

cjyetman:
> FYI I tried using 2dii/r-packages (instead of 2dii/pacta_analysis) but the latex configuration [failed](https://github.com/2DegreesInvesting/PACTA_analysis/pull/424/checks?check_run_id=2077749151) -- even when I copy-pasted the setup I see in 2dii/pacta. But I won't try to debug it as my experiments suggest we can avoid the problem if our base images was rocker/verse.

just revisiting this.... I think the latex compilation errors here are due to tinytex targeting xelatex which is not installed, and I believe that can be avoided by running `tinytex::install_tinytex()` and/or `options(tinytex.tlmgr.path = '/root/.TinyTeX/bin/x86_64-linux/tlmgr')`

maurolepore:
This is followed up at #444

PR #426, opened 2021-03-12
export audit CSV

### LINKED ISSUE #425: Return Audit File as csv
It seems that the audit file in csv format is not being printed for the web tool, however this is required! It is just being saved as rda and this should instead be a csv. 
This isn't critical, however there is a broken link that is returned when the user clicks to download this.

cjyetman:
closes #425 

haven't actually tested this, and I don't know the context of why this is necessary, but.... this should export a CSV of `audit_file` at the same time in the same location as it already exports a RDS of the `audit_file`


Clare2D (approved):
Lgtm

PR #428, opened 2021-03-15
also export invalidsecurities as CSV

### LINKED ISSUE #427: Create additional csv file results
There is a JSON file written entitled invalidsecurities. We however need this same information in csv as well. Providing the user with the JSON file is not so useful! 

additionally, I think we should make the portfolio level results available. This is equity_results_portfolio and bonds_results_portfolio also as csv files.

cjyetman:
closes #427 

I'm not thrilled about this PR. I don't fully understand what's being exported here or why. I don't know why it was originally exported as JSON and now it needs to be CSV. I don't know if we should be doing the same thing with the other files that are being exported here also. I don't know if we should keep exporting JSON files and also export CSV files, or if we should pick one. That all being said, this resolves #427 with minimal changes so at least we'll be able to track how/when/why this was done.

### DIFF
--- a/0_json_functions.R
+++ b/0_json_functions.R
@@ -263,17 +263,13 @@ export_audit_information_jsons <- function(audit_file_ = audit_file,
     portfolio_total_ <- subset(portfolio_total_, investor_name != "Meta Investor")
   }
 
-
-
-
-
   folder_path <- paste0(folder_path, "/")
   if (!is.na(project_name_)) {
     folder_path <- paste0(folder_path, project_name_, "_")
   }
   # function
   export_audit_graph_json(audit_file_, paste0(folder_path, "coveragegraph"))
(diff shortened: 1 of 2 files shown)


Clare2D (approved):
lgtm

Clare2D:
As I mentioned, we need both. One is used to create the graphs that are displayed automatically, and the csv is to make a more user friendly downloadable file. Thanks!

cjyetman:
web tool script test is failing, so putting into draft mode until I figure out why

cjyetman:
forgot to modify one of the function names... guess that's what tests are for 😉 

let's see if it passes this time around

cjyetman:
ok, passing tests now 😊

Clare2D:
👍

PR #430, opened 2021-03-15
Add empty fund_isin column

Clare2D:
The code has error the column fund_isin not found if there is no fund data. This solved this. It doesn't affect the online tool as there will be always fund_data available there. And allows this to run successfully offline without fund data


cjyetman (approved):
haven't tested, but conceptually this makes sense to me

PR #432, opened 2021-03-23
change file to path in write_csv

Clare2D:
write_csv requires path not file! This solves the issue of the file not printing


cjyetman (approved):
ah, easy.... and dumb mistake in the first place 😊

cjyetman:
actually.... documentation says `path` argument is deprecated, and `file` should be correct? https://readr.tidyverse.org/reference/write_delim.html

maurolepore:
Maybe Clare's version of readr was before `path` was deprecated?

PR #434, opened 2021-03-24
add conditional statement

Clare2D:
This ensures that english report is always set for general. To be changed in the future in the case that more languages are made available.


cjyetman (approved):
Lgtm

Issue #436, opened 2021-03-24
get rid of warnings


cjyetman:
`web_tool_script_1.R` currently gives the following warnings, which could probably be easily eliminated...

```r
There were 13 warnings (use warnings() to see them)
Warning message:
In if (!is.na(total_fund_list)) { :
  the condition has length > 1 and only the first element will be used
`summarise()` ungrouping output (override with `.groups` argument)
`summarise()` ungrouping output (override with `.groups` argument)
`summarise()` ungrouping output (override with `.groups` argument)
Joining, by = "holding_id"
Joining, by = "holding_id"
Joining, by = "holding_id"
```

maurolepore:
Details about the first warning shown above:

``` r
# `total_fund_list` comes from two sources, both dataframes with a single column 
# `fund_isin` which length is > 1L
total_fund_list <- readRDS("data/total_fund_list.rds")
str(total_fund_list)
#> 'data.frame':    251577 obs. of  1 variable:
#>  $ fund_isin: Factor w/ 251577 levels "","AD000A1JX363",..: 1 46218 46222 46221 46220 39327 39333 39330 249112 46216 ...

# Not showing the output of this one
# str(fst::read_fst(file.path(file_location, "total_fund_list.fst")))



# (From now on I'll be using fake data with same structure but smaller)
total_fund_list <- data.frame(fund_isin = factor(1:2))
str(total_fund_list)
#> 'data.frame':    2 obs. of  1 variable:
#>  $ fund_isin: Factor w/ 2 levels "1","2": 1 2



# Therefore:

# 1: It seems imprecise to query the entire data.frame. Likely we should
# instead query specifically the column `fund_isin`:

# Good
!is.na(total_fund_list$fund_isin)
#> [1] TRUE TRUE

# Bad
!is.na(total_fund_list)
#>      fund_isin
#> [1,]      TRUE
#> [2,]      TRUE



# 2: The expression returns an object of length > 1. When passing it to `if()`
# likely it should first be wrapped in `all()` or `any()`. The current behaviour
# uses only the first element with a warning

# Good
if (!is.na(total_fund_list$fund_isin)[[1]]) "..."
#> [1] "..."

# Better
if (all(!is.na(total_fund_list$fund_isin))) "..."
#> [1] "..."

# Bad
if (!is.na(total_fund_list$fund_isin)) "..."
#> Warning in if (!is.na(total_fund_list$fund_isin)) "...": the condition has
[...]

PR #438, opened 2021-03-26
add in number of shares check

Clare2D:
Error occurring due to number of shares being passed through as character not numeric.

I updated this to include some more specifications that remove some recurring warning messages, such as dropping the groups and specifying the merge variable.

### DIFF
--- a/0_portfolio_input_check_functions.R
+++ b/0_portfolio_input_check_functions.R
@@ -1180,31 +1180,31 @@ get_fund_coverage <- function(portfolio_raw,
 
   fund_portfolio_total_mapped_value_usd <- fund_portfolio_total  %>%
     group_by(holding_id) %>%
-    summarize(total_mapped_value_usd = sum(value_usd))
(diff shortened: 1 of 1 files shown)


Clare2D:
Not sure what this error message in the checks is about. Any ideas CJ?

cjyetman:
So this will resolve all but the first warning referenced in #436 👍🏻

I don't know what this error in the "Source web-tool scripts" GH Action means exactly... will have to investigate further
```r
Error: Error: Assigned data `as.numeric(portfolio$number_of_shares)` must be compatible with existing data.
✖ Existing data has 2966 rows.
✖ Assigned data has 0 rows.
ℹ Only vectors of size 1 are recycled.
```

cjyetman:
The error in the web tool script occurs because the `portfolio` data.frame does not have a column `number_of_shares` (it has `market_value`). So this code I guess has to be adaptable to only do these checks and coercions if `number_of_shares` exists.

cjyetman (approved):
seems like that did the trick

PR #439, opened 2021-04-02
Avoid warning: In write_csv() arg path is deprecated

### LINKED ISSUE #435: file or path argument for write_csv?
actually.... documentation says `path` argument is deprecated, and `file` should be correct? https://readr.tidyverse.org/reference/write_delim.html

_Originally posted by @cjyetman in https://github.com/2DegreesInvesting/PACTA_analysis/issues/432#issuecomment-804778161_

maurolepore:
Closes #435 

* Avoids this warning:

```r
#> Warning: The `path` argument of `write_csv()` is deprecated as of readr 1.4.0.
#> Please use the `file` argument instead.
```

* Flags with `FIXME` other places with the same problem but not related to #435. Should be followed up in a separate PR.


cjyetman (commented):
Clare originally made this change in #432 and believed that was necessary to get the proper CSV file exported. Can we verify at least the the proper CSV file is being exported before we switch back to this?

maurolepore:
Sure. This experiment (and common sense) suggests they the name of the argument makes no difference in the output (assuming both names are supported and implemented correctly). I suspect Clare was using a computing environment where readr was older -- before `path` was deprecated. 

``` r
packageVersion("readr")
#> [1] '1.4.0'

# I got this object triggering a debugger environment with `browser()`. As it's small, I reproduce it here.
portfolio_total_ <- tibble::tribble(
              ~isin, ~marketValues, ~currency,                     ~flag,
  "WeightGapMarker",           477,     "EUR", "Invalid or missing ISIN",
    "MISSINGWEIGHT",           477,     "EUR", "Invalid or missing ISIN",
                 NA,           477,     "EUR", "Invalid or missing ISIN",
                 NA,           477,     "EUR", "Invalid or missing ISIN",
                 NA,           477,     "EUR", "Invalid or missing ISIN",
                 NA,           477,     "EUR", "Invalid or missing ISIN",
                 NA,           477,     "EUR", "Invalid or missing ISIN",
                 NA,           477,     "EUR", "Invalid or missing ISIN",
                 NA,           477,     "EUR", "Invalid or missing ISIN",
                 NA,           477,     "EUR", "Invalid or missing ISIN"
)

tmp1 <- tempfile(fileext = ".csv")
readr::write_csv(portfolio_total_, file = tmp1)

tmp2 <- tempfile(fileext = ".csv")
readr::write_csv(portfolio_total_, path = tmp2)
#> Warning: The `path` argument of `write_csv()` is deprecated as of readr 1.4.0.
#> Please use the `file` argument instead.

identical(readLines(tmp1), readLines(tmp2))
#> [1] TRUE
```

<sup>Created on 2021-04-02 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

--

I would like to take this as an opportunity to up vote for a shared computing environment (e.g. a docker image with RStudio server installed like rocker/verse plus the packages we need). That would allow us to test out code before we submit issues/PRs.

cjyetman:
I get that theoretically this should be fine, but it should be relatively easy to verify that the file is actually exported as expected. 

The GitHub action should test every PR on a managed system much like Docker, so that should already be happening automatically.

cjyetman:
It may be necessary/prudent to check which version of `readr` is included in the Docker image we create for the web host. At the end of the day, that's the environment that matters the most.

cjyetman:
And for context... I think Clare was confused somehow in thinking this change fixed exporting this file. But that triggered two new Docker images and about 1.5 weeks of back and forth with the web host. I'd rather avoid going through all of that again, and since this is easy to test actual output, we should do that (in as close to live on the web host environment as possible). Our current CI does not test for this output.

cjyetman:
This was a very recent breaking change in readr 1.4.0 2020-10-05. So I think it's very likely that Clare's machine and very possibly the docker image that runs on the web host both have an older version of readr that will Error if we use file instead of path. Again, this is why I think it's imperative to actually test this rather than rely on theoretical justification.

maurolepore:
VERSION

Thanks for explaining the context and risks. A fresh image `2dii_pacta` suggests in production we are using the latest version of readr -- the one that supports both `file` (new)` and `path` (deprecated). Also I see we are using the slightly older (minor) version of R.

```r  
> packageVersion("readr")
[1] '1.4.0'

> R.version
               _                           
platform       x86_64-pc-linux-gnu         
arch           x86_64                      
os             linux-gnu                   
system         x86_64, linux-gnu           
status                                     
major          4                           
minor          0.3                         
year           2020                        
month          10                          
day            10                          
svn rev        79318                       
language       R                           
version.string R version 4.0.3 (2020-10-10)
nickname       Bunny-Wunnies Freak Out   
```


I think it makes sense. The installed version should be the  latest at the time the image `2dii_pacta` was built, right?

It  would  be nice to have  an interactive terminal in the production environment  to see  exactly what we are running  but I don't know if that's  possible and practically doable.


TESTING 


 Can you help with this? You seem to know what the resulting .csv should look like. I don't. I haven't even  looked at the file. I could try research where this all comes from and try figure out what to expect but you seem to already have that context so we could joinly save a lot of time. Feel free to add commits or to close this PR and create a new one.

cjyetman:
Yeah, I will test it when I'm back "in the office". I honestly don't know much more than than what I've already said here. I'm not sure exactly what the file is supposed to look like. I think that the file is potentially offered to users as a download so that they can see what ISINs were not matched in their portfolio, but I'm not exactly sure how/when/where that happens. Clare said it wasn't working and made the original change... I'm not sure if that was based on feedback from the web host or from testing it locally on her machine. 🤷🏻 That's why I'm a bit hesitant/paranoid about it.

maurolepore:
Given this change already caused some trouble, Should we stay safe and implement both `file` and `path`?

```r
readr_new_interface <- function() {
  packageVersion("tidyr") > "x.y.z"
}

if (readr_new_interface()) {
  readr::write_csv(file = "path/to/file")
} else {
  readr::write_csv(path = "path/to/file")
}
```
Inspiration comes from here: https://tidyr.tidyverse.org/articles/in-packages.html#conditional-code

0_json_functions.R
```diff
@@ -348,7 +348,7 @@ export_audit_invalid_data <- function(portfolio_total_, export_path_full) {
   invalidsecurties <- toJSON(portfolio_total_, dataframe = c("columns"))
 
   write(invalidsecurties, file = paste0(export_path_full, ".json"))
-  readr::write_csv(portfolio_total_, path =  paste0(export_path_full, ".csv"))
+  readr::write_csv(portfolio_total_, paste0(export_path_full, ".csv"))
```
maurolepore, line 351: Despite the change in name, the argument `path/file` has always remained in the second position. This solution is backward compatible, consistent with other uses of `write_csv()` in this repo, and light-weight.

cjyetman: hmmm.... I don't love it, but it is an ingenious solution to working in both circumstances 👍🏻

maurolepore: Yeah, I know. If you prefer I can implement `readr_new_interface()` as shown above. Let me know what you'd be happy to approve so we get this out of the way.

cjyetman: Ugh, I don't know.... I guess checking the version is the most proper solution and avoids someone "fixing" this later and breaking it again without leaving annoying comments in the code everywhere this happens @AlexAxthelm what's the least amount of "tech debt" we can create from this situation?

AlexAxthelm: Hmm. I'm a big fan of named arguments, but this does have some elegance to it. 🤔 

If we were only committed to supporting the "web version", I'd say make the change and live with it. But if our goal is to support a wide range of platforms, I think putting the conditional code in is best. And that opens the door for us to start including new features that are dependent upon certain package versions. With that we could start offering messages like:
> Installed version of `readr` is 1.4.0
> Exporting pictures of turtles is not supported with this version of `readr`. To enable this, please upgrade `readr` to at least version 1.6.3
> `<install.packages(blahblahblah)>`

AlexAxthelm: So I guess my opinion is: put the conditionals in, but don't feel compelled to put in too much work to support old versions

cjyetman: I'd prefer to support more than just the web version because in my long term ideal, there would only be one version, and that will need to work in multiple contexts. Maybe if/when this gets package-ized one day, we can enforce stricter dependency rules/versions globally.

So @maurolepore, sorry for all the run around, but let's go for what you suggested with `readr_new_interface()`.

AlexAxthelm: rather than having a `readr_new_interface()` function, could we have the conditional check the version number in place? Checking the version isn't that expensive (computationally), and it avoids having to define yet another function down the line, if the API changes again (`readr_even_newer_interface()`??)

```r
# file argument to write_csv was introduced in readr x.y.z
if (packageVersion("readr") >= "x.y.z") {
  readr::write_csv(file = "path/to/file")
} else {
  readr::write_csv(path = "path/to/file")
}
```

AlexAxthelm: And of course, we'll need to make sure that we have a way to test the old versions too.

cjyetman: @AlexAxthelm that's what I would probably naturally lean toward, but I'm almost certain @maurolepore will prefer making a mini function for it since it will need to be repeated more than once, and I can at least understand the value in that. Maybe we should be making a `save_write_csv()` function that handles all of it? like...
``` r
save_write_csv <- function(data, file, ...) {
  if (packageVersion("readr") >= "1.4.0") {
    readr::write_csv(data, file = file, ...)
  } else {
    readr::write_csv(data, path = file, ...)
  }
}

save_write_csv(data, file = "path/to/file.csv")
```

AlexAxthelm: Oh, I thought this all already inside a function. That seems like a good path to me. We should also start testing against other package versions, to figure out what the minimum version of each of our dependencies already is. But that would require _some_ level of automated testing.

workflow-pacta-cop.Rmd
```diff
@@ -223,8 +223,7 @@ set_project_parameters(file.path(par_file_path, "ProjectParameters.yml"))
 analysis_inputs_path <- set_analysis_inputs_path(twodii_internal, data_location_ext, dataprep_timestamp)
 
 inc_meta_portfolio <- FALSE
-write_csv(portfolio_peers, file = file.path(raw_input_path, paste0(project_name, "_Input.csv")))
-
+write_csv(portfolio_peers, file.path(raw_input_path, paste0(project_name, "_Input.csv")))
```
maurolepore, line 226: Here I prefer to match by position to avoid exposing the internal packages PACTA.analysis.

0_json_functions.R
```diff
@@ -348,7 +350,7 @@ export_audit_invalid_data <- function(portfolio_total_, export_path_full) {
   invalidsecurties <- toJSON(portfolio_total_, dataframe = c("columns"))
 
   write(invalidsecurties, file = paste0(export_path_full, ".json"))
-  readr::write_csv(portfolio_total_, path =  paste0(export_path_full, ".csv"))
+  safe_write_csv(portfolio_total_, file = paste0(export_path_full, ".csv"))
```
maurolepore, line 353: Thanks @cjyetman. How about renaming to `write_csv()` to intentionally overwrite `readr::write_csv()`? That will spread the protection we just built to all calls to `write_csv()`. The only additional change we would need is to remove the namespace from this one call:

image

maurolepore: Also, would you like me to add a workflow to check the web-tool with readr < 1.4.0?

.Rbuildignore
```diff
@@ -51,3 +51,9 @@
 ^test_cases$
 ^test_test_cases\.R$
 
+^computing-environment_cache$
```
cjyetman, line 54: what is this?

cjyetman (approved):

0_json_functions.R
AlexAxthelm: 👏 Don't 👏 intentionally 👏 introduce 👏 namespacing 👏 conflicts 👏

But more seriously, I'd much rather us go through with a regex and replace the function name, and any namespacing, than introduce an identically named function. 

I also think that `safe_write_csv` is a name that promises more than we are planning to put into it right now, but I can't think of a better name off the top of my head. Maybe it'll come to me soon.

@maurolepore Adding checks for readr < 1.4.0 is a good idea, but let's not jump too hard on that yet, since we need to figure out a way to be testing other old packages as well.

.Rbuildignore
maurolepore: Good catch. It's a leftover form another PR that didn't fly. I removed it.

0_json_functions.R
maurolepore: Okay, I keep a different name. I note that while overwriting well established functions is *generally* a bad idea, it is a powerful way to leverage the power of an interface to inject behavior with o code. Now, for example, if we want to spread the new bahaviour we need to change code in all these places:

image

--

Yeah, I agree that "safe" is vague. How about `write_csv_file()`? It reads well and the `_file` bit hints about the `file` argument. 

--

We have a CI pipeline that runs the web tool on gh actions. It's doesn't exactly use the production environment (we could) but it's close. My idea would be to copy the workflow or extend the existing one to run the web tool after a call to `remotes::install_version("readr", "1.3.1")`.

0_portfolio_input_check_functions.R
```diff
@@ -647,7 +647,7 @@ check_funds_wo_bbg <- function(fund_data, fin_data) {
     bind_rows(fund_isins_missing_bbg) %>%
     distinct()
 
-  readr::write_csv(fund_isins_missing_bbg, "data/fund_isins_without_bbg_data.csv")
+  write_csv_file(fund_isins_missing_bbg, file = "data/fund_isins_without_bbg_data.csv")
```
maurolepore, line 650: Wherever possible, we are now using the new `write_csv_file()` and the explicit argumet `file`, which works regardless of the version of readr.

.github/workflows/source-web-tool-scripts.yaml
```diff
@@ -90,9 +90,19 @@ jobs:
...
+      - name: Source web-tool scripts with versioned packages
+        run: |
+          # See #435
+          devtools::install_version("readr", "1.3.1")
+          stopifnot(utils::packageVersion("readr") < "1.4.0")
+
+          setwd("PACTA_analysis")
+          source("R/source_web_tool_scripts.R")
+          source_web_tool_scripts()
+        shell: Rscript {0}
```
maurolepore, line 108: @AlexAxthelm, here is the bit where I extend the existing workflow to test the web tool not only with the latest packages but also with a version of readr that relates to this PR. If we need to test with other versioned-packages, we add the relevant calls to `remotes::install_version()` here.

Where you thinking of some other strategy? Maybe using a docker image on gh with the specific package-versions already installed?

0_json_functions.R
cjyetman: I'm very hesitant with overwriting function names like that.... then it's very easy for a function to have different behavior than you expect... it might cause confusion with the loading order... etc.

cjyetman: > Also, would you like me to add a workflow to check the web-tool with readr < 1.4.0?

I'm not opposed to that, but honestly I'm still struggling to understand how these workflow.md things are meant to be used, so... if you feel there's value in that, then go for it.

.github/workflows/source-web-tool-scripts.yaml
cjyetman: be aware that this repo is still private and this will probably double the GH Action time we're using

AlexAxthelm: Yeah, I'm inclined to keep this as a "latest-only" check, and we can discuss a strategy for checking older versions later? I'll flag an issue for it on ADO.

0_json_functions.R
maurolepore: Oh, the name is confusing. The workflow I'm now talking about is a gh-actions workflow in .github/workflows/ to test the web tool with R < 1.4.0 on CI. 

The file workflow.md has nothing to do with CI or this PR -- it's meant to document the process and tools that people need to run PACTA.

.github/workflows/source-web-tool-scripts.yaml
maurolepore: > be aware that this repo is still private

Do you mean this repo = PACTA_analysis? It's not private, is it?

maurolepore: Ok, I removed the code that tests the web tool with readr < 1.4.0

cjyetman: > Do you mean this repo = PACTA_analysis? It's not private, is it?

you're right, sorry, my mistake

tests/testthat/test-write_csv_file.R
```diff
@@ -0,0 +1,7 @@
+test_that("with readr < 1.4.0 and named argument `file` throws no error", {
+  skip_if(utils::packageVersion("readr") >= "1.4.0")
+
+  x <- tibble::tibble(x = 1)
+  file <- tempfile()
+  expect_error(write_csv_file(x = x, file = file), NA)
+})
```
cjyetman, line 7: Should we test the converse of this too, that there's no warning if we have `readr` >= 1.4.0? That was the primary intention of all this.

cjyetman (approved):
one nit, otherwise approved

tests/testthat/test-write_csv_file.R
maurolepore: That warning is was a "good" warning, reminding the users of `path` that they should use `file`. That is a behavior of `readr::write_csv()` that I would preserve. Also It wouldn't be super easy to implement, I think, as `write_csv_file()` passes arguments to `readr::write_csv()` via the elipsis `...`, so to intercept `path` we would need to see what arguments the user passed via `...`, then rename `path` as `file`. It seems like a lot of room for error to suppress a behavior that is safe, and that will eventually go away -- maybe years from now -- whenever `readr::write_csv()` decides to forget about `path`.

In the interest of moving forward I'll merge, assuming that my interpretation of the issue is correct, and that my resistance to implement it makes sense. But do feel free to follow up and we can open another PR.

AlexAxthelm: Could we use `expect_silent()` for testing the other side? https://testthat.r-lib.org/reference/expect_silent.html

maurolepore: Oh, I now think you mean to see if the caller of the new `write_csv_file()` effectively throws no warning?

export_audit_invalid_data
image

Maybe I could figure out a way to mock the inputs, but I would be okay just looking at the output in Ci should be fine

https://github.com/2DegreesInvesting/PACTA_analysis/pull/439/checks?check_run_id=2298956928

image

maurolepore: Here I see the relevant warning went away.

0_json_functions.R
cjyetman: thanks for the clarification, I misunderstood

tests/testthat/test-write_csv_file.R
cjyetman: you're right... I'm still wrapping my head around the consequences of this

I think my idea was, we started this thread to prevent these warnings from showing up in the log under normal, working conditions... so that's the primary goal. But yeah, I guess we have for sure achieved that by always using the proper argument name depending on the `readr` version.

cjyetman:
a quick side note to this...

Jackson and I now have a reproducible example where the web server does not allow download of these audit file CSVs, while a similar portfolio in the same account does, so I think we have strong evidence now that the problem is not related to this `readr::write_csv` (thankfully) and instead is related to something weird happening on the server.

PR #440, opened 2021-04-02
Update deprecated argument to write_csv()

maurolepore:
Relates to #435. 

Although the modified file does not show output, it should produce this warning:

```r
#> Warning: The `path` argument of `write_csv()` is deprecated as of readr 1.4.0.
#> Please use the `file` argument instead.
```

This PR avoids such warning by using the argument `file` as the warning says.


cjyetman (approved):
I think this is the right thing to do here.

PR #441, opened 2021-04-02
Avoid warning: Extract first element explicitly

maurolepore:
Relates to #436 

When fed with an object of length > 1L, `if()` extracts the first element with a warning. This PR removes such warning in the most backward-compatible way -- by extracting the fist element explicitly. but it also flags `FIXME` because this is likely incorrect and instead we should likely use get an atomic vector with `all()`.

```r
# Okay (backward compatible but likely wrong)
if (long_logical[[1]]) ...

# Likely better
if (all(long_logical)) ...

# Bad
if (long_logical) ...
```

### DIFF
--- a/0_portfolio_input_check_functions.R
+++ b/0_portfolio_input_check_functions.R
@@ -1085,7 +1085,12 @@ process_raw_portfolio <- function(portfolio_raw,
   original_value_usd <- sum(portfolio$value_usd, na.rm = T)
 
   # correct Funds classification by comparing isin to the list of all known funds isins
-  if(!is.null(total_fund_list)){portfolio <- portfolio %>% mutate(asset_type = ifelse(is.element(isin, total_fund_list$fund_isin), "Funds", asset_type))}
+  if (!is.null(total_fund_list)) {
+    portfolio <- portfolio %>%
+      mutate(asset_type = ifelse(
+        is.element(isin, total_fund_list$fund_isin), "Funds", asset_type
+      ))
+  }
   # identify fund in the portfolio
   fund_portfolio <- identify_fund_portfolio(portfolio)
 
--- a/0_portfolio_input_check_functions.R
+++ b/0_portfolio_input_check_functions.R
@@ -1085,7 +1085,7 @@ process_raw_portfolio <- function(portfolio_raw,
   original_value_usd <- sum(portfolio$value_usd, na.rm = T)
 
   # correct Funds classification by comparing isin to the list of all known funds isins
-  # FIXME: Looking at the signature and noticing this is the fist use of
+  # FIXME: Looking at the signature and noticing this is the only use of
   # `total_fund_list` I interpret that the author intended to make
   # `total_fund_list` an optional argument and chose as a default the value
   # `NA`. Here I express that intent explicitely but note that not `NA` but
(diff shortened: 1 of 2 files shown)


cjyetman (commented):
Thanks. I'd prefer to dig a bit deeper and try to understand what the intent was and do it properly rather than add a band-aid though. Can we do that?

0_web_functions.R
```diff
@@ -318,7 +318,13 @@ save_cleaned_files <- function(save_loc,
...
   fst::write_fst(company_emissions, file.path(save_loc, "company_emissions.fst"))
-  if (!is.na(total_fund_list))
+  # FIXME: Looking at the signature and noticing this is the only use of
+  # `total_fund_list` I interpret that the author intended to make
+  # `total_fund_list` an optional argument and chose as a default the value
+  # `NA`. Here I express that intent explicitely but note that not `NA` but
+  # `NULL` is more commonly used as a default value for optional arguments.
+  total_fund_list_not_passed_as_argument <- identical(total_fund_list, NA)
+  if (!is.na(total_fund_list_not_passed_as_argument))
   {fst::write_fst(total_fund_list, file.path(save_loc, "total_fund_list.fst"))}
```
maurolepore, line 328: Same here, although I don't know if / where this code is used.

cjyetman:
I agree with your assessment in the comment... this was intended to be a non-mandatory argument and the default should be `NULL`... so my preference would be to fix the function signature to do it properly and adapt the calls to it appropriately... I think that's better than jumping through hoops to get `NA` to work without warnings and leaving comments in the code explaining why that will probably stay there forever.

maurolepore:
Great. I did that in two commits. The first one just implementing the changes; the second styling the code for readability. I can split that in a separate PR if you prefer, but I think the style change is small and doubt will make this PR hard to understand.

0_portfolio_input_check_functions.R
```diff
@@ -1085,13 +1085,7 @@ process_raw_portfolio <- function(portfolio_raw,
...
 
   # correct Funds classification by comparing isin to the list of all known funds isins
-  # FIXME: Looking at the signature and noticing this is the only use of
-  # `total_fund_list` I interpret that the author intended to make
-  # `total_fund_list` an optional argument and chose as a default the value
-  # `NA`. Here I express that intent explicitely but note that not `NA` but
-  # `NULL` is more commonly used as a default value for optional arguments.
-  total_fund_list_not_passed_as_argument <- identical(total_fund_list, NA)
-  if(total_fund_list_not_passed_as_argument){portfolio <- portfolio %>% mutate(asset_type = ifelse(is.element(isin, total_fund_list$fund_isin), "Funds", asset_type))}
+  if(!is.null(total_fund_list)){portfolio <- portfolio %>% mutate(asset_type = ifelse(is.element(isin, total_fund_list$fund_isin), "Funds", asset_type))}
```
maurolepore: Please review my use us `!is.null()` -- I tend to get confused with this kind of double negatives.

cjyetman: it looks right to me.... but it looks like the original change with `total_fund_list_not_passed_as_argument` was wrong (i.e. it should **not** do the thing if `total_fund_list` was **not** passed). Now I'm confused... I'll have to look closer at it tomorrow or Wednesday.

maurolepore: I think you are right in that my previous code was wrong. So confusing :-)

cjyetman (approved):

PR #443, opened 2021-04-08
Reuse image 2dii/pacta_analysis on CI

maurolepore:


maurolepore:
This was an experimental PR. I reached a good state and I'll submit a separate cleaner PR.

PR #444, opened 2021-04-08
Reuse image 2dii/r-packages and simplify workflow

maurolepore:
Thanks!. This is awesome if it works. This maybe closes (finally) #269

This PR reuses our image [`2dii/r-packages`](https://hub.docker.com/r/2dii/r-packages). 

Benefits:
* The CI and production environment are more similar.
* GH-actions should run faster because R packages are pre-installed in the image.
* The workflow file to configure gh-actions is much simpler, so easier to maintain.

### DIFF
--- a/.github/workflows/source-web-tool-scripts.yaml
+++ b/.github/workflows/source-web-tool-scripts.yaml
@@ -1,32 +1,15 @@
-on:
-  push:
-    branches:
-      - main
-      - master
-  pull_request:
-    branches:
-      - main
-      - master
-
-name: Source web-tool scripts
-
+name: Source web tool scripts in container r2dii/r-packages
+on: push
 jobs:
-  Test-r-scripts:
-    runs-on: ${{ matrix.config.os }}
-
-    name: ${{ matrix.config.os }} (${{ matrix.config.r }})
-
+  Source-web-tool-scripts:
+    container: 2dii/r-packages:latest
+    runs-on: ubuntu-18.04
+    name: Source web tool scripts in container r2dii/r-packages
     strategy:
       fail-fast: false
-      matrix:
-        config:
-          - {os: ubuntu-18.04,   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest"}
-
     env:
       R_REMOTES_NO_ERRORS_FROM_WARNINGS: true
-      RSPM: ${{ matrix.config.rspm }}
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
-
     steps:
       # When checking out the repository that triggered a workflow, `ref:`
       # defaults to the reference or SHA for that event. Otherwise, uses the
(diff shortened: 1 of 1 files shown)


.github/workflows/source-web-tool-scripts.yaml
```diff
@@ -1,32 +1,15 @@
...
+name: Source web tool scripts in container r2dii/r-packages
+on: push
 jobs:
-  Test-r-scripts:
-    runs-on: ${{ matrix.config.os }}
-
-    name: ${{ matrix.config.os }} (${{ matrix.config.r }})
-
+  Source-web-tool-scripts:
+    container: 2dii/r-packages:latest
```
maurolepore, line 5: This is the most interesting bit.

cjyetman: is this pulling from Docker Hub or from GH Containers? just curious

cjyetman:
Thanks!. This is awesome if it works. This maybe closes (finally) #269?

cjyetman:
so `2dii/r-packages` should already have `tinytex` installed, plus a handful of tlmgr packages that we need, right? Is installing `tinyxtex` again here causing the install to be destroyed and then causing the R script to then trigger reinstalling all those tlgmr packages, which leads to many line in the log?

maurolepore:
> so 2dii/r-packages should already have tinytex installed, plus a handful of tlmgr packages that we need, right? 

Yeah, that's what I think based on 2dii/r-packages/Dockerfile. But I kept getting a latex error and it went away by rerunning this:

```r
          install.packages('tinytex')
          tinytex::install_tinytex()
```

It's kind of dark magic to me.

Also, surprising is that I got this to pass without this:

```bash
# From docker/pacta/Dockerfile
RUN echo "options(tinytex.tlmgr.path = '/root/.TinyTeX/bin/x86_64-linux/tlmgr')" > .Rprofile
RUN chmod -R a+rwX /root
RUN /root/.TinyTeX/bin/*/tlmgr path add
```

.github/workflows/source-web-tool-scripts.yaml
maurolepore: I didn't setup GH Containers, so I think it's pulling from Docker Hub.

cjyetman:
this is one of my test scripts
```sh
#! /bin/bash

# docker load --input 2dii_pacta.tar.gz

portfolio_name="pensionfund"
working_dir="$(pwd)"/working_dirs/pensionfund
user_results="$(pwd)"/user_results/4

docker run --rm -ti \
  --network none \
  --user 1000:1000 \
  --memory="4g" \
  --mount type=bind,source="$working_dir",target=/bound/working_dir \
  --mount type=bind,source="$user_results",target=/user_results \
  2dii_pacta \
  /bound/bin/run-r-scripts "$portfolio_name"
```

maybe we can make the GitHub action login to the instance with `--user 1000:1000` also?

maurolepore:
Good catch @cjyetman! 

While I don't understand fully understand the details, this seems to explain why re-installing tinytex heps -- because the installation is done with a `HOME` and `PATH` that GitHub knows about, right?

In any case, I'll like to start merging PRs. What else do we need?

cjyetman (approved):
I'm ok with this PR as-is. 👍🏻

AlexAxthelm:
I'm missing some context. Is `tinytex` installed relative to `$HOME`?

cjyetman:
I guess `$HOME` is user dependent, so that's hard to answer universally.

PR #446, opened 2021-04-12
445 offline workflow supervisors

### LINKED ISSUE #445: Set up remote offline workflow for supervisors (DNB)
Given a set of slightly adjusted input data, the target of this is to enable supervisors to run PACTA_analysis without sending us data, using the web tool or having one of us run the analysis for them at their premises.

This mainly requires a different financial_data data set, as described [here](https://docs.google.com/document/d/1p-k4M1-ku8y6UIsFrZLhAdUzNoZDcMiLn36RIAxZkoY/edit) and slightly adjusted ald_scenario data for equity. More detailed descriptions to be found in the corresponding data_prep issue https://github.com/2DegreesInvesting/data_preparation/issues/67.

There are also some stricter requirements for the raw portfolio inputs in this workflow.

The offline workflow for supervisors will then largely be the following logic:

- set up the folder structure as would be required under any offline PACTA_analysis project
- run `1_portfolio_check_initialisation.R`, making sure parameters and portfolio names are set as required
- entirely skip `2_project_input_analysis.R`
- instead run a new script detailed below
- run `3_run_analysis.R` (and potentially the fourth script, if this type of output is required)

The new script will:

- read in the raw portfolio
- inner join the financial data
- add a holding id
- split the resulting prepared portfolio by asset type (`Equity` and `Bonds`)
- set the id column for each of the resulting data sets, based on their asset type (`company_id` for equity and `corporate_bond_ticker` for bonds)
- export the prepared portfolio data sets as rds files to the project directory under `./30_Processed_Inputs/projectname_equity_portfolio.rda` and `./30_Processed_Inputs/projectname_bonds_portfolio.rda`

**Note:** this workflow assumes for example that the user has prepared the input portfolio so that all holding s have the same currency and that fund data have already been unpacked as far as the user wants/needs to.

jacobvjk:
closes #445 

This PR:

- adds an alternative second step to the offline PACTA workflow
- this step can be used to prepare the portfolio data whenever the user needs to run PACTA offline on their own machines and cannot grant us access to any of their data
- It requires different input data from our side (and the user's side!), as described [here](https://docs.google.com/document/d/1p-k4M1-ku8y6UIsFrZLhAdUzNoZDcMiLn36RIAxZkoY/edit#heading=h.7fb5m5qd2r64)
- It leads to a new workflow, described in detail [here](https://docs.google.com/document/d/1rBMLBOtQJGwoQVGjy6qRH71hiv9XFeXMI86eE50yunQ/edit#)
- introduces a boolean that indicates whether the script is to be run remotely or not
- runs checks and definitions of column types based on that boolean as they differ between the workflows.

Note:

- I am absolutely open to discussing the implementation of this workflow, as some of the choices I made are rather for reasons of getting this out soon than for thoroughly thought through reasons regarding future restructuring of the package.

### DIFF
--- a/1_portfolio_check_initialisation.R
+++ b/1_portfolio_check_initialisation.R
@@ -27,6 +27,7 @@ source("0_sda_approach.R")
 
 project_name <- "BondsTest"
 twodii_internal <- TRUE
+run_remotely <- FALSE
 # TRUE or FALSE: TRUE means that the code is running on a 2dii laptop with dropbox connection
 
 #####################################################################
--- a/2b_portfolio_prep_supervisor_workflow.R
+++ b/2b_portfolio_prep_supervisor_workflow.R
@@ -1,0 +1,28 @@
+financial_data <- readRDS(file.path(analysis_inputs_path, "financial_data.rda"))
+raw_pf <- read_csv(file.path(raw_input_path, paste0(project_name, "_Input.csv")),
+                   col_types = "cccnnccn")
+
+
+prep_pf <- raw_pf %>%
+  inner_join(financial_data, by = "isin") %>%
+  add_holding_id() %>%
+  rename(current_shares_outstanding_all_classes = current_shares_outstanding)
+
+port_raw_all_eq <- prep_pf %>%
+  filter(asset_type == "Equity") %>%
+  mutate(
+    id = company_id,
+    id_name = "company_id"
+  )
+
+port_raw_all_eq %>% write_rds(file.path(proc_input_path, paste0(project_name, "_equity_portfolio.rda")))
+
+
+port_raw_all_cb <- prep_pf %>%
+  filter(asset_type == "Bonds") %>%
+  mutate(
+    id = corporate_bond_ticker,
+    id_name = "corporate_bond_ticker"
+  )
+
+port_raw_all_cb %>% write_rds(file.path(proc_input_path, paste0(project_name, "_bonds_portfolio.rda")))
--- a/2b_portfolio_prep_supervisor_workflow.R
+++ b/2b_portfolio_prep_supervisor_workflow.R
@@ -1,3 +1,5 @@
+# use this script instead of script 2, in case the logical run_remotely == TRUE
+
 financial_data <- readRDS(file.path(analysis_inputs_path, "financial_data.rda"))
 raw_pf <- read_csv(file.path(raw_input_path, paste0(project_name, "_Input.csv")),
                    col_types = "cccnnccn")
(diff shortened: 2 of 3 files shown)


jacobvjk:
re R CMD checks on mac: `Error: Failed to get R 4.0.5`

maurolepore:
> re R CMD checks on mac: Error: Failed to get R 4.0.5

https://github.com/2DegreesInvesting/r2dii.plot.static/pull/76#pullrequestreview-633809317

maurolepore (approved):
LGTM but I'm not a good person to judge this code -- 'm not familiar with it and and have no way to easily see the impact of the change (e.g. reproducible example or test). 

I'm inclined to approve without much investigation because:

* based on the PR's first comment -- this changes do not impact the user-facing "online" tool so it feels quite safe. Also, having the

* I think we should have only one code base and the one it's now public is the so called "online" tool. So I suspect that changes to the "offline" tool won't affect the code in the long term.

The only general comment I can make is about the commended code. Why not remove it?

jacobvjk:
@maurolepore this is the part I am open to talking about. I don't remove it, because this works specifically for the new remote offline version. In principle, I think the files could be adjusted in the normal offline version as well to all fit the same logic, but that involves a few loops with AR and data_preparation. The main difference being that I have replaced bbg IDs in some of the data sets with AR's company IDs, because we cannot share bbg IDs. The type difference is that bbg IDs are mixed character and numbers and therefore the id column was cast as character, whereas the AR id is numeric. so in order to avoid issues with a join, I removed this. The following check regarding col types could probably be kept/adjusted as well, but since there is no documentation about the col names that refer to the order of these col types, it's hard to know haw it would have to be adjusted. Maybe @cjyetman can also chime in on this?

cjyetman:
@jacobvjk Can we have a short chat about this today? Honestly, it pains me to see yet another separate workflow for PACTA rather than one universal one, but I don't want to make an unnecessary blocker for you.

jacobvjk:
@cjyetman absolutely, I'm by no means tied to doing it this way. It's really more a suggestion that I'm hoping to turn into something that won't cause further headaches

cjyetman:
FYI... macOS CI is working again... I reran the jobs so it's passing now

jacobvjk:
so I added another boolean in the initialisation of the project in script 1. Based on that the user either should run the standard script 2 or 2b. For now they are separate files, but if we find consensus that this should rather all happen in script 2 and be decided by the boolean as well, then I am happy to merge that into one script. In script 3, column types and checks depend on the boolean as well. fyi @cjyetman

maurolepore:
I'm happy for you to merge whenever you are ready, based on your conversation with CJ and personal judgement.

PR #450, opened 2021-04-27
number_of_shares should always be numeric

cjyetman:
if number_of_shares is not in the original portfolio CSV, then it is added... but it should be added as a numeric vector filled with NAs rather than a logical vector filled with NAs

### DIFF
--- a/0_portfolio_input_check_functions.R
+++ b/0_portfolio_input_check_functions.R
@@ -200,7 +200,7 @@ check_missing_cols <- function(portfolio, grouping_variables) {
   required_input_cols <- c("holding_id", "market_value", "currency", "isin", grouping_variables, "number_of_shares")
 
   if (!"number_of_shares" %in% colnames(portfolio)) {
-    portfolio$number_of_shares <- NA
+    portfolio$number_of_shares <- NA_real_
   }
 
 


0_portfolio_input_check_functions.R
```diff
@@ -200,7 +200,7 @@ check_missing_cols <- function(portfolio, grouping_variables) {
   required_input_cols <- c("holding_id", "market_value", "currency", "isin", grouping_variables, "number_of_shares")
 
   if (!"number_of_shares" %in% colnames(portfolio)) {
-    portfolio$number_of_shares <- NA
+    portfolio$number_of_shares <- NA_real_
```
AlexAxthelm, line 203: Are fractional shares permissible? If so, this looks good. Otherwise, would `NA_integer_` be appropriate?

AlexAxthelm (approved):

0_portfolio_input_check_functions.R
cjyetman: good question...

1. yes, sort of.... if number of shares is not defined by the portfolio file, it is calculated (for Equity) by market_value / share_price.... I don't know if there's a strong argument for **not** rounding those values
2. the check that prompted this checks for numeric class... though that could also be changed presumably, but clearly logical is not ideal, which is what `NA` will evaluate to.... https://github.com/2DegreesInvesting/PACTA_analysis/blob/11ea54588f9b618c6f1b09c5f712146f5f0628f3/0_portfolio_input_check_functions.R#L85-L96

AlexAxthelm: Works for me then. :shipit:

PR #456, opened 2021-05-27
add a file encoding check action on PR

cjyetman:
adds a GitHub Action that will trigger on any commit on a PR, that will check the file encoding of every file that has been modified in that PR (unless it's a binary file) and trigger and error if any of the files are not either UTF-8 or ASCII

### DIFF
--- a/.github/workflows/pr-check-file-encodings.yml
+++ b/.github/workflows/pr-check-file-encodings.yml
@@ -1,0 +1,21 @@
+name: check file encodings in PR
+on: [pull_request]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


maurolepore (approved):
LGTM. I believe this builds on the lessons learned from a similar PR in a separate repo.

cjyetman:
> I believe this builds on the lessons learned from a similar PR in a separate repo.

yes, exactly.... I waited to push it here until I was fully confident with it working elsewhere...
https://github.com/2DegreesInvesting/data-steward-handover/actions/workflows/pr-file-encodings.yaml
https://github.com/2DegreesInvesting/data_preparation/actions/workflows/pr-check-file-encodings.yml

PR #457, opened 2021-05-28
Create PA2021PE

NayraHerrera:
Parameters file

PR #459, opened 2021-05-28
Add parameters file for Peru.

AlexAxthelm:


cjyetman:
this appears to be doing the same thing as #457, no?

AlexAxthelm:
You appear to be correct 🤦 I should check the PRs in the future. I'll close that one, and merge in #457

PR #464, opened 2021-06-09
use "null" instead of "NA" for JSON output coveragetextvar.json

cjyetman:
this should fix a broad class of issues with the audit function on the platform because any "NA" values in `coveragetextvar.json` were causing the front end to show a spinning cog indefinitely with no error, message, or explanation... which could have been the case in many situations, e.g. there were no valid bonds or no valid equity holdings in the portfolio

fyi @catarinabrg 
🤞🏻

### DIFF
--- a/0_json_functions.R
+++ b/0_json_functions.R
@@ -358,22 +358,22 @@ export_audit_textvar_json <- function(portfolio_total_, export_path_full) {
 
   included <-
     portfolio_total_ %>%
-    filter(has_valid_input == TRUE) %>%
-    pull(value_usd) %>%
+    dplyr::filter(has_valid_input == TRUE) %>%
+    dplyr::pull(value_usd) %>%
     sum(na.rm = TRUE) %>%
     round()
 
   bonds <-
     portfolio_total_ %>%
-    filter(asset_type == "Bonds") %>%
-    pull(value_usd) %>%
+    dplyr::filter(asset_type == "Bonds") %>%
+    dplyr::pull(value_usd) %>%
     sum(na.rm = TRUE) %>%
     round()
 
   equity <-
     portfolio_total_ %>%
-    filter(asset_type == "Equity") %>%
-    pull(value_usd) %>%
+    dplyr::filter(asset_type == "Equity") %>%
+    dplyr::pull(value_usd) %>%
     sum(na.rm = TRUE) %>%
     round()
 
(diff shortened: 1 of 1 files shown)


0_json_functions.R
```diff
@@ -354,41 +354,32 @@ export_audit_invalid_data <- function(portfolio_total_, export_path_full) {
...
-
-  keys <- c("\"total\"", "\"included\"", "\"bonds\"", "\"equity\"")
-  values <- c(total_v, included_v, bonds_v, equity_v)
-
-
-  json_head <- "{"
-  json_tail <- "}"
+  included <-
+    portfolio_total_ %>%
+    filter(has_valid_input == TRUE) %>%
```
AlexAxthelm, line 361: Non-blocking: for future work, can we get in the habit of calling it as `dplyr::filter()`?

AlexAxthelm (changes requested):
You're adding a dependency on `jsonliite::toJSON`, but not putting it in `DESCRIPTION` Other than that, this seems reasonable to me.

0_json_functions.R
cjyetman: wrt to `jsonlite`: I verified that the Docker has `jsonlite` installed, at least according to [the Docker file that creates the image that it is based on](https://github.com/2DegreesInvesting/docker/blob/9efe76454b91a45f47e32da5d5d7d440d6a8a644/r-packages/Dockerfile#L28). As far as it being in `DESCRIPTION`... that's kinda a Mauro thing because this repo is not an R package so I basically ignore that stuff.

cjyetman: wrt to `dplyr::filter()`: sure, fair, I'll make that change

AlexAxthelm (approved):
Looks Good :shipit:

0_json_functions.R
AlexAxthelm: > wrt to `jsonlite`: I verified that the Docker has `jsonlite` installed, at least according to [the Docker file that creates the image that it is based on](https://github.com/2DegreesInvesting/docker/blob/9efe76454b91a45f47e32da5d5d7d440d6a8a644/r-packages/Dockerfile#L28). As far as it being in `DESCRIPTION`... that's kinda a Mauro thing because this repo is not an R package so I basically ignore that stuff.

@maurolepore, FYI

cjyetman:
thanks... let me merge tomorrow after I test those latest changes locally

0_json_functions.R
AlexAxthelm: > wrt to `dplyr::filter()`: sure, fair, I'll make that change

switching to `dplyr::filter()` isn't a huge issue, but i've occasionally run into problems in the past where R reverts back to `base:filter()` in certain situations. It's a pain of a bug to reproduce, but it's pretty easy to avoid.

cjyetman: indeed

PR #467, opened 2021-07-05
771 meta investor supervisor workflow

jacobvjk:
closes ADO https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/771

This PR:
- adds the creation of a meta portfolio to the offline remote supervisor workflow
- this did not exist before, but is an important part of the analysis for many users
- it does not affect any other workflows
- the required parameter is supplied through `AnalysisParameters.yml` and calling the global parameters in the first step of the analysis
- in the `3_run_analysis` script, a magic number is replaced with the corresponding configuration

### DIFF
--- a/2b_portfolio_prep_supervisor_workflow.R
+++ b/2b_portfolio_prep_supervisor_workflow.R
@@ -10,6 +10,8 @@ prep_pf <- raw_pf %>%
   add_holding_id() %>%
   rename(current_shares_outstanding_all_classes = current_shares_outstanding)
 
+prep_pf <- add_meta_portfolio(prep_pf, inc_meta_portfolio)
+
 port_raw_all_eq <- prep_pf %>%
   filter(asset_type == "Equity") %>%
   mutate(
--- a/3_run_analysis.R
+++ b/3_run_analysis.R
@@ -209,7 +209,7 @@ if (file.exists(bonds_inputs_file)) {
 gather_and_save_project_results(results_path, aggregation_level = "portfolio")
 gather_and_save_project_results(results_path,
   aggregation_level = "company",
-  year_filter = c(start_year, start_year + 5),
+  year_filter = c(start_year, start_year + time_horizon),
   allocation_filter = "portfolio_weight",
   portfolios_per_file = 250
 )


3_run_analysis.R
```diff
@@ -209,7 +209,7 @@ if (file.exists(bonds_inputs_file)) {
 gather_and_save_project_results(results_path, aggregation_level = "portfolio")
 gather_and_save_project_results(results_path,
   aggregation_level = "company",
-  year_filter = c(start_year, start_year + 5),
+  year_filter = c(start_year, start_year + time_horizon),
```
cjyetman, line 212: I can't easily tell if/where `time_horizon` comes from and if it will always be available at this point. Not saying it's wrong, just hard for me to verify. I trust your judgement here.

2b_portfolio_prep_supervisor_workflow.R
```diff
@@ -10,6 +10,8 @@ prep_pf <- raw_pf %>%
   add_holding_id() %>%
   rename(current_shares_outstanding_all_classes = current_shares_outstanding)
 
+prep_pf <- add_meta_portfolio(prep_pf, inc_meta_portfolio)
```
cjyetman, line 13: I honestly don't know what `add_meta_portfolio()` is or what it does, nor what `inc_meta_portfolio` is in this context, but assuming it does what you want, this makes total sense to me.

cjyetman (approved):

3_run_analysis.R
jacobvjk: it comes form `AnalysisParameters.yml` or `ProjecParameters.yml` (depending on the context. It is available as soon as we run `set_global_parameters()` or `set_project_parameters()`

2b_portfolio_prep_supervisor_workflow.R
jacobvjk: It simply duplicates the entire portfolio and adds investor_name = "Meta Investor" and portfolio_name = "Meta Portfolio" for every row in the duplicate, so that the entire input can be processed as one summary analysis across investors and portfolios. This is only relevant in offline PACTA I think, since it is a common case to process many portfolios by many investors at the same time in that scenario.

PR #471, opened 2021-08-02
Update package-documentation

maurolepore:
This PR updates the documentation of the PACTA.analysis package.

Note that the source of the documentation did not change, only
the rendered man/ file. This may be because readr updated the
source, and we inherit documentation from readr. So running
`devtools::document()` with no changes in `R/<filename>.R` does 
produce changes in `man/<filename>.Rd`.﻿


cjyetman (approved):
weird, but.... seems harmless to me

PR #472, opened 2021-08-02
Dockerize this repo

maurolepore:
Relates to [AB#964](https://dev.azure.com/2DegreesInvesting/a5201d6a-87ad-4acf-b862-9cba0e1ae681/_workitems/edit/964)
Relates to [AB#847](https://dev.azure.com/2DegreesInvesting/a5201d6a-87ad-4acf-b862-9cba0e1ae681/_workitems/edit/847)

This PR dockerizes this repo. The image can be used as usual from the terminal or interactively via RStudio server. For details see [computing-environment.md](https://github.com/2DegreesInvesting/PACTA_analysis/blob/dockerize-this-repo/computing-environment.md).

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -59,5 +59,4 @@
 ^computing-environment\.Rmd$
 ^computing-environment\.md$
 ^docker-compose\.yml$
-^helpers\.Rmd$
 ^\.env$
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -55,3 +55,9 @@
 ^workflow-pacta-cop\.Rmd$
 ^workflow\.Rmd$
 ^workflow\.md$
+^Dockerfile$
+^computing-environment\.Rmd$
+^computing-environment\.md$
+^docker-compose\.yml$
+^helpers\.Rmd$
+^\.env$
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,4 @@ integration-test.html
 
 # local test case test output
 test_cases/*
+.env
--- a/.gitignore
+++ b/.gitignore
@@ -53,4 +53,6 @@ integration-test.html
 
 # local test case test output
 test_cases/*
+
+# Other
 .env
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,5 @@
+FROM rocker/verse:latest
+MAINTAINER "Mauro Lepore" maurolepore@gmail.com
+
+COPY docker /docker
+RUN Rscript /docker/install.R
--- a/computing-environment.Rmd
+++ b/computing-environment.Rmd
@@ -1,0 +1,94 @@
+---
+output: github_document
+---
+
+```{r setup, include=FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>",
+  config_general = "man/figures/README-",
+  out.width = "100%",
+  cache = TRUE
+)
+
+devtools::load_all()
+```
+
+```{r helpers, include=FALSE}
+#' Look into a file
+#' @examples
+#' tmp <- tempfile()
+#' writeLines("a\nb\nc", tmp)
+#' look_into(tmp)
+#' look_into(tmp, 2)
+look_into <- function(path, n = -1L) {
+  lines <- readLines(path, n, encoding = "UTF-8")
+  writeLines(lines)
+  
+  invisible(path)
+}
+
+#' Find a file matchin a regexp
+#' @examples
+#' find_file("[.]profile$", "~")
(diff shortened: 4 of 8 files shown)


jdhoffa:
FYI. I get the following error when I try to run the dockerfile added here: 

Screenshot 2021-08-03 at 09 48 02

cjyetman (commented):
can you put all this stuff in a sub-directory please? I'd rather not add more mess to the already drastically cluttered root directory

maurolepore:
This PR may be closed in favor of whatever other approach @AlexAxthelm and @jdhoffa come up with. 

But answering @cjyetman's question, it seems it's possible to move docker-compose.yml and Dockerfile elsewhere but [./ is the default path](https://docs.docker.com/compose/compose-file/compose-file-v3/#service-configuration-reference). I agree the root of this project is already polluted but it seems that the Docker framework really likes this directory. Maybe more experienced Docker devleopers (cc' @AlexAxthelm) have an informed opinion?

maurolepore:
Thanks @jdhoffa. I think I got wrong the meaning of the "version" field. I thought it was an arbitrary number as the one we give to R packages, but reading more carefully I now think it's the version of the docker-compose syntax which in turn needs to be compatible with the version of my Docker engine. 

I note you built a docker-compose.yml with `version: "3.2" so I'm using that version here too -- just to be sure it'll work for you. Can you confirm?

--

What's surprising is that with the wrong version I did not get an error:

image



Even if this PR goes nowhere I'm curious so I pushed a change to "0.0.1". Can you see if that works for you now?

maurolepore:
Relates to AB#964

jdhoffa:
Weird, it definitely doesn't work for me if the version number is funky: 
Screenshot 2021-08-04 at 10 53 06

works fine with version `3.2`:
Screenshot 2021-08-04 at 11 17 55

Also NOTE: `compose` is now part of the `docker` CLI, so we can call `docker compose` instead of `docker-compose` :-)

cjyetman:
> @jdhoffa `docker compose` (rather than `docker-compose`) is (I believe) still in beta. I've had a few issues with getting consistent results when I tried space over dash.

yes!

jdhoffa:
Good to know!

jdhoffa:
I am unable to login with the credentials outlined in `./computing_environment.md`.

maurolepore:
I'm conflicted. I acknowledge new comments here -- in a closed PR -- might be distracting, but I also find the discussion here useful for me to learn more about docker and compose -- particularly how it might behave differently on different computers (maybe because of different versions of the docker engine or compose?). So I'll answer assuming you can unsubscribe, but I'm also okay if anyone wants to lock the conversation and move the discussion elsewhere.

--

> I am unable to login with the credentials outlined in ./computing_environment.md. -- @jdhoffa 

Weird. So you fail to login even though this is true?

1. You have ./env with:

```bash
RSTUDIO_PASSWORD=abc
```

2. You use these credentials:

Username: rstudio
Password: abc

AlexAxthelm:
Let's keep the conversation here. I agree that it's not the ideal plaace, but it's already flowing, and in a place where we can refer to it later.

and for reference, I'm able to start the Rstudio session and login.

jdhoffa:
> @jdhoffa `docker compose` (rather than `docker-compose`) is (I believe) still in beta. I've had a few issues with getting consistent results when I tried space over dash.

So the only reason I mentioned it is because I'm now getting messages from docker CLI to start using it without the dash. Is it still in beta if they're recommending using it? 

Screenshot 2021-08-05 at 10 19 02

jdhoffa:
So it doesn't seem to work for me with either `abc` or `"abc"`. Am I doing something silly? 

With `RSTUDIO_PASSWORD=abc`
Screenshot 2021-08-05 at 10 24 39

Screenshot 2021-08-05 at 10 25 27

With `RSTUDIO_PASSWORD="abc"`
Screenshot 2021-08-05 at 10 30 23
Screenshot 2021-08-05 at 10 30 41

maurolepore:
@jdhoffa this thread seems related to your problem: https://community.rstudio.com/t/rstudio-server-open-source-login-issue/14179/4

I'm curious to see if this simpler way to run RStudio Server also exposed the problem (from https://hub.docker.com/r/rocker/rstudio):

```bash
docker run --rm -p 8787:8787 -e PASSWORD=yourpasswordhere rocker/rstudio
```

This is simpler in that  (a) rocker/rstudio is more basic than rocker/verse, and (b) the password is passed as an argument to `docker run` instead of as an environmental variable devfined in ./env.

Also I'm not sure if the password needs to meet some criteria, so maybe  try different kinds of passwords, e.g. imshort, iamalongpassword, alpha, alphanum123?

PR #473, opened 2021-08-03
933 supervisor wf flags and overview

jacobvjk:
closes ADO 933: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/933

all changes below refer to the supervisor workflow.

These changes are required to provide the relevant files to the stress testing script after running PACTA EQ/CB with the supervisor workflow

This PR:
- adds some validity flags to the portfolio in the preparation script of the supervisor workflow
- returns `total_portfolio` in the supervisor workflow and saves it to the `30_Processed_Inputs` directory
- returns `overview_portfolio` in the supervisor workflow and saves it to the `30_Processed_Inputs` directory
- adds functionality to `get_ald_raw()` in `portfolio_test.R`, so that it will automatically read the correct file in the supervisor workflow and go on as before in any other workflow

### DIFF
--- a/0_portfolio_test.R
+++ b/0_portfolio_test.R
@@ -33,21 +33,21 @@ get_ald_scen <- function(portfolio_type) {
   return(ald)
 }
 
-get_ald_raw <- function(portfolio_type) {
+get_ald_raw <- function(portfolio_type, supervisor_workflow = FALSE) {
+  if (supervisor_workflow == TRUE) {
+    filename_eq_raw <- "equity_ald_scenario_map.rda"
+    filename_cb_raw <- "bonds_ald_scenario_map.rda"
+  } else {
+    filename_eq_raw <- "masterdata_ownership_datastore.rda"
+    filename_cb_raw <- "masterdata_debt_datastore.rda"
+  }
+
   if (portfolio_type == "Equity") {
-    if (twodii_internal == FALSE & run_remotely == TRUE) {
-      ald_raw <- read_rds(paste0(analysis_inputs_path, "/equity_ald_scenario_map.rda"))
-    } else {
-      ald_raw <- read_rds(paste0(analysis_inputs_path, "/masterdata_ownership_datastore.rda"))
-    }
+    ald_raw <- read_rds(file.path(analysis_inputs_path, filename_eq_raw))
   }
 
   if (portfolio_type == "Bonds") {
-    if (twodii_internal == FALSE & run_remotely == TRUE) {
-      ald_raw <- read_rds(paste0(analysis_inputs_path, "/bonds_ald_scenario_map.rda"))
-    } else {
-      ald_raw <- read_rds(paste0(analysis_inputs_path, "/masterdata_debt_datastore.rda"))
-    }
+    ald_raw <- read_rds(file.path(analysis_inputs_path, filename_cb_raw))
   }
 
 
--- a/0_portfolio_test.R
+++ b/0_portfolio_test.R
@@ -35,11 +35,19 @@ get_ald_scen <- function(portfolio_type) {
 
 get_ald_raw <- function(portfolio_type) {
   if (portfolio_type == "Equity") {
-    ald_raw <- read_rds(paste0(analysis_inputs_path, "/masterdata_ownership_datastore.rda"))
+    if (twodii_internal == FALSE & run_remotely == TRUE) {
+      ald_raw <- read_rds(paste0(analysis_inputs_path, "/equity_ald_scenario_map.rda"))
+    } else {
+      ald_raw <- read_rds(paste0(analysis_inputs_path, "/masterdata_ownership_datastore.rda"))
+    }
   }
 
   if (portfolio_type == "Bonds") {
-    ald_raw <- read_rds(paste0(analysis_inputs_path, "/masterdata_debt_datastore.rda"))
+    if (twodii_internal == FALSE & run_remotely == TRUE) {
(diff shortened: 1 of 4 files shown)


0_portfolio_test.R
```diff
@@ -35,11 +35,19 @@ get_ald_scen <- function(portfolio_type) {
 
 get_ald_raw <- function(portfolio_type) {
   if (portfolio_type == "Equity") {
-    ald_raw <- read_rds(paste0(analysis_inputs_path, "/masterdata_ownership_datastore.rda"))
+    if (twodii_internal == FALSE & run_remotely == TRUE) {
```
cjyetman, line 38: this makes me extremely nervous.... `get_ald_raw()` now has new dependencies on `twodii_internal` and `run_remotely` being available in the global environment, which makes debugging and stepping through the code more difficult, and could break something if `get_ald_raw()` ever gets called without it.

I would suggest...

Make `twodii_internal` and `run_remotely` (or one combined object/variable) arguments of the function `get_ald_raw()` and set sane defaults so that the default behaviour is always what it was before this change... then wherever you want to call this function and have a different behaviour (which I assume must be the case because otherwise you wouldn't be making this change), call this function and explicitly set this/these variable/s so that the function behaves the way you need it there. e.g.....

```r
get_ald_raw <- function(portfolio_type, stress_test_way == FALSE) {
  
  if (stress_test_way) {
    eq_filename <- "equity_ald_scenario_map.rda"
    bonds_filename <- "bonds_ald_scenario_map.rda"
  } else {
    eq_filename <- "masterdata_ownership_datastore.rda"
    bonds_filename <- "masterdata_debt_datastore.rda"
  }

  if (portfolio_type == "Equity") {
    ald_raw <- read_rds(file.path(analysis_inputs_path, eq_filename))
  }

  if (portfolio_type == "Bonds") {
    ald_raw <- read_rds(file.path(analysis_inputs_path, bonds_filename))
  }
}
```

then wherever you want this new behaviour, use...
```r
get_ald_raw(portfolio_type, stress_test_way == TRUE)
```

2b_portfolio_prep_supervisor_workflow.R
```diff
@@ -12,7 +12,34 @@ prep_pf <- raw_pf %>%
...
+portfolio_total <- prep_pf %>%
+  mutate(valid_input = case_when(
+    !has_valid_value_usd ~ FALSE,
+    !has_valid_isin ~ FALSE,
+    TRUE ~ TRUE
+  ))
+
+portfolio_overview <- portfolio_summary(portfolio_total)
+
+portfolio_overview %>% write_rds(file.path(proc_input_path, paste0(project_name, "_overview_portfolio.rda")))
```
cjyetman, line 39: if possible, please save RDS files with an ".rds" extension

2b_portfolio_prep_supervisor_workflow.R
```diff
@@ -12,7 +12,34 @@ prep_pf <- raw_pf %>%
...
+    )
+  )
+
+# taken from overall_validity_flag()
+portfolio_total <- prep_pf %>%
+  mutate(valid_input = case_when(
+    !has_valid_value_usd ~ FALSE,
+    !has_valid_isin ~ FALSE,
+    TRUE ~ TRUE
+  ))
```
cjyetman, line 35: is this the same as
```r
mutate(valid_input = all(has_valid_value_usd, has_valid_isin))
```
?

0_portfolio_test.R
jacobvjk: valid point, will take care of that

2b_portfolio_prep_supervisor_workflow.R
jacobvjk: I would much prefer changing all of these at once for the entire PACTA repo, rather than just doing it here. This file will be saved as an rda in other workflows as well after all. Maybe it's time for a clean up soon?

0_portfolio_test.R
AlexAxthelm: I agree with @cjyetman that this adds an invisible dependency. A different strategy would be to define a checker function, and set some flags via envvars or options, but I would prefer explicit arguments when possible.

2b_portfolio_prep_supervisor_workflow.R
cjyetman: yeah, I'm all too aware that this is unavoidable in some cases... just always have to say it when I see it 😬

0_portfolio_test.R
jacobvjk: implemented

2b_portfolio_prep_supervisor_workflow.R
jacobvjk: looks like it is. I stuck to the same style as in `add_portfolio_flags()`.. but I can change that, if preferred

jacobvjk:
@cjyetman @AlexAxthelm this is ready to re-review anytime :)

2b_portfolio_prep_supervisor_workflow.R
cjyetman: fine as this, just thought the syntax was bit over-complicated... if it reads well to you, leave it

0_portfolio_test.R
cjyetman: perfect! thanks for being flexible with that

cjyetman (approved):
lgtm

PR #476, opened 2021-08-05
Add basic dockerfile

maurolepore:
Relates to [AB#966](https://dev.azure.com/2DegreesInvesting/a5201d6a-87ad-4acf-b862-9cba0e1ae681/_workitems/edit/966)


maurolepore:
@jdhoffa I'm making this a PR to make it easy to work.

DESCRIPTION
```diff
@@ -1,48 +1,51 @@
...
-             role = "aut",
-             email = "clare@2degrees-investing.org"),
-      person(given = "Jacob",
-             family = "Kastl",
-             role = "aut",
-             email = "jacob.kastl@gmail.com"),
       person(given = "2 Degrees Investing Initiative",
-             role = c("cph", "fnd"),
+             role = c("aut", "ctr", "cre"),
              email = "contact@2degrees-investing.org"))
```
jdhoffa, line 7: I think there's a mismatched bracket here actually

jdhoffa:
Sounds good!

maurolepore:
@jdhoffa 

The commit above produces an image that seems to successfully runs pacta. Note it focuses on the computing environment only and relies on mounting PACTA_analysis and pacta-data. I'm pretty sure the approach of copying PACTA_analysis also works, but for now I prefer to do it this way -- mainly so we can push the image to dockerhub and pull it from GH Actions, to prove this works on CI. Happy to discuss and revert.

I built the image 'test' with:

```bash
    cd /home/mauro/git/PACTA_analysis
    docker build -t test .
```

Then I run pacta with:

```bash
    docker run  --rm -ti --workdir /root/PACTA_analysis -v /home/mauro/git/:/root test ./bin/run-pacta.sh
```

Details:

- prune DESCRIPTION to include the relevant packages only.
- prune Dockerfile to focus only on the computing environment.
- New bin/run-pacta.sh
TODO: Edit docker-compose.yml to do what this long command does.

jdhoffa:
Thanks a lot @maurolepore, I have two main comments here, touching two topics: 

- The functionality of the PR on my machine currently
- Critiques of the "mounting all directories" approach

## Functionality
I ran 
``` bash
docker build -t test .
docker run --rm -ti --workdir /root/PACTA_analysis -v ~/Documents/github/2dii:/root test ./bin/run-pacta.sh
```
which gave me the following error: 
Screenshot 2021-08-06 at 11 28 28

`R` doesn't have a "Killed" error, which leads me to believe this is an operating system memory constraint? Maybe I need to adjust my docker settings to give them access to more of the host machine's memory? (The only reason I think it must be my docker settings, is because you aren't getting this error). 

I hopped into the container to poke around, and it is definitely the second `web_tool_script_2.R` that spits out the error (which is where all the analysis happens, so would make sense if it is a memory constraint): 
Screenshot 2021-08-06 at 11 27 35
Screenshot 2021-08-06 at 11 27 42
Screenshot 2021-08-06 at 11 27 48
Screenshot 2021-08-06 at 11 27 54

I will continue to mess around with my docker settings and see if I can figure it out. 

## Criticism of Mounting All Directories
I am a little against mounting `PACTA_analysis`, and the quickest way I can explain why is as follows. In my mental model, the `PACTA_analysis` directory is the entire "core" of the application, and I want to "freeze" it inside the image, and expose only the inputs and outputs of the application to the user, via mounts. 

One reason I think this is better than mounting the `PACTA_analysis` repo itself, is that we don't want "running the PACTA tool" to touch the source code of the app, and pollute the git staging area. For example, when I run the dockerfile as it is written above, my staging area has some fun artifacts pop-up: 
Screenshot 2021-08-06 at 11 44 33

To your point of being able to push this to dockerhub, I am curious what exactly we are trying to achieve there. The dockerfile as it exists right now just copies in a DESCRIPTION file from some repo, and installs dependencies, I'm not sure how useful this is to have on dockerhub? Maybe I misunderstand, can we push a container to dockerhub, where we build it using our locally defined mounts? 

I can imagine two interesting/ useful cases: 

1. We push a "standard 2dii" computing environment to dockerhub, with a frozen version of R, and a lot of useful tidyverse packages already installed/ prepped to run PACTA and other analyses
2. We push the "PACTA" tool itself to dockerhub (which would be fine to do actually, since the whole `PACTA_anaysis` repo is public), or conversely we push the tool to some private container registry. 

Can you elaborate on why copying `PACTA_analysis` into the image would prevent us from pushing to dockerhub/ running CI on GH actions/ etc.? 

Hope this makes sense, my mental model of these tools is still not totally clear, so if I am saying anything silly please point it out :-)

AlexAxthelm:
@jdhoffa I confirm that your `Killed` error looks like an Out Of Memory (OOM), enforced by the Linux process keeper.

I also agree with the spirit of your criticism of mounting the project as a volume, but I'll add some further thoughts in a more full review I'm writing now.

jdhoffa:
Yup, just confirming that adusting my docker resources to the following: 

- CPUS: 5
- Memory: 10GB
- Swap: 2.5GB
- Disk image size: 560GB

solves the issue. Those settings are probably overkill, but they work ¯\_(ツ)_/¯

DESCRIPTION
AlexAxthelm: Style: 
* for functions with arguments spanning multiple lines, open paren at end of line, close paren on own line.
* For functions with more than 2 arguments, we should split to multiple liines, with 1 argument per line.
* Name your arguments

examples:
```r
# good
foo <- bar(x = bax, y = "bay", z = c(baz, baw))

# better
foo <- bar(
  x = bax,
  y = "bay",
  z = c(baz, baw)
)

# don't do this
foo <- bar(bax,
  "bay",
  z = c(baz, baw))
```

DESCRIPTION
```diff
@@ -1,48 +1,39 @@
...
-Description: For more information visit
-    <https://2degrees-investing.org/>.
+Description: We abuse this DESCRIPTION file to make it easy for Docker to
```
AlexAxthelm, line 8: Blocking: If you're openly admitting that you're abusing a design pattern, it means you're probably using the wrong design pattern.

DESCRIPTION
```diff
@@ -1,48 +1,39 @@
-Package: PACTA.analysis
-Title: Helpers for 'PACTA_analysis'
+Package: PACTA.analysis.dependencies
```
AlexAxthelm, line 1: Ultra-Blocking:
🤦 I think I know where I miscommunicated, that led you to think this is the appropriate method. I mentioned in one of the other PRs that I prefer using DESCRIPTION to manage dependencies over the other methods. What I failed to follow up with is "...as part of a full migration to an R package". 

Using this mechanism in this context is hacky, and the entire point of this exercise is to reduce our tech debt. Going with this approach instead is giving us a partial conversion to a package, but missing out on most of the benefits.

I would rather have our dependencies defined in a script or in the dockerfile. We can chaat later about what that looks like, but I'm imagining something along the lines of:

```r
pacta_dependencies <- c(
  "foo",
  "bar",
  "bax"
)

install.packages(pacta_dependencies)
```

and in the dockerfile:

```
COPY  install_deps.R .
RUN [Rscript, "install_deps.R"]
```

Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM rocker/r-ver:latest
+
+RUN Rscript -e 'install.packages("remotes")'
+
+COPY DESCRIPTION /r-pkg/DESCRIPTION
```
AlexAxthelm, line 5: Reference: See comment in `DESCRIPTION`, line 1.

bin/run-pacta.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+tmp=$(tempfile)
+head -n 211 web_tool_script_2.R >> "$tmp"
```
AlexAxthelm, line 3: Blocking: Okay, I see that this is to remove running Stress Testing from the process, which is going to make things easier during development. That makes sense, but I don't like hard coding line numbers into scripts like this. 

More generally, don't edit scripts (like R files) as part of the execution process. If you want to  avoid running the ST components, wrap them in an `if` statement in the actual script, and make it conditional on either an R option or an envvar.

docker-compose.yml
```diff
@@ -0,0 +1,8 @@
+version: "3.9"
```
AlexAxthelm, line 1: Non-blocking: I'd prefer if we used an earlier syntax, since then users who (somehow) have older versions of docker don't need to update to run this.

docker-compose.yml
```diff
@@ -0,0 +1,8 @@
...
+services:
+  app:
+    command: /bin/bash
```
AlexAxthelm, line 4: Blocking: Don't define `command:` as part of docker-compose.yml. Define the default command as part of Dockerfile, and if you want to run an exceptional command as part of development, put that in `docker-compose.override.yml`

DESCRIPTION
AlexAxthelm: A slightly different version that would give us better version control might be something along the lines of:
```r
packages <- rbind(
  c("foo", "12.3"),
  c("Bar", ">= 1.2.3"),
  c("baz", "1.5")
)
for (i in nrow(packages)){
  remotes::install_version(package = packages[i,1], version = packages[i,2])
}
```

docker-compose.yml
```diff
@@ -0,0 +1,8 @@
...
+    build: .
+    volumes: 
+      - ../pacta-data:/pacta-data
```
AlexAxthelm, line 7: Suggestion: when in active development, I do find it useful to mount the entire project (or at least the code) directory as a volume, so that I can edit in my native (host) environment, and have those changes picked up instantly by the container. Helps reduce time when iterating, it requires running with `docker-compose -f docker-compose.yml up`, after there's a candidate, but at least you don't have to do that every time you make a little change.

AlexAxthelm (changes requested):
I see some major design choices that need to be changed before this gets much further along, so we don't introduce fragility and tech debt.

I know we're meeting later today, and we can talk more about this then, but the quck verrsion of the process I suggest  when developing in Docker is:

1. Set up your `docker-compose` files (`docker-compose.yml`, and `docker-compose.oveerride.yml`) to give you the mappings you want, with a nearly empty dockerfile (generally just `FROM:`)
2. Add in some detail to the Dockerfile, and try to run a single script
3. Add dependencies and alter dockerfile as needed
4. start trying to run more scripts each time
5. repeat steps 3 & 4 until the entire project is working as expected.

jdhoffa:
This PR was never intended to be merged, we just wanted an easy platform to collaborate:

> @jdhoffa I'm making this a PR to make it easy to work.

I have closed it, and opened an analogous PR on my local fork marked "DO NOT MERGE" to make it clear that is 100% experimental, and put everyone at ease.  https://github.com/jdhoffa/PACTA_analysis/pull/2

maurolepore:
Thanks @AlexAxthelm's for your comments. We'll be following up with changes on Jackson's fork.

Here I respond only to comments that seem to not need changes in code:

> @AlexAxthelm RE Code style on DESCRIPTION

We implement style automatically -- not manually -- with 
`usethis::use_tidy_description()` (for DESCRIPTION) and `usethis::use_tidy_style()` (for .R files). I'm not sure why usethis implements two different styles in those two different kinds of files.  I agree it's not ideal but I prefer to stick with the automatic tools and suffer from the small style inconsistency rather than to have to maintain the style manually.

> @jdhoffa RE why copying PACTA_analysis into the image would prevent us from pushing to dockerhub/ running CI on GH actions/ etc.?

We could still publish the built image to dockerhub and use it on GitHub Actions but it would be weird because GH-Actions workflows are more commonly used to test the code you have in a repo in one or multiple computing environment you specify (maybe as a matrix). In other words, the workflows encourage the separation between computing environment and code. For example, consider this workflow that on every push PACTA_analysis it runs the web tools scripts on a container from 2dii/r-packages:latest (hosted on rhub). I believe this is a typical use of GH-Actions, just as we do for R packages, we want to know if after some change in the source code the software continuous to work as expected in some known computing environments we specify:

```yaml
# .github/workflows/source-web-tool-scripts.yaml
name: Source web tool scripts in container r2dii/r-packages
on: push

# THIS IS THE COMPUTING ENVIRONMENT
jobs:
  Source-web-tool-scripts:
    container: 2dii/r-packages:latest
    ...
    steps:
      ...

    # THIS IS THE CODE
      - name: Checkout PACTA_analysis
        uses: actions/checkout@v2
        ...

     # THIS IS WHAT RUNS
      - name: Source web-tool scripts
        run: |
          setwd("PACTA_analysis")
          source("R/source_web_tool_scripts.R")
          source_web_tool_scripts()
        shell: Rscript {0}
```

PR #480, opened 2021-08-23
move stress testing to webtool 3 so webtool 1-2 is only the "PACTA" p…

cjyetman:
…rocess

this has been a long time coming.... I don't know the history of why stress testing was added to web tool script 2... I guess because the results of it were needed to create the interactive report? Anyway, this finally cleanly separates "doing PACTA" and "doing something with the results of PACTA". I think this is a good step toward a lot of the ideas/discussions we've had about making the web tool easier to use in different contexts, separating running PACTA from doing stuff with the PACTA outputs, etc.

To be honest, I didn't put a whole lot of thought into this, so I'd be happy if many of you take a look and share any thought or concerns. (also tagging @andybue who I can't add as a reviewer?)

### DIFF
--- a/web_tool_script_2.R
+++ b/web_tool_script_2.R
@@ -209,14 +209,6 @@ if (file.exists(bonds_inputs_file)) {
   }
 }
 
-# provide parameters for stress test
-invisible(set_portfolio_parameters(file_path = fs::path(par_file_path, paste0(portfolio_name_ref_all, "_PortfolioParameters.yml"))))
-# set environment variable for stress test data path
-options("ST_DATA_PATH" = stress_test_data_location)
-# run 2dii stress test
-source(file.path(stress_test_path, "web_tool_stress_test.R"))
-# run stress test with external scenarios (IPR)
-source(file.path(stress_test_path, "web_tool_external_stress_test.R"))
 
 rm(port_raw_all_eq)
 rm(port_raw_all_cb)
--- a/web_tool_script_3.R
+++ b/web_tool_script_3.R
@@ -32,6 +32,21 @@ if(project_code == "GENERAL"){
 
 analysis_inputs_path <- set_analysis_inputs_path(twodii_internal, data_location_ext, dataprep_timestamp)
 
+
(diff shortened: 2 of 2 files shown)


web_tool_script_2.R
```diff
@@ -209,14 +209,6 @@ if (file.exists(bonds_inputs_file)) {
...
-invisible(set_portfolio_parameters(file_path = fs::path(par_file_path, paste0(portfolio_name_ref_all, "_PortfolioParameters.yml"))))
-# set environment variable for stress test data path
-options("ST_DATA_PATH" = stress_test_data_location)
-# run 2dii stress test
-source(file.path(stress_test_path, "web_tool_stress_test.R"))
-# run stress test with external scenarios (IPR)
-source(file.path(stress_test_path, "web_tool_external_stress_test.R"))
 
 rm(port_raw_all_eq)
 rm(port_raw_all_cb)
```
maurolepore, line 214: I see the stress test code moves from before to after some global variables are removed `port_raw_all_eq`, `port_raw_all_cb`, etc.) Is this okay @jacobvjk? In other words, does the stress test code work without those variables?

maurolepore (approved):
LGTM, assuming it's safe to move the stress test code after a number of global variables are removed. See my comment for details.

web_tool_script_2.R
cjyetman: web tool scripts 1-3 all run independently in separate environments, i.e. webtool script 3 cannot rely on anything being set in webtool script 1 or 2... so all necessary environment variables should be set at the top of each script

jacobvjk: The stress test reads the pacta results from `working_dir/40_Results`. Whether or not any intermediate pacta results are removed from the environment in the meantime should pose no problem. As long as all paramters and paths are set, it should be good

jacobvjk (approved):
lgtm

jdhoffa:
FKN YAY WOO PARTY TIME <3 <3 <3

(this is my new alternative to LGTM)

PR #481, opened 2021-08-31
move transitionmonitor Docker image creator into the PACTA_analysis repo

cjyetman:


maurolepore (approved):
LGTM. 

What's the plan for the other images higher on this stack: 2dii/system and 2dii/r-packages? Will they stay in 2DegreesInvesting/docker or move here too? I'm not sure what's best but I'm curious to hear what others think makes more sense.

cjyetman:
> What's the plan for the other images higher on this stack: 2dii/system and 2dii/r-packages?

Honestly, I was hoping to move away from them. I tried basing off rocker/verse, but that didn't go so well, though that might have been due to architecture changes. Anyway, I decided to go for less surprises. So for now, it's just about moving the critical pieces here. Maybe we integrate the package installs here? Maybe we combine all 3 Docker files into one?

AlexAxthelm (approved):

cjyetman:
For clarity, these are the three Dockerfiles that I would like to fully integrate into one Dockerfile here in PACTA_analysis somehow...
https://github.com/2DegreesInvesting/docker/blob/master/system/Dockerfile
https://github.com/2DegreesInvesting/docker/blob/master/r-packages/Dockerfile
https://github.com/2DegreesInvesting/docker/blob/master/pacta/Dockerfile

cjyetman:
I experimented a bit more with using a rocker/verse image. Unfortunately, rocker/verse is built on rocker/rstudio, and from rocker/rstudio on, the image requires setting a password when starting an instance or it errors out, there's no clear way to me to get into a shell in the instance as is normal when running a docker instance. So those are rather big obstacles that I don't know how to get around immediately. Happy to hear any suggestions.

maurolepore:
> there's no clear way to me to get into a shell in the instance as is normal when running a docker instance ... Happy to hear any suggestions.


For context, this is the typical way to run a container to use rstudio from the web browser (similar to this [example](https://www.rocker-project.org/)):

```bash
docker run -e PASSWORD=yourpassword -p 8787:8787 rocker/verse
```

And this is how you can specify `bash` as an entry point (similar to one example [here](https://www.rocker-project.org/use/managing_users/)):


```bash
docker run --ti rocker/verse bash
```

*You may need `--user rstudio`:

> if you are sharing a local volume with the host, any files you create or modify will become owned by root. To avoid this, run interactive terminal sessions with the default [rstudio] user ... when sharing volumes with the host -- https://www.rocker-project.org/use/managing_users/

cjyetman:
> > there's no clear way to me to get into a shell in the instance as is normal when running a docker instance ... Happy to hear any suggestions.
> 
> 
> 
> 
> 
> For context, this is the typical way to run a container to use rstudio from the web browser (similar to this [example](https://www.rocker-project.org/)):
> 
> 
> 
> ```bash
> 
> docker run -e PASSWORD=yourpassword -p 8787:8787 rocker/verse
> 
> ```
> 
> 
> 
> And this is how you can specify `bash` as an entry point (similar to one example [here](https://www.rocker-project.org/use/managing_users/)):
> 
> 
> 
> 
> 
> ```bash
> 
> docker run --ti rocker/verse bash
> 
> ```
> 
> 
> 
> *You may need `--user rstudio`:
> 
> 
> 
> > if you are sharing a local volume with the host, any files you create or modify will become owned by root. To avoid this, run interactive terminal sessions with the default [rstudio] user ... when sharing volumes with the host -- https://www.rocker-project.org/use/managing_users/

Thanks. Yeah, I figured out the bash trick. It concerns me though because that could be a substantial change to how Constructiva runs the image, plus running a server, so using a rocker image might be a deal breaker.

AlexAxthelm:
the `PASSWORD` envvar s  only necessary if you actually start RStudio. you can bypass that to get  straight  into an R session with `docker run -it rocker/rstudio R`, but at that  point, we should reconsider if having the Rstudio tooling in the image is   worth it.

Instead, I'd prefer to have the PACTA image, and then have a second image for interactive work that extends that one, and  includes Rstudio.

(I'd also prefer to not base our production images on rocker, since they use `debian:stable` as a base, which has some known security issues. Nothing major, but I'd prefer to avoid introducing those if it's not terribly burdensome. The biig probleem iis that getting Rstudio to run on `alpine` is a _pain_.

cjyetman:
As far as I'm concerned (in the interest of transitionmonitor), RStudio is definitely not needed.

Between stable, testing and unstable, which distribution do you prefer?

AlexAxthelm:
> Between stable, testing and unstable, which distribution do you prefer?

Unless there is a compelling reason to use one of the others, we should always be  using  either LTS or stable releases. My issue is in  the choice of Debian over Alpine Linux for the base image. Debian has a lot more in it, and exposes a lot more attack surface.

cjyetman:
putting this into draft mode because I still have to test if this fully achieves the goal, which is dependent on recent and currently on-going changes in pacta and friends

cjyetman:
I will remove the rocker version of the Dockerfile from this PR, because I haven't got that fully working yet. Long term, I think we should try to move towards that.

I also want to specify explicitly the current version of the 2dii/r-packages Docker image that is used as the base, because I now know that recent changes in bookdown v0.22+ and R v4.1.0+ cause the transitionmonitor docker images to fail. Need to work out how to tag the current version, or maybe refer to it by it's sha (?)
https://github.com/2DegreesInvesting/PACTA_analysis/blob/7f3d7eb7c610977eaf3450fb36c3944fd81a3185/transitionmonitor_docker/Dockerfile#L1

Once the explicit version tagging is sorted, I will open this for review. I used it to create a hotfix Docker image that appeared to behave as expected, so... I'd prefer to move towards using this rather than the docker repo asap.

cjyetman:
rocker version removed, and we can target the specific base image with its sha, so this is ready to go.... last chance for comments

AlexAxthelm:
looks good. To be clear, the SHA256 referenced is the image hash from DockerHub, not a git commit?

cjyetman:
> looks good. To be clear, the SHA256 referenced is the image hash from DockerHub, not a git commit?

Correct

PR #482, opened 2021-09-08
enable setting data_location_ext based on the financial_timestamp in …

cjyetman:
enable setting data_location_ext based on the financial_timestamp in the project parameters

In the past, `data_location_ext` was set when running `setup_project()`, which sources [deduplicate/set_portfolio-name-ref-all_working-location_and_web-parameters.R](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/deduplicate/set_portfolio-name-ref-all_working-location_and_web-parameters.R), which calls [`set_web_parameters()`](https://github.com/2DegreesInvesting/PACTA_analysis/blob/897433f71c18edb6ba2cf6798f4c961158f27b0d/0_web_functions.R#L74-L89), which loads either [parameter_files/WebParameters_docker.yml](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/parameter_files/WebParameters_docker.yml) or [parameter_files/WebParameters_2dii.yml](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/parameter_files/WebParameters_2dii.yml), both of which hard code the location of the 2019Q4 data with `data_location_ext: ../pacta-data/2019Q4/`. Later, [`set_project_parameters()`](https://github.com/2DegreesInvesting/PACTA_analysis/blob/11ea54588f9b618c6f1b09c5f712146f5f0628f3/0_global_functions.R#L46-L128) is run, which loads the appropriate, project specific parameter file, which is where the `financial_timestamp` is usually set with something like `timestamp: 2020Q4`. There are also "portfolio parameters" that are loaded, but only the current "project parameter" files contain a timestamp which can then be used to determine which input data to use. It's possible that `setup_project()` was originally intended to load a file that contains all settings in one, rather than 3 separate parameter files with potentially overlapping settings, but as it is now, all 3 are necessary. That may actually be advantageous because I can imagine cases where there should be some cascading preferences, e.g. project sets default language to EN but user wants report in DE, but this process has not been well thought out.

In order to get 2020Q4 working, we need to set `data_location_ext` to `../pacta-data/2020Q4/`. It probably should be that `data_location_ext` is set to `../pacta-data/`, and then the `financial_timestamp` should be appended to it to get the final location of the data. However, that would likely require a change that would impact existing 2019Q4 accounts and their parameter files, so I'm avoiding that option.

Instead, what I do here is wait till all 3 parameter files have been loaded (in each of the web tool scripts because they all run in a unique environment). At that point `data_location_ext` should have been set by the "web parameters" file, and `financial_timestamp` should have been set by the "project parameters" file, so I'm simply combining them to set the real `data_location_ext` with `data_location_ext <- sub("[0-9]{4}Q[0-9]", financial_timestamp, data_location_ext)`.

Additionally, I've modified [`set_project_parameters()`](https://github.com/2DegreesInvesting/PACTA_analysis/blob/11ea54588f9b618c6f1b09c5f712146f5f0628f3/0_global_functions.R#L46-L128) so that it does not set `financial_timestamp`, which is as of yet is never included in the "web parameters" files, so it always sets `financial_timestamp` to `NULL`. This is particularly a problem when the stress test code runs and reloads the "web parameters" file but not the "project parameters" file.

These code changes should probably be commented, and the commented code deleted, but for now I left it like this or brevity.

interested in thoughts from @AlexAxthelm @jdhoffa @maurolepore @jacobvjk @MirijaHa @catarinabrg @daisy-pacheco @fariiaakh 

I believe that using the 2020Q4 data on transitionmonitor is blocked by https://github.com/2DegreesInvesting/r2dii.stress.test.data/issues/28 because the stress testing bit cannot complete without error

### DIFF
--- a/0_web_functions.R
+++ b/0_web_functions.R
@@ -85,7 +85,7 @@ set_web_parameters <- function(file_path) {
   twodii_internal <<- cfg$parameters$twodii_internal
   new_data <<- cfg$parameters$new_data
 
-  financial_timestamp <<- cfg$parameters$financial_timestamp
+  # financial_timestamp <<- cfg$parameters$financial_timestamp
 }
 
 set_portfolio_parameters <- function(file_path) {
--- a/web_tool_script_1.R
+++ b/web_tool_script_1.R
@@ -23,6 +23,9 @@ set_portfolio_parameters(file_path = fs::path(par_file_path, paste0(portfolio_na
 
 set_project_parameters(file.path(working_location, "parameter_files",paste0("ProjectParameters_", project_code, ".yml")))
 
+data_location_ext <- sub("[0-9]{4}Q[0-9]", financial_timestamp, data_location_ext)
+
+
 # need to define an alternative location for data files
 analysis_inputs_path <- set_analysis_inputs_path(twodii_internal, data_location_ext, dataprep_timestamp)
 
--- a/web_tool_script_1.R
+++ b/web_tool_script_1.R
@@ -23,7 +23,7 @@ set_portfolio_parameters(file_path = fs::path(par_file_path, paste0(portfolio_na
 
 set_project_parameters(file.path(working_location, "parameter_files",paste0("ProjectParameters_", project_code, ".yml")))
 
-data_location_ext <- sub("[0-9]{4}Q[0-9]", financial_timestamp, data_location_ext)
+data_location_ext <- sub("20[0-9]{2}Q[1-4]", financial_timestamp, data_location_ext)
 
 
 # need to define an alternative location for data files
--- a/web_tool_script_2.R
+++ b/web_tool_script_2.R
@@ -24,7 +24,7 @@ set_portfolio_parameters(file_path = fs::path(par_file_path, paste0(portfolio_na
 
 set_project_parameters(file.path(working_location, "parameter_files", paste0("ProjectParameters_", project_code, ".yml")))
 
-data_location_ext <- sub("[0-9]{4}Q[0-9]", financial_timestamp, data_location_ext)
+data_location_ext <- sub("20[0-9]{2}Q[1-4]", financial_timestamp, data_location_ext)
 
 # need to define an alternative location for data files
 analysis_inputs_path <- set_analysis_inputs_path(twodii_internal, data_location_ext, dataprep_timestamp)
--- a/web_tool_script_2.R
+++ b/web_tool_script_2.R
@@ -24,6 +24,8 @@ set_portfolio_parameters(file_path = fs::path(par_file_path, paste0(portfolio_na
 
 set_project_parameters(file.path(working_location, "parameter_files", paste0("ProjectParameters_", project_code, ".yml")))
 
+data_location_ext <- sub("[0-9]{4}Q[0-9]", financial_timestamp, data_location_ext)
+
 # need to define an alternative location for data files
 analysis_inputs_path <- set_analysis_inputs_path(twodii_internal, data_location_ext, dataprep_timestamp)
 
--- a/web_tool_script_3.R
+++ b/web_tool_script_3.R
@@ -17,6 +17,8 @@ set_portfolio_parameters(file_path = fs::path(par_file_path, paste0(portfolio_na
 
 set_project_parameters(file.path(working_location, "parameter_files",paste0("ProjectParameters_", project_code, ".yml")))
 
+data_location_ext <- sub("[0-9]{4}Q[0-9]", financial_timestamp, data_location_ext)
+
 if(project_code == "PA2020FL"){
   peer_group = case_when(
     peer_group %in% c("other")~ "Others",
--- a/web_tool_script_3.R
+++ b/web_tool_script_3.R
@@ -17,7 +17,7 @@ set_portfolio_parameters(file_path = fs::path(par_file_path, paste0(portfolio_na
 
 set_project_parameters(file.path(working_location, "parameter_files",paste0("ProjectParameters_", project_code, ".yml")))
 
-data_location_ext <- sub("[0-9]{4}Q[0-9]", financial_timestamp, data_location_ext)
+data_location_ext <- sub("20[0-9]{2}Q[1-4]", financial_timestamp, data_location_ext)
 
 if(project_code == "PA2020FL"){
   peer_group = case_when(


0_web_functions.R
```diff
@@ -85,7 +85,7 @@ set_web_parameters <- function(file_path) {
   twodii_internal <<- cfg$parameters$twodii_internal
   new_data <<- cfg$parameters$new_data
 
-  financial_timestamp <<- cfg$parameters$financial_timestamp
+  # financial_timestamp <<- cfg$parameters$financial_timestamp
```
MirijaHa, line 88: I have a general question,
I often see parts of code commented out instead of deleted and do not entirely understand the reasoning behind it. For me commented out code always signals that this is something that will become important again very soon and that one still somehow needs to have in the back of one`s mind. Is that the case here? :)

cjyetman: see my original explanation above: "These code changes should probably be commented, and the commented code deleted, but for now I left it like this or brevity."

cjyetman: i.e. I will not merge it like that... I just thought it was easier for people to review and give feedback on the change if it's super clear what has changed (depending on how they view it)

MirijaHa: ok, thx for the information

web_tool_script_1.R
```diff
@@ -23,6 +23,9 @@ set_portfolio_parameters(file_path = fs::path(par_file_path, paste0(portfolio_na
 
 set_project_parameters(file.path(working_location, "parameter_files",paste0("ProjectParameters_", project_code, ".yml")))
 
+data_location_ext <- sub("[0-9]{4}Q[0-9]", financial_timestamp, data_location_ext)
```
AlexAxthelm, line 26:
```suggestion
data_location_ext <- sub("20[0-9]{2}Q[1-4]", financial_timestamp, data_location_ext)
```
I don't want to deal with false positives, and if we ever deal with data from before 2000, or (somehow) this is still running a millennium from now, we can change it then

web_tool_script_2.R
```diff
@@ -24,6 +24,8 @@ set_portfolio_parameters(file_path = fs::path(par_file_path, paste0(portfolio_na
 
 set_project_parameters(file.path(working_location, "parameter_files", paste0("ProjectParameters_", project_code, ".yml")))
 
+data_location_ext <- sub("[0-9]{4}Q[0-9]", financial_timestamp, data_location_ext)
```
AlexAxthelm, line 27:
```suggestion
data_location_ext <- sub("20[0-9]{2}Q[1-4]", financial_timestamp, data_location_ext)
```

see comment in `web_tool_script_1`

web_tool_script_3.R
```diff
@@ -17,6 +17,8 @@ set_portfolio_parameters(file_path = fs::path(par_file_path, paste0(portfolio_na
 
 set_project_parameters(file.path(working_location, "parameter_files",paste0("ProjectParameters_", project_code, ".yml")))
 
+data_location_ext <- sub("[0-9]{4}Q[0-9]", financial_timestamp, data_location_ext)
```
AlexAxthelm, line 20:
```suggestion
data_location_ext <- sub("[0-9]{4}Q[0-9]", financial_timestamp, data_location_ext)
```

Same as above

AlexAxthelm (commented):
1. what does `ext` in `data_location_ext` signify? Why isn't is just `data_location`?
2. what if we didn't use a `sub`, but instead something along the lines of:
```r
# I don't know if exists() is the right function to check, maybe isnull instead :shrug:
if (exists(cfg$paths$data_location_ext)) {
  data_location_ext <<- cfg$paths$data_location_ext
} else {
  data_location_ext <<- file.path(cfg$paths$pacta_data, cfg$parameters$financial_timestamp))
}
```

We would need to put up a lot more guards around the filepath (case sensitivity, existence checks, etc.), but the general idea is that we don't need to change the old params files, but moving forward, we also don't need to keep the structure just because it's always been that way. Instead, we can define a path to the directory with all the quarter datasets, and let the timestamp make a decision (and not lean on treating filepaths as strings)

3. A more aggressive idea would be to take the above mechanism, but _do_ go back and change older param files. It seems like it would be a straightforward substitution that we could have Constructiva run all at once, and then we don't have to worry about varying parameter formats.

cjyetman:
> 1. what does `ext` in `data_location_ext` signify? Why isn't is just `data_location`?

don't know


> 2. what if we didn't use a `sub`, but instead something along the lines of:
> 
> ```r
> # I don't know if exists() is the right function to check, maybe isnull instead :shrug:
> if (exists(cfg$paths$data_location_ext)) {
>   data_location_ext <<- cfg$paths$data_location_ext
> } else {
>   data_location_ext <<- file.path(cfg$paths$pacta_data, cfg$parameters$financial_timestamp))
> }
> ```

in the code and parameter files' current state, `exists(cfg$paths$data_location_ext)` will always be `TRUE`

web_tool_script_2.R
cjyetman: fair enough 😄

web_tool_script_3.R
cjyetman: your suggestion didn't come through here?

cjyetman:
> > 1. what if we didn't use a `sub`, but instead something along the lines of:
> > 
> > ```r
> > # I don't know if exists() is the right function to check, maybe isnull instead :shrug:
> > if (exists(cfg$paths$data_location_ext)) {
> >   data_location_ext <<- cfg$paths$data_location_ext
> > } else {
> >   data_location_ext <<- file.path(cfg$paths$pacta_data, cfg$parameters$financial_timestamp))
> > }
> > ```
> 
> in the code and parameter files' current state, `exists(cfg$paths$data_location_ext)` will always be `TRUE`

actually, that's not right.... it's even worse...

`cfg` never exists in the global environment, it is created in sub-functions (or sub-sub-sub-functions), and the its values are assigned to objects in the global environment à la `data_location_ext <<- cfg$paths$data_location_ext` 😱 

https://github.com/2DegreesInvesting/PACTA_analysis/blob/897433f71c18edb6ba2cf6798f4c961158f27b0d/0_web_functions.R#L74-L89

so the more correct answer is: `exists(cfg$paths$data_location_ext)` will always be `FALSE`, but `exists(data_location_ext)` (probably what you intended) will always be `TRUE`

AlexAxthelm:
> actually, that's not right.... it's even worse...
> cfg never exists in the global environment, it is created in sub-functions (or sub-sub-sub-functions), and the its values are assigned to objects in the global environment à la data_location_ext <<- cfg$paths$data_location_ext 😱

I came to the same conclusion in parallel. 🤦 

My goal isn't to check if `data_location_ext` exists, but to _always_ create it, just in a meaningful way. I think the code block that I'm looking for (now that I've actually played with `config` a bit more) is something close to:

```r
cfg <- config::get(file = file_path)

if (is.null(cfg$paths$data_location_ext)) {
  data_location_ext <<- cfg$paths$data_location_ext
} else {
  data_location_ext <<- file.path(cfg$paths$pacta_data, cfg$parameters$financial_timestamp))
}
```
Where we check if the parameter is in the config using `is.null` rather than `exists`

There is a separate issue in which we need to be careful with naming our config parameters, since `config::get()` does partial name matching.

Example config.yml (from the documentation):
```yml
#config.yml
default:
  trials: 5
  dataset: "data-sampled.csv"
```
```r
R> cfg <- config::get()
R> cfg$trials
[1] 5
R> cfg$dataset
[1] "data-sampled.csv"
R> # This is the problem
R> cfg$data
[1] "data-sampled.csv"
R> cfg$d
[1] "data-sampled.csv"
R> cfg$x
NULL
```

cjyetman:
> ```r
> cfg <- config::get(file = file_path)
> 
> if (is.null(cfg$paths$data_location_ext)) {
>   data_location_ext <<- cfg$paths$data_location_ext
> } else {
>   data_location_ext <<- file.path(cfg$paths$pacta_data, cfg$parameters$financial_timestamp))
> }
> ```

don't think this would work... if the config file does not set `data_location_ext`, then `is.null(cfg$paths$data_location_ext)` will be `TRUE` and `data_location_ext` in the global environment will be set to `NULL`...  if `data_location_ext` is set in the config file, then it will set `data_location_ext` in the global environment based on some other parameters in the config file, which may or may not exist? `pacta_data` or instance does not exist in any current parameter files

oops.... I think you missed a `!`? Now I understand better, but....

 if the config file does set `data_location_ext`, then `!is.null(cfg$paths$data_location_ext)` will be `FALSE` and `data_location_ext` in the global environment will be set to whatever the config file says (that will always be the case when `set_web_parameters()` runs and will always be `../pacta-data/2019Q4/`)...  then when `set_project_parameters()` runs,  `cfg$paths$data_location_ext` will be `NULL` (for all existing project parameter files) , do it will set `data_location_ext` to `file.path(cfg$paths$pacta_data, cfg$parameters$financial_timestamp))`, which is achievable for new project parameter files by adding a `pact_data` key, but requires changes to all existing parameter files. unless we add something like...

```r
cfg <- config::get(file = file_path)

if (!is.null(cfg$paths$data_location_ext)) {
    data_location_ext <<- cfg$paths$data_location_ext
} else if (!is.null(cfg$paths$pacta_data)) {
    data_location_ext <<- file.path(cfg$paths$pacta_data, cfg$parameters$financial_timestamp))
}
```
then the old config files would work too? maybe?


> There is a separate issue in which we need to be careful with naming our config parameters, since `config::get()` does partial name matching.

yeah.... I hate these yml files

cjyetman:
this also might present a problem down the road if the user parameters and the project parameters clash

web_tool_script_3.R
cjyetman: I changed it equivalently

maurolepore (approved):
LGTM.

I let you and others with more context review the "why" aspect of this. 

RE the "how", my only thought is to maybe extract a lil wrapper named to clarify what's going on here. For example:

```r
# some/where.R
replace_quarter <- function(path, timestamp) { ... }

# web_tool_script_{1-3}.R
data_location_ext %>% replace_quarter(timestamp = financial_timestamp)
```

This clarifies that 1) the main input is a path, 2) the main action is a replacement, 2) and the main thing being replaced is a quarter timestamp.

cjyetman:
I think maybe I'd like to change the strategy.

I don't know the history or even intended purpose of these files:
[parameter_files/WebParameters_docker.yml](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/parameter_files/WebParameters_docker.yml) or [parameter_files/WebParameters_2dii.yml](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/parameter_files/WebParameters_2dii.yml)

I'd like to consider those "default settings", but open up the idea of overriding *any* of those settings in the project or user parameter files (which are already loaded after the web parameters, except in stress testing where we'll need to make sure that happens). Then if we want to change the `data_location_ext`, it will be clear that we can/should add something like this to the *project* parameters files (rather than feel like we have to let that be set in the web parameters and at most modify it with other keys/values from other parameters):
```yaml
default:
    paths:
        data_location_ext: ../pacta-data/2019Q4/
```

Any existing project parameters that correctly want to use 2019Q4 data can stay unchanged and they will simply benefit from the 2019Q4 data location being set by default. Any new or existing project parameter files that actually want to use 2020Q4 data will need to have the `data_location_ext: ../pacta-data/2020Q4/`.

For clarity, this PR will be closed, and a new one will be created that will touch a bunch of the existing parameter files, but ideally nothing else (I think) (and also point to the stress testing code that will need to change).

Does that make sense @AlexAxthelm ?

cjyetman:
superseded by #483

PR #483, opened 2021-09-09
enable setting/overriding data_location_ext in the project parameters

cjyetman:
supersedes #482

enable setting/overriding data_location_ext in the project parameters

in the future, when needing to set a project to use any dataset other than the default `data_location_ext: ../pacta-data/2019Q4/`, the "*project* parameters" file should have something like...
```yaml
        paths:
                data_location_ext: ../pacta-data/2020Q4/
```

because the "*project* parameters" file is always loaded *after* the default "web parameters" file, the `data_location_ext` value in the "*project* parameters" file will override the value from the "web parameters" file (the `setup_project()` function loads the default "web parameters" file, and the `set_project_parameters()` function loads the "project parameters" file).
https://github.com/2DegreesInvesting/PACTA_analysis/blob/897433f71c18edb6ba2cf6798f4c961158f27b0d/web_tool_script_1.R#L16-L24
https://github.com/2DegreesInvesting/PACTA_analysis/blob/897433f71c18edb6ba2cf6798f4c961158f27b0d/web_tool_script_2.R#L17-L25
https://github.com/2DegreesInvesting/PACTA_analysis/blob/897433f71c18edb6ba2cf6798f4c961158f27b0d/web_tool_script_3.R#L12-L18

This is also currently the case with both web tool scripts in stress testing...
https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/2f80bdd3875c0abf1bd1f9fe6b4d9de12c0d4dc5/web_tool_external_stress_test.R#L31-L48
https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/2f80bdd3875c0abf1bd1f9fe6b4d9de12c0d4dc5/web_tool_stress_test.R#L56-L79

One long term consideration is that the stress testing tool/s do not currently load the "*portfolio* parameters" files, so if we ever enable some cascading of preferences between default, project and portfolio, then we may need to add that to stress testing, and also review the order in which we load them to enable the proper cascading precedence.

### DIFF
--- a/0_global_functions.R
+++ b/0_global_functions.R
@@ -46,6 +46,10 @@ set_col_types <- function(grouping_variables, fixed_col_types) {
 set_project_parameters <- function(file_path){
   cfg <- config::get(file = file_path)
 
+  if (!is.null(cfg$paths$data_location_ext)) {
+    data_location_ext <<- cfg$paths$data_location_ext
+  }
+
   project_report_name <<- cfg$reporting$project_report_name
   display_currency <<- cfg$reporting$display_currency
   currency_exchange_value <<- as.numeric(cfg$reporting$currency_exchange_value)
--- a/parameter_files/ProjectParameters_PA2021NO.yml
+++ b/parameter_files/ProjectParameters_PA2021NO.yml
@@ -1,14 +1,17 @@
 default:
 
+    paths:
+        data_location_ext: ../pacta-data/2020Q4/
+
     reporting:
         project_report_name: norway
         display_currency: NOK
         currency_exchange_value: 0.11
 
     parameters:
-        timestamp: 2019Q4
-        dataprep_timestamp: 2019Q4_250220
-        start_year: 2020
+        timestamp: 2020Q4
+        dataprep_timestamp: 2020Q4_transitionmonitor
+        start_year: 2021
         horizon_year: 5
         select_scenario: WEO2019_SDS
         scenario_auto: ETP2017_B2DS


AlexAxthelm:
I think this is a good start, but I think we need to have a broader conversation about configurations (and config files) in general.

What I think would be a good structure to start moving towards might involve moving away from the `config` package (since it expexts a single config file), and instead have ordered inheritance/overrides (like most systems that use YAML, such as docker-compose files).

What I'm thinking is that the order would be:

1. Default config
2. Project Config
3. User Config
4. Portfolio Config

by bringing all these together, we could define a whole config that gets more and more tailored to the particular PACTA run as we add more information. Not all files are necessary. So for example, an analyst running locally could just put all the parameters into a single portfolio file and not worry about project or user configs. 

This would also help modularize the PACTA ecology. Module (thinking Stress Testing and Interactive reports, basically anything that is not "core" PACTA) authors can define their own default params as part of the module, and then during the run we can pass the overall config to the module, and use the same mechanisms to handle overrides there.

```yaml
#incomplete user params
user:
  given_name: Alex
  family_name: Axthelm
  name_order: FL
  organization: 2 Degrees Investing Initiative

modules:
  stress_testing: TRUE
  interactive_report: TRUE
  pdf_report: FALSE
  market_report: FALSE
  
pacta:
  timestamp: 2020q4
  start_year: 2023

# This section is optional, and is ignored by everything that isn't stress testing
stress_testing:
  foo: 4
  bar: baz
```

cjyetman:
> What I think would be a good structure to start moving towards might involve moving away from the `config` package (since it expexts a single config file), and instead have ordered inheritance/overrides (like most systems that use YAML, such as docker-compose files).

[config pkg does do inheritance](https://rstudio.github.io/config/articles/introduction.html#defaults-and-inheritance), we just (typically) don't really use it correctly

but this is more about maintaining backward compatibility and keeping transitionmonitor running smoothly while enabling critical features on a short timeframe.... I'm not super comfortable rolling out big changes like that by the time 2020Q4 needs to be available.... though I'd be happy to let someone else take over responsibility for that 😄

AlexAxthelm:
> I'm not super comfortable rolling out big changes like that by the time 2020Q4 needs to be available

completely agree. That's a longer-term goal

> config pkg does do inheritance

I think the difference between what `config` does and what I'm imagining is for `config::get()`, everything has to be in a single file, where I'm imagining something that could be call upon multiple files in an ordered vector.
```r 
cfg <- get_config(c("portfolio.yml", "project.yml", "user.yml", "porfolio.yml"))
```

cjyetman:
I think this would achieve the same thing, no?

```r
cfg <- get_config("portfolio.yml")
cfg <- get_config("project.yml")
cfg <- get_config("user.yml")
```

I'm finally realizing/understand that's maybe what was intended with the web tool scripts, but it wasn't clear because they set the parameters through a sub function.

AlexAxthelm:
> I think this would achieve the same thing, no?
> 
> ```r
> cfg <- get_config("portfolio.yml")
> cfg <- get_config("project.yml")
> cfg <- get_config("user.yml")
> ```

I would expect that to overwrite, not integrate.  So for example:

```yaml
# portfolio
foo:
  a: 1
  b: 2
```
```yaml
# project
foo:
  a:  10
  c: 3
```

If I read in both files and integrated them, I would want to see something that would be the same as if I had read in
```yaml
foo:
  a: 10
  b: 2
  c: 3
```

cjyetman:
``` r
writeLines("default:\n  foo:\n    a: 1\n    b: 2", "portfolio.yml")
writeLines("default:\n  foo:\n    a: 10\n    c: 3", "project.yml")

config::merge(
  config::get(file = "portfolio.yml"),
  config::get(file = "project.yml")
)
#> $foo
#> $foo$b
#> [1] 2
#> 
#> $foo$a
#> [1] 10
#> 
#> $foo$c
#> [1] 3
#> 
#> 
#> attr(,"config")
#> [1] "default"
#> attr(,"file")
#> [1] "/private/var/folders/s6/n7932s8j3bjgxnlshctm14q40000gn/T/RtmpuVD3Mt/reprex-5a1501914bf-plump-imago/portfolio.yml"
```

AlexAxthelm:
Thanks @cjyetman that looks exactly like what I had been hoping for.

I think a nice little wrapper function that accepts a list of files, and then recursively merges them would solve a lot of the issues that we discussed recently. 

cc @MirijaHa

cjyetman:
With the [latest change in stress testing](https://github.com/2DegreesInvesting/r2dii.climate.stress.test/pull/47) in place, I've now verified that this PR seems to work as expected. I believe this is the lowest impact, least risk solution to letting the projects set the path to the appropriate data for the specified quarter. @AlexAxthelm can you approve?

AlexAxthelm (approved):
Thanks @cjyetman.

Issue #484, opened 2021-09-14
don't use all_of() in filter()


cjyetman:
https://github.com/2DegreesInvesting/PACTA_analysis/blob/897433f71c18edb6ba2cf6798f4c961158f27b0d/0_web_functions.R#L51

This appears to be an "off-label" usage of [`all_of()`](url), and frankly, I find it very difficult to predict its behavior, even though apparently it seems to work the same as the more standard use of `filter()`, even when abused...
``` r
library(dplyr)
getNamespaceVersion("dplyr")
#> version 
#> "1.0.7"
getNamespaceVersion("tidyselect")
#> version 
#> "1.1.1"

df <- tibble(x = LETTERS[1:5])
df
#> # A tibble: 5 × 1
#>   x    
#>   <chr>
#> 1 A    
#> 2 B    
#> 3 C    
#> 4 D    
#> 5 E

my_filter <- "A"
df %>% filter(x == my_filter)
#> # A tibble: 1 × 1
#>   x    
#>   <chr>
#> 1 A
df %>% filter(x == all_of(my_filter))
#> # A tibble: 1 × 1
#>   x    
#>   <chr>
#> 1 A

my_filter <- c("A", "B")
df %>% filter(x == my_filter)
#> Warning in x == my_filter: longer object length is not a multiple of shorter
#> object length
#> # A tibble: 2 × 1
#>   x    
#>   <chr>
#> 1 A    
#> 2 B
df %>% filter(x == all_of(my_filter))
#> Warning in x == all_of(my_filter): longer object length is not a multiple of
#> shorter object length
#> # A tibble: 2 × 1
#>   x    
#>   <chr>
#> 1 A    
#> 2 B

my_filter <- NULL
df %>% filter(x == my_filter)
#> Error: Problem with `filter()` input `..1`.
#> ℹ Input `..1` is `x == my_filter`.
#> x Input `..1` must be of size 5 or 1, not size 0.
df %>% filter(x == all_of(my_filter))
#> Error: Problem with `filter()` input `..1`.
#> ℹ Input `..1` is `x == all_of(my_filter)`.
#> x Input `..1` must be of size 5 or 1, not size 0.

my_filter <- c("A", "B")
df %>% filter(x %in% my_filter)
#> # A tibble: 2 × 1
#>   x    
#>   <chr>
#> 1 A    
#> 2 B
df %>% filter(x == all_of(my_filter))
[...]

PR #486, opened 2021-09-15
removes off-label use of all_of()

cjyetman:
- removes off-label use of `dplyr::all_of()` fixing issue #484 
- uses the `.env$` [syntax which is more explicit](https://rlang.r-lib.org/reference/tidyeval-data.html) and avoids the need for the appended "_" on the object name


maurolepore (approved):
LGTM. Great use of the `.env$` pronound :+1:

PR #489, opened 2021-09-23
DO NOT MERGE, just demonstration purposes

jdhoffa:


Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM rocker/r-ver:latest
+
+RUN Rscript -e 'install.packages("remotes")'
+
+COPY DESCRIPTION /bound/DESCRIPTION
+RUN Rscript -e 'remotes::install_deps("/bound", dependencies = TRUE)'
+
+COPY . /bound
+RUN chmod -R +x /bound
```
jdhoffa, line 9: this is just a brute force way to get things working, not ideal

R/run_pacta.R
```diff
@@ -0,0 +1,28 @@
...
+    system(command[[i]])
+    message("End portfolio: ", portfolios[[i]])
+  }
+}
+
+#' Get the name of the portfolios
+portfolios <- function(path = "../input", regexp = "_Input[.]csv") {
+  csv <- fs::dir_ls(path, regexp = regexp)
+  fs::path_ext_remove(fs::path_file(csv))
+}
```
jdhoffa, line 27: In principle, I'm not a fan of having an R function that calls bash commands. This was implemented just because it makes looping over all input portfolios quite easy, but we could also just write a bash loop that does this.

docker-compose.override.yml
```diff
@@ -0,0 +1,9 @@
+version: "3.2"
+services:
+  app: 
+    volumes: 
+      - /Users/jdhoffa/Documents/github/2dii/pacta-data:/pacta-data:ro
+      - /Users/jdhoffa/Desktop/pacta/input:/input:ro
+      - /Users/jdhoffa/Desktop/pacta/output:/output
```
jdhoffa, line 7: obviously this will need to change, @AlexAxthelm would love comments

cjyetman: relative local paths?

cjyetman: ☝🏻 that was a suggestion

R/run_pacta.R
cjyetman: maybe use [withr](https://withr.r-lib.org)?

cjyetman: also note that this commit 897433f71c18edb6ba2cf6798f4c961158f27b0d is super relevant here because it moves the stress testing to web_tool_script_3.R

jdhoffa: yup that's what made this possible :)

docker-compose.override.yml
jdhoffa: I'd prefer to allow flexibility to the user to specify data/ input and output locations when they run `docker-compose`. Relative paths still bake in a hard dependency (that the user should have input, output and data located somewhere in the neighborhood of `PACTA_analysis`). 

Mauro had the idea to specify these in a .env file, and then run something along the lines of: 
``` bash
source .env
docker-compose up
```

but then you would need to specify, in the docker-compose file
``` yml
...
volumes: 
   - ${PACTA_DATA}:/pacta-data:ro
```

and that could also be a bit wonky.

jdhoffa: Allow me to elaborate why I don't want to hard-code a relative dependency. 
Let's say the user clones `PACTA_analysis`, this should be the "tool", similar to Chrome, or any other application, that allows them to "run" PACTA. They should not need to "change" anything in that tool to run it.

Imagine they also clone `pacta-data` (in this example, they have access). This should also live in some folder, and be untouched. 

Now, perhaps they might want to run PACTA on multiple portfolios, in multiple different directories. They should be able to do so, by calling `docker-compose` and specifying the path to the "input" and "output" directories of their choosing (and they should be able to run this over and over again as many times as they wish). Even the ability to specify different data paths is helpful (especially to us) if we want to point to staging data or other versions of the dataset. 

If the input/ output needs to be somewhere relative to `PACTA_analysis`, then they would need to copy their data into that folder whenever they want to run the tool, which seems like an awkward UI.

R/run_pacta.R
jdhoffa: Note to self: `R/run_pacta.R` and `R/source_web_tool_scripts.R` are basically the same, likely they can be combined into one function.

docker-compose.override.yml
cjyetman: So... I was not suggesting a hard dependency on relative paths, only that it is one possibility (so that example code in the repo doesn't need to have usernames in it, and at least hypothetically could work on a user's machine)

jdhoffa: Yup I understand, it was how we had initially solved this problem in `pactaCore`, but it comes with it's own host of issues (that I sorta mentioned above). 

I left my username in as an obvious flag that "this is not how this should be done and needs to be fixed". It all relates to a convo @AlexAxthelm had been having on slack on alternatives. I opened this PR exactly to demonstrate what I'm doing and try to get some feedback ;-) lol. But apologies as he has much more context that I didn't explicitly state in this PR

bin/run-pacta
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+here=$(pwd)
+pacta=${1:-$here}
+parent=$(dirname $pacta)
+
+cp -r "$parent"/input/*.yml "$pacta"/working_dir/10_Parameter_File
+cp -r "$parent"/input/*.csv "$pacta"/working_dir/20_Raw_Inputs
```
jdhoffa, line 8: Note to slef: Since we are just dumping whatever input into the `working_dir/20_Raw_Inputs` folder, and this folder, by default, already contains `TestPortfolio_Input.csv`, the output of running this command will always contain a `TestPortfolio_Input` output.

jdhoffa: Slef!? Who's slef?!!

jdhoffa: I am very funny.

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM rocker/r-ver:latest
```
AlexAxthelm, line 1: pin version. 
```suggestion
FROM rocker/r-ver:4.1.1
```

Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+
+RUN Rscript -e 'install.packages("remotes")'
+
+COPY DESCRIPTION /bound/DESCRIPTION
```
AlexAxthelm, line 5: NB: I would choose a different directory name than `bound`. That's not  very descriptive. Perhaps `pacta_analysis`?

Dockerfile
AlexAxthelm: not ideal, but this works for me.

R/run_pacta.R
```diff
@@ -0,0 +1,28 @@
...
+run_pacta <- function(wd = here::here()) {
+  withr::local_dir(wd)
+
+  portfolios <- portfolios()
+  command <- glue::glue("Rscript --vanilla web_tool_script_1.R {portfolios};\
+                          Rscript --vanilla web_tool_script_2.R {portfolios}")
+
+  for (i in seq_along(command)) {
+    message("Start portfolio: ", portfolios[[i]])
+    system(command[[i]])
```
AlexAxthelm, line 18: generally prefer `system2` over `system`, but only if that's actually the way to go.  https://ropensci.org/blog/2021/09/13/system-calls-r-package/

R/run_pacta.R
AlexAxthelm: I'm glad you recognize that this is a  problematic construction. Note that your current implementation starts an R session from within an R session. 🤯 

Instead, I would probably do this as a bash script, and have that the the CMD in the dockerfile.

Dockerfile
AlexAxthelm: Am I missing something, or does this dockerfile not have a CMD or ENTRYPOINT?

bin/run-pacta
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
AlexAxthelm, line 1: When working in Docker, I prefer that scripts are `sh` compliant, rather than assuming that `bash` is available. in particular, `alpine` doesn't  have `bash` (instead it has `ash`, IIRC)

bin/run-pacta
```diff
@@ -0,0 +1,11 @@
...
+
+here=$(pwd)
```
AlexAxthelm, line 3: I see what's going on, but I don't like it. I'd prefer to have a shell function or script that reads some configurations (like where pacta data is) from envvars, and then calls R on the correct file, regardless of where the user currently is (`pwd`)

something that could be invoked along the lines of 
```bash
cd ~/Downloads
export PACTA_DATA=/tmp/PD202q4
run_pacta ~/portfolios/foo.csv
```

docker-compose.override.yml
AlexAxthelm: I would change this file to `docker-compose.override.example.yml`, and call it good. then include instructions in the readme on how to copy it, and set your own path correctly.

If you do that, I'd also switch these volume mounts to be in long syntax, since  that will be easier for folks to change correctly (don't have to worry about stopping at `:`)

AlexAxthelm (changes requested):
A lot of good effort here, but there are some things (especially around invocation) that I see would cause issues for users.

Dockerfile
jdhoffa: Unfortunately, there are some hard-coded expectations in `PACTA_analysis`, that make it run a certain way if it detects that it is being run in a folder called `bound`. (This is to "detect that it is in transition monitor" or something like that). The least intrusive way around that is to follow the same protocol that Transition Monitor does, to ensure the tool runs the same in both environments.

cjyetman:
💯 agree that the existing README sucks. However, changing it to something else that sucks and/or doesn't even remotely capture what some other parts of this repo are for and are currently actively being maintained and worked on doesn't seem like a good idea.... especially not sneaking it in with a PR intended to add some new feature.

cjyetman:
I guess maybe I would expect a PR to modify a README if necessary, maybe add a new section to it... I would not expect it to completely change it.

regardless of whether this PR merges or not... this repo is still used for the transitionmonitor purposes... and that code that's not being actively maintained is still used by some, even if we believe it's inadvisable.... and I suspect some will continue to use the code in this repo outside of a Docker image (I definitely will)

jdhoffa:
Also my plan initially was to add a section to the `README`, but when I read through it, it didn't seem like it was worth keeping much of what was already there.

cjyetman:
ok... reasonable, healthy suggestion time... can we rather create a new PR where we exclusively make and discuss a complete overhaul of the README, give a variety of developers/users/stakeholders an opportunity to participate and comment etc.?

jdhoffa:
Sure, I mean I would imagine we can start that effort separately. 

In my opinion, the changes to the README here: 
- don't remove anything that works/ is useful (maybe the system requirements bit, but even that seems a bit strange to me)
- don't introduce anything harmful
- only introduce improvements

So in general, I would prefer to leave the changes as is in this particular PR, and then start a separate effort to expand the documentation to touch the other useful parts of the repo. 

It's sorta the same approach I take with refactoring, like obviously a full concerted refactor of the whole repository would be nice, but if a change I introduce involves touching/ refactoring certain parts of the code around it, and it "doesn't harm anything, and only improves" then I like to take the chance to gain the improvement quickly, rather than waiting for a large-scale effort that might take a long time/ get deprioritized/ not actually happen. 

But ya, that's just my two-cents/ general development approach, if you're very against that, or have extra time to do a rewrite of all possible use-cases of the repo/ get input from all relevant stakeholders, then I'm happy to help out with that effort as much as I can (though I also don't really have a lot of spare time).

cjyetman:
As a reviewer, I take it as my responsibility to genuinely review *all* of the changes in the PR, so if you remove things from the top level README that are not related to this PR, then you...
1. create more work for me as reviewer that's unnecessary for this PR
2. cause me as a reviewer to completely switch contexts as I'm reviewing this PR (thinking about how the new code works here feels very different to me than thinking about whether or not a paragraph Andy added to the README 9 months ago is important to someone from one of the many perspectives that people come to PACTA from)

I would suggest not doing that.

AlexAxthelm:
I agree. README is a *documentation* file, which means that changes to it should be _documenting_ any changes made in the PR. Broader restructuring/cleaning of README should be a separate PR.

PR #492, opened 2021-09-24
update GH actions to use ubuntu-latest

cjyetman:
- update GH actions to use ubuntu-latest
- use more recent, lighter weight R CMD check action "check-standard.yaml"
- add a fix to encoding checker to ignore files in the PR that have been removed


AlexAxthelm (commented):
I'll let @maurolepore make the actual approval, but I don't see any issues with this.

maurolepore (approved):
LGTM

PR #493, opened 2021-09-27
retain `scenario_source` when aggregating portfolio and company results

cjyetman:
The `scenario_source` column is advantageous to have when there are scenarios with the same name but the source is different. Leaving this column in the results is advantageous especially for stress testing (@jacobvjk maybe you can pull this branch to test things on your end?).

For now this is a draft until we are more certain that this will not have a negative impact, but I'm pretty sure this will be a good thing.

### DIFF
--- a/0_portfolio_test.R
+++ b/0_portfolio_test.R
@@ -125,7 +125,7 @@ aggregate_company <- function(df) {
     df <- df %>%
       ungroup() %>%
       select(
-        all_of(grouping_variables), scenario, allocation,
+        all_of(grouping_variables), scenario_source, scenario, allocation,
         id, company_name, financial_sector, port_weight,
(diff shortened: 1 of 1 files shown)


jacobvjk:
thanks @cjyetman this should indeed be really helpful. I will check in with @MirijaHa tomorrow to review this thoroughly

cjyetman:
After some thought about this, I honestly don't know **how** I/we will get "more certain that this will not have a negative impact", so.... I'm thinking this should just be reviewed by a few people who have enough context to have some intuition about whether this is "safe" or not. @AlexAxthelm @jacobvjk @jdhoffa ?

jacobvjk:
@cjyetman : we thought about this in the ST side, and would go so far to say that we need this column. so will approve

jacobvjk (approved):
lgtm

PR #496, opened 2021-10-06
1923 fix pacta supervisors prep

jacobvjk:
closes ADO 1923: https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/1923

This PR:
- fixes the data preparation in the pacta supervisors workflow to ensure no holdings that cannot be processed appropriately get into the eq and cb portfolios
- it makes reading in the raw portfolio more robust


jdhoffa (approved):
LGTM and doesn't touch any not-supervisor-workflow code so I think you can merge.

PR #499, opened 2021-10-08
filter out short positions immediately on portfolio import

### LINKED ISSUE #463: PACTA erroneously incuding short positions
After running PACTA with the current code I found that some results on portfolio level showed negative values in `plan_alloc_wt_tech_prod` & `plan_emission_factor`. As I am using port_weight, this can only happen (in theory) if the underlying ALD has negative production values (rather unrealistic but you never know) or that we are including short positions (negative values for `market_value` / `value_usd`) in the analysis. I think the latter is the case. 

This would be a bug in my opinion, as we dont want to include short positions as this would distort the analysis. 

Here is a reprex comparing results from two projects, one of which used an older version (from early autumn last year) of the code and one the current version

```r
library(r2dii.utils)
library(dplyr)

# results with old pacta code
bonds_results_old <- read_rds(
  fs::path(path_dropbox_2dii(),"/PortCheck_v2/10_Projects/mfm_v7/40_Results/Bonds_results_company.rda")
) %>% 
  filter(port_weight < 0) %>% 
  filter(year == 2020, scenario == "B2DS")
  
equity_results_old <- read_rds(
  fs::path(path_dropbox_2dii(),"/PortCheck_v2/10_Projects/mfm_v7/40_Results/Equity_results_company.rda")
) %>% 
  filter(port_weight < 0) %>% 
  filter(year == 2020, scenario == "B2DS")

bonds_results_old %>% 
  as_tibble() %>% 
  select(portfolio_name, company_name, port_weight, allocation_weight, ald_sector, technology ,plan_tech_prod, plan_alloc_wt_tech_prod) %>% 
  distinct() 

equity_results_old  %>% 
  as_tibble() %>% 
  select(portfolio_name, company_name, port_weight, allocation_weight, ald_sector, technology ,plan_tech_prod, plan_alloc_wt_tech_prod) %>% 
  distinct()

# results with new pacta code
bonds_results_new <- read_rds(
  fs::path(path_dropbox_2dii(),"/PortCheck_v2/10_Projects/mfm_v8/40_Results/Bonds_results_company.rda")
) %>% 
  filter(port_weight < 0) %>% 
  filter(year == 2020, scenario == "B2DS")

equity_results_new <- read_rds(
  fs::path(path_dropbox_2dii(),"/PortCheck_v2/10_Projects/mfm_v8/40_Results/Equity_results_company.rda")
) %>% 
  filter(port_weight < 0) %>% 
  filter(year == 2020, scenario == "B2DS")

bonds_results_new %>% 
  as_tibble() %>% 
  select(portfolio_name, company_name, port_weight, allocation_weight, ald_sector, technology ,plan_tech_prod, plan_alloc_wt_tech_prod) %>% 
  distinct() 

equity_results_new  %>% 
  as_tibble() %>% 
  select(portfolio_name, company_name, port_weight, allocation_weight, ald_sector, technology ,plan_tech_prod, plan_alloc_wt_tech_prod) %>% 
  distinct()
```

When comparing the two versions of the code (I still have a copy of it) I think I found the commit since when we are analysing short positions: https://github.com/2DegreesInvesting/PACTA_analysis/commit/7e125aad61baafd39bd3d7e7d608d400694fe4f7

More specific, this line was deleted in `3_run_analysis.R `: `port_eq <- port_eq %>% filter(port_weight > 1e-6)`

I dont know the motivation why we had this filter in the first place, but it lead to the fact that we kicked out short positions from `port_eq`, as negative values are below ` 1e-6`. Without this filter we include short positions since this commit.

This raises a couple of questions from my end:

- Do we intend kickout short positions at a different part of the code?
- If not, how should we kickout short positions in the investor code?
- Why do we only excluded short position in `port_eq` and not also in `port_cb`? It is possible to have negative bonds as well.

cjyetman:
closes #463

### DIFF
--- a/0_web_functions.R
+++ b/0_web_functions.R
@@ -194,6 +194,9 @@ get_input_files <- function(portfolio_name_ref_all) {
     stop("Multiple investors detected. Only one investor at a time can be anaylsed")
   }
 
+  # remove holdings with <= 0 market value (short positions)
+  portfolio <- portfolio %>% filter(market_value > 0)
+
   return(portfolio)
 }
 


cjyetman:
@FrederickFa maybe something equivalent needs to happen in here?https://github.com/2DegreesInvesting/PACTA_analysis/blob/abf4ef9acd22347994e193c8fef8baf537a4398d/0_portfolio_input_check_functions.R#L14-L48

FrederickFa (changes requested):
@cjyetman thanks picking this up so quickly!

I am not familiar with the web tool scripts, but in my opinion a change is needed in both the web tool script and the offline workflow. 

However, in my opinion we shouldn't kick out short positions completely from PACTA, but only make sure they are not included in the analysis. Hence, the way we did it earlier (as mentioned in #463 ) is the right place to me.

Why? Because otherwise we change the weights of positions in the portfolio. As short positions have negative values, filtering them would increase the portfolio size and thus also decrease the weight of single positions. This would change the sector exposure later on. 

However, I am not sure how to actually include short positions when calculating sector exposures (as negative exposures dont really make sense). I think this would be important to research and also maybe engage with people who are more familiar with those matters. For the moment, however, I would propose we interfere as little as possible with the weights of a portfolio, until we figure out what is the best way. Therefore, importing the portfolio as it is, but just making sure negative marker values are not included in the analysis.

cjyetman:
This is why I didn't do this earlier... it didn't fall off my radar, I just don't know the proper how/what/where.

FrederickFa:
I understand, but leaving it as it is now is in my opinion definitely worse than filtering them.

FrederickFa:
It is also interesting that this is not documented anywhere. At least I could not find any information about our communication on short positions either in the PACTA COP documentation or in the Toolbox.

0_web_functions.R
```diff
@@ -194,6 +194,9 @@ get_input_files <- function(portfolio_name_ref_all) {
     stop("Multiple investors detected. Only one investor at a time can be anaylsed")
   }
 
+  # remove holdings with <= 0 market value (short positions)
+  portfolio <- portfolio %>% filter(market_value > 0)
```
AlexAxthelm, line 198:
```suggestion
  portfolio <- portfolio %>% dplyr::filter(market_value > 0)
```

AlexAxthelm (approved):
This definitely needs to have a methodological review, and a more permanent decision, but this looks like a good fix for now.

FrederickFa:
> This definitely needs to have a methodological review, and a more permanent decision, but this looks like a good fix for now.

@AlexAxthelm So you see rather no risk with the points I mentioned when filtering at the very beginning?

FrederickFa:
> > So you see rather no risk with the points I mentioned when filtering at the very beginning?
> 
> No, the risks you brought up are still relevant. But in the absence of any better suggestions, I think we should go with the behavior you suggested in the [discussion on #463](https://github.com/2DegreesInvesting/PACTA_analysis/issues/463#issuecomment-857669907):

But the behavior I suggested in #463 is different from the proposed changes in this PR. 

The change in this PR kicks out all short positions completely from the entire PACTA, hence we interfere with the portfolio structure. 

My proposal is to only filter short positions within the analysis part, hence avoiding any weird negative production values. But the portfolio itself remains untouched.
 ______
> > I would also propose to only filter position above zero market value.
> 
> I am assuming that your position on the solution hasn't changed, and since you're flagging this as a major issue, we need to get a solution that is functional (if not perfect).
> 

As the proposed change in this PR is `market_value > 0 `, this would be fine from my side.
_______
> > Also, this PR doesn't solve the issue within the offline workflow.
> 
> It seems as though each analyst has their own different "offline" workflow at this point, so until we have consolidated central structure that we can get everyone to agree upon, I don't think we can support those beyond offering changes to `get_input_files()`. If someone is willing to bypass that function, then they are taking on the responsibility for maintaining their different workflow themselves.

Ok. So in case for MFM, we still use the workflow consisting of [1_portfolio_check_initialisation.R](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/1_portfolio_check_initialisation.R), [2_project_input_analysis.R](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/2_project_input_analysis.R) and [3_run_analysis.R](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/3_run_analysis.R), and update our repositories for new updates. I would have thought that this is the only offline workflow. 

Therefore, when going back to #463, I would have proposed a change to [3_run_analysis.R](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/3_run_analysis.R): to filter short positions from getting included in the analysis.

AlexAxthelm:
> But the behavior I suggested in #463 is different from the proposed changes in this PR.

I see the difference now. thank you for explaining. I'll be continuing this discussion on ADO. AB #661

cjyetman:
> My proposal is to only filter short positions within the analysis part, hence avoiding any weird negative production values. But the portfolio itself remains untouched.
 
@FrederickFa if you can be more precise about what/where this means, I'll take another stab at it. You should be aware though that this so-called "offline" version of PACTA has basically been abandoned and not maintained for many months now, so at some point its results may diverge from those that the web tool scripts produce.

To be honest, I'm not sure how to interpret all the comments in [PRODUCT BACKLOG ITEM 661](https://dev.azure.com/2DegreesInvesting/2DegreesInvesting/_workitems/edit/661/), so I'm not 100% sure there is agreement on this.

FrederickFa:
Thanks @cjyetman for picking this up!
_______
> To be honest, I'm not sure how to interpret all the comments in PRODUCT BACKLOG ITEM 661, so I'm not 100% sure there is agreement on this

I think there is still no agreement but rather a set of possible solutions. As leaving the current situation is not a good idea, I guess the best way in the short term is to be pragmatic and just filter out short position in the analysis parts. So, maybe add the following code after the mentioned lines:

- Filter out short positions in equity portfolio:
https://github.com/2DegreesInvesting/PACTA_analysis/blob/477da47bbc2c45730bea051f5850b39f1fb24e5b/3_run_analysis.R#L54
`port_eq <- port_eq %>% filter(port_weight > 0)`

- Filter out short positions in bonds portfolio:
https://github.com/2DegreesInvesting/PACTA_analysis/blob/477da47bbc2c45730bea051f5850b39f1fb24e5b/3_run_analysis.R#L154
`port_cb <- port_cb %>% filter(port_weight > 0)`

_______

> You should be aware though that this so-called "offline" version of PACTA has basically been abandoned and not maintained for many months now, so at some point its results may diverge from those that the web tool scripts produce.

I think this issue really needs to be discussed again because, as I have mentioned in the past, MFM relies on using the offline code (@looreen @Antoine-Lalechere ). But I can understand the reason for this, maintaining to different codes takes time, so do you see a possibility to run the web tool script offline to create similar results in the structure of the offline code? Or do you have a different idea to ensure consistency? I mean, for our credibility, PACTA results should not be different if an asset manager uses TM or sees his fund displayed on MFM.

cjyetman:
superseded by #533

cjyetman:
> I think this issue really needs to be discussed again because, as I have mentioned in the past, MFM relies on using the offline code (@looreen @Antoine-Lalechere ). But I can understand the reason for this, maintaining to different codes takes time, so do you see a possibility to run the web tool script offline to create similar results in the structure of the offline code? Or do you have a different idea to ensure consistency? I mean, for our credibility, PACTA results should not be different if an asset manager uses TM or sees his fund displayed on MFM.

I don't know precisely what you mean by "similar results", but the web tool scripts do produce PACTA results very much like the other version. The primary difference is that it's increasingly trying to do *just* PACTA well, but not other things that complicate the process, so for instance, if one wants to run multiple portfolios, than the expectation is that they run each portfolio through the PACTA process individually. That might sound daunting or inconvenient, but I believe that looping through the PACTA process multiple times is something that should be done outside of PACTA, not inside. I'd be happy to do a tutorial session with you to get you up to speed on that.

cjyetman:
I've personally walked through using the web tool scripts locally with @Antoine-Lalechere and I believe he is comfortable doing that on his own now, so maybe @Antoine-Lalechere can say whether that's adequate for MFM purposes since he has context from both sides?

Antoine-Lalechere:
The problem of the `web_tool_script`s is that it can only run for 1 portfolio at a time and give output file by portfolio, which is not something that can be used as such by MFM as we use to run 12000 different portfolios. I've talked with @AlexAxthelm who said that a workflow suiting MFM needs (basically dividing by portfolio, then executing just PACTA on a single portfolio, then binding results needed by MFM (40_Results folder output) can be provided by PACTA team. @FrederickFa @looreen @cjyetman

cjyetman:
Thanks @Antoine-Lalechere. That sounds like it will end up being something very similar to the "meta report data creator" you've used recently. Totally feasible, but does not yet exist.

jacobvjk:
this requirement (running Pacta offline for 12k portfolios) is also something that would be a big step towards a workflow suitable for supervisors. data sharing is still another issue then of course, but in principle, this sounds like many non-platform projects could benefit

PR #490, opened 2021-09-23
`PACTA_analysis` gains function `calculate_tdm()`

jdhoffa:
Here is my first stab at translating the IPR "Transition Disruption Metric" into code. 
It's definitely not perfect, but what I've done is the following: 

- Add a new exported function and corresponding tests to the `PACTA_analysis` package (I know this might be a contentious decision, and feel free to discuss, this doesn't need to be the way we do this. I just thought the benefits of having automated tests outweigh whatever reason not to do it)
- Add a couple of helper functions to ensure writing tests is easy and quick `fake_pacta_results()`(basically this is just a quick way to output variations on the `Equity_pacta_results_portfolio.rda` file.)
- Add a bunch of really typical/ obvious tests. 

See the reprex below for the functionality of the new function: 
``` r
library(dplyr, warn.conflicts = FALSE)
devtools::load_all()

# See also `fake_tdm_data()`
fake_equity_port <- tibble::tibble(
  investor_name = "some_investor",
  portfolio_name = "some_portfolio",
  allocation = "portfolio_weight",
  equity_market = "GlobalMarket",
  scenario_geography = "Global",
  ald_sector = "Power",
  technology = rep(c("RenewablesCap", "OilCap"), each = 4),
  year = rep(c(2020, 2021, 2025, 2030), 2),
  plan_alloc_wt_tech_prod = c(1, 1, 1, 1, 1, 2, 3, 4), 
  scen_alloc_wt_tech_prod = c(1, 2, 3, 4, 1, 0.75, 0.5, 0.25), 
  plan_carsten = c(0.5, 0.3, 0.25, 0.2, 0.5, 0.7, 0.75, 0.8),
)

grouping_vars <- c(
  "investor_name", 
  "portfolio_name", 
  "equity_market", 
  "scenario_geography"
)

out <- calculate_tdm(
  fake_equity_port, 
  2020, 
  grouping_vars
  )
#> Joining, by = c("investor_name", "portfolio_name", "equity_market", "scenario_geography", "ald_sector", "technology")

select(out, ald_sector, technology, tdm_tech, tdm_sec, reference_year)
#> # A tibble: 2 × 5
#>   ald_sector technology    tdm_tech tdm_sec reference_year
#>   <chr>      <chr>            <dbl>   <dbl>          <dbl>
#> 1 Power      RenewablesCap     2       4.67           2020
#> 2 Power      OilCap            7.33    4.67           2020
```

[...]

PR #506, opened 2021-11-12
`PACTA.analysis` gains `calculate_tdm` function

jdhoffa:
This PR supersedes the long-lived #490. 

This PR adds: 
* a function to calculate the transition disruption metric
* documentation explaining the metric in detail
* TESTSSSSSS

The code could likely still use some refactoring, there are definitely aspects I'm not perfectly happy with, but we need to get moving along with this. 

BIG TODO:
- [x] `calculate_tdm` should gain a `character` argument called `scenario` to indicate which scenario the TDM should be calculated with. I haven't yet implemented this as I don't still know exactly what the scenario name will be when it is prepared


jdhoffa:
Note, I squashed many many commits, but since #490 is so long -ived and the history is so messy, I think it makes more sense to start again somewhat fresh like this.

R/calculate_tdm.R
```diff
@@ -0,0 +1,305 @@
...
+#' pacta_results
+#'
+#' calculate_tdm(pacta_results, t0 = 2020)
+calculate_tdm <- function(data, t0, t1 = 5, t2 = 10, additional_groups = NULL) {
+  stopifnot(
+    is.data.frame(data),
+    is.numeric(t0),
+    is.numeric(t1),
+    is.numeric(t2)
+  )
```
maurolepore, line 47: I'm becoming fond of vctrs because it is rigurous and has nice error messages

``` r
# OK
t0 <- 2020
vctrs::vec_assert(t0, numeric(), size = 1L)

# Nice error message
t0 <- "2020"
vctrs::vec_assert(t0, numeric(), size = 1L)
#> Error: `t0` must be a vector with type <double>.
#> Instead, it has type <character>.

# Nice error message
t0 <- c(2020, 2021)
vctrs::vec_assert(t0, numeric(), size = 1L)
#> Error: `t0` must have size 1, not size 2.

# Is numeric the right prototpye?
t0 <- 2020L
vctrs::vec_assert(t0, numeric(), size = 1L)
#> Error: `t0` must be a vector with type <double>.
#> Instead, it has type <integer>.
```

<sup>Created on 2021-11-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/calculate_tdm.R
```diff
@@ -0,0 +1,305 @@
...
+
+warn_zero_rows <- function(data) {
+  # NOTE: This function will only work if the allocation method is
+  # portfolio_weight. ownership_weight outputs 0 for all carsten metric values,
+  # and thus we would be dividing by zero otherwise...
+  message <- 'Filtering for "portfolio_weight" allocation, outputs 0 rows'
+  warn(message, class = "has_zero_rows")
+
+  # Pass `data` to inline inside return()
+  invisible(data)
```
maurolepore, line 93: I don't understand what this means. I think the best way to explain `invisible(data)` is to allow using this function in a pipe. But then it's a well documented design pattern so I don't think you need to explain it.

https://design.tidyverse.org/out-invisible.html

R/calculate_tdm.R
```diff
@@ -0,0 +1,305 @@
...
+  left_join(data, monotonic_factors, by = groups)
+}
+
+add_tdm <- function(data, groups) {
+  data %>%
+    group_by(!!!rlang::syms(groups)) %>%
+    mutate(
+      .numerator = .data$scenario_production_plus_t2 - .data$portfolio_production_plus_t1,
+      .denominator = .data$scenario_production_plus_t2 - .data$scenario_production_t0,
+      # TODO: This case we
```
maurolepore, line 179: I don't understand this.

R/calculate_tdm.R
```diff
@@ -0,0 +1,305 @@
...
+
+check_calculate_tdm <- function(data, t0, t1, t2, groups) {
+  check_crucial_names(data, crucial_columns())
+
+  check_crucial_years(data, t0, t1, t2, groups)
+
+  check_unique_by_year_and_groups(data, groups)
+
+  invisible(data)
+}
```
maurolepore, line 241: :heart:

R/calculate_tdm.R
```diff
@@ -0,0 +1,305 @@
...
+
+  data <- data %>%
+    group_by(!!!rlang::syms(c("year", groups))) %>%
+    dplyr::summarize(rows_by_year_and_group = dplyr::n()) %>%
+    mutate(rows_are_unique = .data$rows_by_year_and_group == 1)
+
+  ok <- all(data$rows_are_unique)
+  if (!ok) {
+    rlang::abort(
+      "multiple_values_per_year",
```
maurolepore, line 258: If you're moving the second argument to position 1, I think it's best to name it explicitely, i.e. `class = ...`

R/calculate_tdm.R
```diff
@@ -0,0 +1,305 @@
...
+      message = glue(
```
maurolepore, line 259: The argument `message` is quite clever. You can get a nicer output. See `message` at `abort()`.

``` r
rlang::abort(c("Title.", x = "Problem.", i = "Hint"))
#> Error: Title.
#> x Problem.
#> ℹ Hint
```

<sup>Created on 2021-11-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/calculate_tdm.R
```diff
@@ -0,0 +1,305 @@
...
+    dplyr::summarize(missing_data = sum(is.na(.data$plan_alloc_wt_tech_prod)))
+
+  ok <- all(missing_crucial_years$missing_data == 0)
+
+  if (!ok) {
+    rlang::abort(
+      "missing_crucial_years",
+      message = glue(
+        "Data must contain all crucial years, by group:
+      {paste0('`', crucial_years, '`', collapse = ', ')} \n
```
maurolepore, line 285: You may like `glue_collapse()`

``` r
glue::glue_collapse(letters[1:10], ", ", last = " and ", width = 50)
#> a, b, c, d, e, f, g, h, i and j
glue::glue_collapse(letters, ", ", last = " and ", width = 50)
#> a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p,...
```

<sup>Created on 2021-11-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

tests/testthat/test-calculate_tdm.R
```diff
@@ -0,0 +1,160 @@
...
+
+test_that("errors if data has multiple values per year", {
+  fake_tdm_data(
+    year = c(2020, 2020, 2025, 2030),
+    plan_alloc_wt_tech_prod = c(1, 1, 3, 4),
+    scen_alloc_wt_tech_prod = c(1, 2, 3, 4),
+    plan_carsten = c(0.5, 0.3, 0.25, 0.2),
+  ) %>%
+    calculate_tdm(2020) %>%
+    expect_error(class = "multiple_values_per_year")
```
maurolepore, line 69: Consider `expect_snapshot_error()` to confirm the message prints with the format you want.

maurolepore (approved):
LGTM. Thanks! The code reads clearly; the test titles too.

Here are some suggestions.

R/calculate_tdm.R
jdhoffa: hahaha i think this was actually something you added in a commit!! maybe I'm wrong, but I have no idea what is meant by that comment.

DESCRIPTION
```diff
@@ -38,11 +38,15 @@ Imports:
     renv,
     rlang,
     rmarkdown,
+    tidyr,
```
cjyetman, line 41: are you certain `tidyr` is not used anywhere else in this repo? this seems a bit dangerous

DESCRIPTION
```diff
@@ -38,11 +38,15 @@ Imports:
...
     tibble,
     withr
 Suggests: 
-    testthat
+    knitr,
```
cjyetman, line 45: presumably this is only for the documentation and won't affect any code that would run on the website? At the moment the Docker image is locked in to a old version of R and specific versions of a few packages, including knitr <v1.34 because of requirements in create_interactive_report (see [create_interactive_report/issues/422](https://github.com/2DegreesInvesting/create_interactive_report/issues/422)

DESCRIPTION
```diff
@@ -38,11 +38,15 @@ Imports:
...
+    testthat (>= 3.0.0)
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.1.2
+Config/testthat/edition: 3
```
cjyetman, line 51: there are existing tests.... are you certain that switching to version 3 of testthat will not cause problems?

NAMESPACE
```diff
@@ -5,11 +5,29 @@ S3method(clean_quo,quosures)
...
 export(clean_quo)
 export(quo_chr)
 export(unclean_column_names)
 export(write_csv_file)
+importFrom(dplyr,all_of)
+importFrom(dplyr,arrange)
+importFrom(dplyr,case_when)
+importFrom(dplyr,contains)
+importFrom(dplyr,distinct)
+importFrom(dplyr,filter)
```
cjyetman, line 19: I think @AlexAxthelm prefers that `dplyr::filter()` is always namespaced, not sure exactly why but I it must conflict with some other common `filter()` function

DESCRIPTION
jdhoffa: Out of curiosity, why would it matter if `tidyr` is used anywhere else in the repo?

R/calculate_tdm.R
```diff
@@ -0,0 +1,305 @@
+#' Calculate and add the transition disruption metric to a dataset
+#'
+#' This function takes a typical PACTA output, and uses it to calculate the
+#' transition disruption metric.
+#'
+#' @param data A dataset, like the `*_results_portfolio.rda` outputs of PACTA.
```
cjyetman, line 6: maybe this should say that it accepts a data.frame (or a tibble) explicitly, because it almost sounds like it might take a path to an RDA file (which is actually a RDS file).

R/calculate_tdm.R
```diff
@@ -0,0 +1,305 @@
...
+#' @param t0 The start year, from which the TDM value should be calculated.
```
cjyetman, line 7: can you suggest what would be a reasonable setting for this?

R/calculate_tdm.R
```diff
@@ -0,0 +1,305 @@
...
+#' @param t1 The number of years into the future for which there is production
+#'   data. Default is 5 years.
+#' @param t2 The number of years into the future against which you want to
+#'   calculate disruption. Default is 10 years.
+#' @param additional_groups Character vector. The names of columns to group by,
+#'   in addition to these ones (which are always used):
+#'
+#'   `r toString(crucial_tdm_groups())`
+#'
+#' @return A tibble with the columns specified in the `...` input, as well as:
```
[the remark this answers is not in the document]
cjyetman: I don't understand this either

DESCRIPTION
jdhoffa: All automated tests (i.e. those that are in the `tests/testthat/` directory) are passing, as can be seen in the R CMD Checks on CI (I also manually run all tests before each push). 

Do you expect other problems? In particular, are there tests that use `testthat` that aren't run automatically when executing `ctrl + shift + T`? Or, do you expect problems that wouldn't be exposed by the test failing?

DESCRIPTION
jdhoffa: I believe this is an artifact of using `usethis::use_article()`. Since the package is under "Suggests", it is not actually used or required by any part of the package, it is just an indication that `knitr` could be useful to the package:

Screenshot 2021-11-14 at 22 05 42
e

My guess is that this is included automatically, since articles are not actually bundled with the package (in comparison to vignettes). See here: https://usethis.r-lib.org/reference/use_vignette.html

R/calculate_tdm.R
jdhoffa: Yup good shout! Normally I would point to a dummy dataset but we still need to create those.

R/calculate_tdm.R
jdhoffa: Personally, I don't think we should set a default argument for this parameter (since we would have to update it every year). 

Or do you mean to explain further what "start year" means?

R/calculate_tdm.R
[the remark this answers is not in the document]
jdhoffa: Yup, the name of the  `...` argument was changed to `additional_groups`. @maurolepore this was actually also you're doing!! ;-) heh. 

Will fix it, thanks guys :-)

R/calculate_tdm.R
jdhoffa: Ah it seems I forgot to finish my thought here lol. Will fix. 

For my own reference: what I wanted to explain here is that this where we handle the potential for a 0 denominator (resulting in an undefined value). I think we should actually NOT set this to 0, but rather decide on some form of warning, or "TDM not applicable" value to explain why this wouldn't make sense. @Antoine-Lalechere I believe, agreed with warning the user of this. 

I will think on an appropriate solution this week.

R/calculate_tdm.R
```diff
@@ -0,0 +1,305 @@
+#' Calculate and add the transition disruption metric to a dataset
+#'
+#' This function takes a typical PACTA output, and uses it to calculate the
+#' transition disruption metric.
+#'
+#' @param data A dataset, like the `*_results_portfolio.rda` outputs of PACTA.
+#' @param t0 The start year, from which the TDM value should be calculated.
+#' @param t1 The number of years into the future for which there is production
+#'   data. Default is 5 years.
```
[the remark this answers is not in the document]
jdhoffa: Hmm, ya it does feel as though it deserves emphasis. I explained a bit in an addednum to the article, but basically the default value of `t1=5` comes from the fact that our asset level data is only 5 years forward looking. In principle, this value could be anything, in practice, we will find that it is likely never going to be something other than 5 years. 

The default value of `t2=10` sounds like a pretty arbitrary choice, in that "Jakob and co. decided that 10 years was a good mid-point to estimate how disrupted you might be by a transition". A lot of jargon, but basically there's nothing to say that `t2=20` or `t2=6`

R/calculate_tdm.R
cjyetman: until [this issue](https://github.com/2DegreesInvesting/create_interactive_report/issues/422) is resolved, we won't be able to use the `vctrs` package or any package that is not already in the Docker image

R/calculate_tdm.R
```diff
@@ -0,0 +1,305 @@
...
+  invisible(data)
+}
+
+add_time_step <- function(data, t0, t1, t2) {
+  data %>%
+    mutate(
+      time_step = case_when(
+        .data$year == t0 ~ "t0",
+        .data$year == t0 + t1 ~ "plus_t1",
+        .data$year == t0 + t2 ~ "plus_t2"
```
cjyetman, line 102: isn't it more important to use the `.env$` pronoun than the `.data$` since the `.data$` variables are used by default?

R/calculate_tdm.R
```diff
@@ -0,0 +1,305 @@
...
+      )
+    )
+}
+
+add_monotonic_factor <- function(data, t0, t1, t2, groups) {
+  data <- data %>%
+    group_by(!!!rlang::syms(groups)) %>%
+    mutate(
+      end_year = last(.data$year),
```
cjyetman, line 111: are you sure `dplyr::last()` does what you want?
```r
x <- 2025:2020
dplyr::last(x)
# [1] 2020
max(x)
# [1] 2025
```

R/calculate_tdm.R
```diff
@@ -0,0 +1,305 @@
...
+      end_year_is_t0_t2 = .data$end_year == t0 + t2
+    )
+
+  if (all(data$end_year_is_t0_t2)) {
+    monotonic_factors <- data %>%
+      mutate(
+        .is_monotonic = TRUE,
+        monotonic_factor = dplyr::if_else(.data$.is_monotonic, 1, -1),
+        .is_monotonic = NULL
```
cjyetman, line 120: is this not the same as 
```r
mutate(monotonic_factor = 1)
```

R/calculate_tdm.R
```diff
@@ -0,0 +1,305 @@
...
+      ) %>%
+      select(
+        -c(
+          .data$year,
+          .data$end_year,
+          .data$plan_carsten,
+          .data$plan_alloc_wt_tech_prod,
+          .data$scen_alloc_wt_tech_prod,
+          .data$end_year_is_t0_t2
+        )
```
cjyetman, line 130: is this normal to wrap variable names in a `c()` in `select()`?

R/calculate_tdm.R
```diff
@@ -0,0 +1,305 @@
...
+    distinct() %>%
+    ungroup()
+}
+
+add_aggregate_tdm <- function(data, t0, groups) {
+  data %>%
+    group_by(!!!rlang::syms(groups)) %>%
+    ungroup(.data$technology) %>%
+    mutate(
+      tdm_sec = .data$plan_carsten * sum(.data$tdm_tech) / sum(.data$plan_carsten),
```
cjyetman, line 198: do you need to use `na.rm = FALSE` here to be safe?

DESCRIPTION
cjyetman: weird, I thought this was a removal when I read it

DESCRIPTION
cjyetman: any idea what will happen when `devtools::load_all()` is called if `knitr` is not installed? like here https://github.com/2DegreesInvesting/PACTA_analysis/blob/675ed324da6a2e6315af2c36ab30336d1a891e49/web_tool_script_1.R#L3

DESCRIPTION
cjyetman: just concerned/curious because last time I experimented with "edition 3" it broke a bunch of my tests so I just reverted

R/calculate_tdm.R
cjyetman: yeah, like just an explanation of what it usually would be (relative to the given dataset).... like "this would usually be the earliest year in your PACTA results dataset"

DESCRIPTION
jdhoffa: So it looks like `knitr` is a dependency of `rmarkdown`, and `rmarkdown` IS actually imported by this package. 
(i.e. `devtools::load_all()` will fail if `knitr` is not installed, but it's not because `knitr` is a suggested package).

R/calculate_tdm.R
jdhoffa: afaik this is best practice, but i'm gonna divert that question to @maurolepore for a better explanation

R/calculate_tdm.R
jdhoffa: I have done it frequently when dropping variables (mainly because you can extract the `-` sign out, so it's a little easier to read). I don't know if that means it is `norma;` or not

DESCRIPTION
cjyetman: yeah, so `knitr` is installed in the current Docker image. It's stuck on version 1.33 or less though, so we have to be careful about what/how relies on it.

jdhoffa: yup, so again this is totally fine since it's `Suggests`. Nothing relies on it in any way.

DESCRIPTION
jdhoffa: Ok, and by "broke tests", you mean they were failing after? Since they are not failing here, do you think it's safe to assume we are good?

NAMESPACE
AlexAxthelm: `base::filter`

R/calculate_tdm.R
jdhoffa: You are correct.

DESCRIPTION
cjyetman: probably... I'm pretty sure it caused some of my tests to error out.... I think it was in the data_prep repo though

cjyetman: I'm a little worried because I haven't fully wrapped my head around the edition 3 changes yet, so I fear it may hamper my ability to write tests, but I guess I'll need to get there at some point anyway

R/calculate_tdm.R
AlexAxthelm: and in general, I like not adding new dependencies to production code unless there is a _really_ big benefit to them.

NAMESPACE
cjyetman:
```r
base::filter
# Error in get(name, envir = ns, inherits = FALSE) : 
#   object 'filter' not found
```
?

cjyetman: ah, nevermind `stats::filter`

cjyetman: there's a `base::Filter` too 🤯

R/calculate_tdm.R
cjyetman: I think the `.data$` usage is "best practice" because it avoids `R CMD check` warnings.

I personally believe it's "good practice" to call environment variables (versus data variables) explicitly using the `.env$` pronoun because by default dplyr will use the data variable if there is a data variable and an environment variable with the same name.

e.g.
```r
library(dplyr)

t0 <- 2020

tibble(x = "ABC", t0 = 1984) %>% 
  mutate(t1 = t0 + 5)
# # A tibble: 1 x 3
#   x        t0    t1
#   <chr> <dbl> <dbl>
# 1 ABC    1984  1989

tibble(x = "ABC", t0 = 1984) %>% 
  mutate(t1 = .env$t0 + 5)
# # A tibble: 1 x 3
#   x        t0    t1
#   <chr> <dbl> <dbl>
# 1 ABC    1984  2025
```

R/calculate_tdm.R
cjyetman: to be clear, it's not really the create_interactive_report issue, but that issue makes updating the base Docker image we use problematic, which makes adding a new package dependency to it problematic... not impossible, but dicey

R/calculate_tdm.R
jdhoffa: ah ok, i mean in that case we definitely want `.data` here and not `.env`.

R/calculate_tdm.R
jdhoffa: `na.rm = FALSE` is the default, no?

R/calculate_tdm.R
jdhoffa: this seems interesting, but for the purpose of this PR I'm not going to go into the pros/ cons of `vctrs`. Lets save it for another discussion :-)

R/calculate_tdm.R
cjyetman: sorry, I meant `na.rm = TRUE`.... if there's any chance at all that `.data$tdm_tech` or `.data$plan_carsten` could have an `NA` in them (or a `NULL`), you definitely want to use `na.rm = TRUE`, because otherwise this will resolve to  `NA`, which is probably not what you want (the default sucks imo), e.g.
```r
sum(c(1, 2, 3))
# [1] 6
sum(c(1, 2, 3, NA))
# [1] NA
sum(c(1, 2, 3, NA), na.rm = TRUE)
# [1] 6
```

R/calculate_tdm.R
cjyetman: so in the case of `.data$year == t0 ~ "t0",` you want `t0` to be the column in the data, not the `t0` that is passed as an argument?

jdhoffa:
Ok @maurolepore @cjyetman I have added a `scenarios` argument, that allows the user to pass a character vector of scenarios, which will pre-filter the dataset so that TDM is only calculated for those scenarios. 

The default is `IPR FPS 2021`, which is equivalent to the name found int he dataset that Alex shared. 

I believe this PR is now ready to be merged, but will re-request a review from you both in case you believe there are any other blocking issues!

maurolepore (approved):
LGTM. Thanks! But please address the RCMD check issues.

cjyetman:
> I believe this PR is now ready to be merged, but will re-request a review from you both in case you believe there are any other blocking issues!

I wanna actually give this a go, which hopefully will be possible in the coming days as I get the scenario prep and data_prep things working with IPR... can we review and merge this after that?

jdhoffa:
Ya, that is fair. My thoughts were, just given the timeline, I would also like to start working on some things downstream of this (ie. starting to figure out where to include this in `web_tool_*`, and thinking about output data format so that Monika becomes unblocked). I am somewhat blocked until this gets merged (as I don't really want to branch off this and risk merge conflicts). 

If you think data prep will only take a day or two, then I can totally wait that long, but if it's more likely to take a week or two, then I'd prefer to get moving (so I'll leave you to be the judge of that haha). 

Also, just FYI, it is possible to iterate and test quickly against multiple different `*_results_portfolio.rda` type datasets by using the `fake_tdm_data()` function. 

Something like: 
``` r
devtools::load_all(".")

data <- fake_tdm_data(
    technology = rep(c("RenewablesCap", "OilCap"), each = 4),
    year = rep(c(2020, 2025, 2030, 2050), 2),
    plan_alloc_wt_tech_prod = c(rep(0.1, 4), rep(5, 4)),
    scen_alloc_wt_tech_prod = c(1, 0.5, 0.25, 4, 1, 2, 3, 0.25),
    plan_carsten = 0.5,
 )

 out <- calculate_tdm(data, 2020)
```

cjyetman:
I think that I will have a 2020Q4 analysis inputs folder that includes the preliminary IPR data, at least one to test with, by the end of today.

jdhoffa:
Ok hell ya, then that's totally fine!

cjyetman:
> I think that I will have a 2020Q4 analysis inputs folder that includes the preliminary IPR data, at least one to test with, by the end of today.

So I ran into unexpected memory issues running data_prep with the IPR data, so I didn't get through the full process yet, but I still hope/anticipate getting that done today and testing this function out on real data.

cjyetman:
this is what my first attempts to try this out looked like (which may inform some improvements)
``` r
devtools::load_all()
#> ℹ Loading PACTA.analysis

port_all_eq <- readRDS("working_dir/40_Results/TestPortfolio_Input/Equity_results_portfolio.rda")

port_all_eq$year %>% unique() %>% sort()
#> [1] 2021 2022 2023 2024 2025 2026 2030

port_all_eq$scenario %>% unique() %>% sort()
#>  [1] "ETP2017_B2DS"         "ETP2017_NPSRTS"       "ETP2017_SDS"         
#>  [4] "GECO2019_1.5c"        "GECO2019_2c_m"        "GECO2019_ref"        
#>  [7] "IPR2021_IPR FPS 2021" "WEO2019_CPS"          "WEO2019_SDS"         
#> [10] "WEO2019_SPS"          "WEO2020_SDS"          "WEO2020_SPS"  

calculate_tdm(port_all_eq, t0 = 2021)
#> Error: Data must contain all crucial years, by group:
#>   x c(2021, 2026, 2031)
#>   ℹ Are all crucial years present in input data?
#>   Run `rlang::last_error()` to see where the error occurred.

calculate_tdm(port_all_eq, t0 = 2021, t2 = 9)
#> Error: Data must be unique by year and groups for the following columns
#>  x scen_alloc_wt_tech_prod,plan_alloc_wt_tech_prod...
#>  ℹ Are you sure you have included the correct groupings?
#>  Run `rlang::last_error()` to see where the error occurred.

calculate_tdm(port_all_eq, t0 = 2021, t2 = 9, additional_groups = c("investor_name", "portfolio_name", "scenario_source", "scenario", "allocation", "equity_market", "scenario_geography", "year"))
#> Error: Must group by variables found in `.data`.
#>  * Column `year` is not found.
#>  Run `rlang::last_error()` to see where the error occurred.

calculate_tdm(port_all_eq, t0 = 2021, t2 = 9, additional_groups = c("investor_name", "portfolio_name", "scenario_source", "scenario", "allocation", "equity_market", "scenario_geography"))
#> Warning message:
#> Filtering for `scenarios`, outputs 0 rows 

calculate_tdm(port_all_eq, t0 = 2021, t2 = 9, scenarios = "IPR2021_IPR FPS 2021", additional_groups = c("investor_name", "portfolio_name", "scenario_source", "scenario", "allocation", "equity_market", "scenario_geography"))
#> Error: Problem with `mutate()` column `.numerator`.
#> ℹ `.numerator = .data$scenario_production_plus_t2 - .data$portfolio_production_plus_t1`.
#> x non-numeric argument to binary operator
#> ℹ The error occurred in group 1: technology = "Ac-Electric Arc Furnace", ald_sector = "Steel", investor_name = "Test", portfolio_name = "TestPortfolio_Input", scenario_source = "IPR2021", scenario = "IPR2021_IPR FPS 2021", allocation = "portfolio_weight", equity_market = "DevelopedMarket", scenario_geography = "Global".
#> Run `rlang::last_error()` to see where the error occurred.
#> In addition: Warning messages:
#> 1: Values are not uniquely identified; output will contain list-cols.
#> * Use `values_fn = list` to suppress this warning.
#> * Use `values_fn = length` to identify where the duplicates arise
#> * Use `values_fn = {summary_fun}` to summarise duplicates 
#> Error: Problem with `mutate()` column `.numerator`.
#>  ℹ `.numerator = .data$scenario_production_plus_t2 - .data$portfolio_production_plus_t1`.
#>  x non-numeric argument to binary operator
#>  ℹ The error occurred in group 1: technology = "Ac-Electric Arc Furnace", ald_sector = "Steel", investor_name = "Test", portfolio_name = "TestPortfolio_Input", scenario_source = "IPR2021", scenario = "IPR2021_IPR FPS 2021", allocation = "portfolio_weight", equity_market = "DevelopedMarket", scenario_geography = "Global".
#>  Run `rlang::last_error()` to see where the error occurred.
```

cjyetman:
I find it potentially confusing that, in the arguments and in the code, `t0` refers to a year like `t0 = 2021`, while `t1` and `t2` refer to a quantity of years like `t1 = 5` and `t2 = 10`.... and then in the code we have stuff like 
```r
mutate(
      end_year = max(.data$year),
      end_year_is_t0_t2 = .data$end_year == t0 + t2
    )
```
instead of
```r
mutate(
      end_year = max(.data$year),
      end_year_is_t0_t2 = .data$end_year == t2
    )
```

or, for instance, we might have...
```r
data %>% filter(year %in% c(t0, (t0 + t1), (t0 + t2)))
```
instead of...
```r
data %>% filter(year %in% c(t0, t1, t2))
```

cjyetman:
I narrowed down my initial problems using this to `add_monotonic_factor()`. I'm getting the following output from it when I step through the 
```r
filtered_data
# # A tibble: 462 x 30
#    investor_name portfolio_name    scenario_source scenario       allocation    equity_market
#    <chr>         <chr>             <chr>           <chr>          <chr>         <chr>        
#  1 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
#  2 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
#  3 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
#  4 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
#  5 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
#  6 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
#  7 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
#  8 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
#  9 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
# 10 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
# # … with 452 more rows, and 24 more variables: scenario_geography <chr>, year <dbl>,
# #   ald_sector <chr>, technology <chr>, plan_tech_prod <dbl>, plan_alloc_wt_tech_prod <dbl>,
# #   plan_carsten <dbl>, plan_emission_factor <dbl>, scen_tech_prod <dbl>,
# #   scen_alloc_wt_tech_prod <dbl>, scen_carsten <dbl>, scen_emission_factor <dbl>,
# #   plan_sec_prod <dbl>, plan_alloc_wt_sec_prod <dbl>, plan_sec_carsten <dbl>,
# #   plan_sec_emissions_factor <dbl>, scen_sec_prod <dbl>, scen_alloc_wt_sec_prod <dbl>,
# #   scen_sec_carsten <dbl>, scen_sec_emissions_factor <dbl>, plan_tech_share <dbl>,
# #   scen_tech_share <dbl>, trajectory_deviation <dbl>, trajectory_alignment <dbl>

add_monotonic_factor(filtered_data, t0, t1, t2, groups)
# # A tibble: 3,234 x 50
# # Groups:   technology, ald_sector, investor_name, portfolio_name, scenario_source,
# #   scenario, allocation, equity_market, scenario_geography [66]
#    investor_name portfolio_name    scenario_source scenario       allocation    equity_market
#    <chr>         <chr>             <chr>           <chr>          <chr>         <chr>        
#  1 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
#  2 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
#  3 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
#  4 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
#  5 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
#  6 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
#  7 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
#  8 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
#  9 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
# 10 Test          TestPortfolio_In… IPR2021         IPR2021_IPR F… portfolio_we… DevelopedMar…
# # … with 3,224 more rows, and 44 more variables: scenario_geography <chr>, year <dbl>,
# #   ald_sector <chr>, technology <chr>, plan_tech_prod.x <dbl>,
# #   plan_alloc_wt_tech_prod <dbl>, plan_carsten <dbl>, plan_emission_factor.x <dbl>,
# #   scen_tech_prod.x <dbl>, scen_alloc_wt_tech_prod <dbl>, scen_carsten.x <dbl>,
# #   scen_emission_factor.x <dbl>, plan_sec_prod.x <dbl>, plan_alloc_wt_sec_prod.x <dbl>,
# #   plan_sec_carsten.x <dbl>, plan_sec_emissions_factor.x <dbl>, scen_sec_prod.x <dbl>,
# #   scen_alloc_wt_sec_prod.x <dbl>, scen_sec_carsten.x <dbl>,
# #   scen_sec_emissions_factor.x <dbl>, plan_tech_share.x <dbl>, scen_tech_share.x <dbl>,
# #   trajectory_deviation.x <dbl>, trajectory_alignment.x <dbl>, end_year <dbl>,
# #   end_year_is_t0_t2 <lgl>, plan_tech_prod.y <dbl>, plan_emission_factor.y <dbl>,
# #   scen_tech_prod.y <dbl>, scen_carsten.y <dbl>, scen_emission_factor.y <dbl>,
# #   plan_sec_prod.y <dbl>, plan_alloc_wt_sec_prod.y <dbl>, plan_sec_carsten.y <dbl>,
# #   plan_sec_emissions_factor.y <dbl>, scen_sec_prod.y <dbl>,
# #   scen_alloc_wt_sec_prod.y <dbl>, scen_sec_carsten.y <dbl>,
# #   scen_sec_emissions_factor.y <dbl>, plan_tech_share.y <dbl>, scen_tech_share.y <dbl>,
# #   trajectory_deviation.y <dbl>, trajectory_alignment.y <dbl>, monotonic_factor <dbl>
```

In this case, the data we have stops at 2030, so we must assume that the data is "monotonic". The `add_monotonic_factor()` function checks if this is the case first and has an `if` structure to deal with this type of case by simply creating a new column `monotonic_factor` and setting the value to `1`. It does that here...
https://github.com/2DegreesInvesting/PACTA_analysis/blob/59c20de8e6ef9cf1afff7955fc0172eedfe4c7e4/R/calculate_tdm.R#L133-L147

I propose that we simplify that section to...
```r
monotonic_factors <- 
  data %>%
  ungroup() %>% 
  select(.env$groups) %>%
  distinct() %>% 
  mutate(monotonic_factor = 1L)
```

I tried this locally, passed all tests, and then I get...
```r
calculate_tdm(port_all_eq, t0 = 2021, t2 = 9, scenarios = "IPR2021_IPR FPS 2021", additional_groups = c("investor_name", "portfolio_name", "scenario_source", "scenario", "allocation", "equity_market", "scenario_geography"))
#> # A tibble: 66 x 12
#>    technology   ald_sector tdm_tech   tdm_sec reference_year investor_name portfolio_name    
#>    <chr>        <chr>         <dbl>     <dbl>          <dbl> <chr>         <chr>             
#>  1 Electric     Automotive     1.74  9.01               2021 Test          TestPortfolio_Inp…
#>  2 Electric_HDV Automotive     2.00  0.0151             2021 Test          TestPortfolio_Inp…
#>  3 FuelCell     Automotive     2.00  0.00185            2021 Test          TestPortfolio_Inp…
#>  4 FuelCell_HDV Automotive     2.00  0.000248           2021 Test          TestPortfolio_Inp…
#>  5 Hybrid       Automotive     0     2.67               2021 Test          TestPortfolio_Inp…
#>  6 Hybrid_HDV   Automotive     0     0.000857           2021 Test          TestPortfolio_Inp…
#>  7 ICE          Automotive     2.17 43.0                2021 Test          TestPortfolio_Inp…
#>  8 ICE_HDV      Automotive    47.1   2.26               2021 Test          TestPortfolio_Inp…
#>  9 Freight      Aviation       0     0                  2021 Test          TestPortfolio_Inp…
#> 10 Passenger    Aviation       0     0                  2021 Test          TestPortfolio_Inp…
#> # … with 56 more rows, and 5 more variables: scenario_source <chr>, scenario <chr>,
#> #   allocation <chr>, equity_market <chr>, scenario_geography <chr>

```

cjyetman (approved):
thanks @jdhoffa! latest version works as expected 🚀

PR #507, opened 2021-11-15
enable reading `holdings_date` from portfolio parameters

cjyetman:
enable reading `holdings_date` from portfolio parameters and setting `data_location_ext` with proper precedence

this will allow setting `data_location_ext` with the following precedence:
1. project parameters
2. portfolio parameters
3. web parameters (default)

I renamed the imported project parameter from `data_location_ext` to `proj_data_location_ext`. Originally, I allowed `set_project_parameters()` to just overwrite the `data_location_ext` set by  `set_web_parameters()` by default. Now instead the project data location parameter is read in as a unique variable (or as `NULL`) so that `set_analysis_inputs_path()` can properly set `analysis_inputs_path` based on the precedence above.

concerns I still have:
1. my understanding of this is that during the analysis `analysis_inputs_path` should always be used to find the input data, though some code may look/read in from `data_location_ext` (this also potentially applies to stress testing @jacobvjk @MirijaHa)... this is not a new problem, but now that we've created 2 ways of setting `analysis_inputs_path` to something different than what `data_location_ext` is set to by default, it kind of ups the ante for things to go wrong
2. I did not want to change the function signature of `set_analysis_inputs_path()`, and then be forced to change every call to it in every repo, so... `set_analysis_inputs_path()` now relies on variables `proj_data_location_ext` and `port_holdings_date` being available in a parent or global environment.

### DIFF
--- a/0_global_functions.R
+++ b/0_global_functions.R
@@ -46,9 +46,7 @@ set_col_types <- function(grouping_variables, fixed_col_types) {
 set_project_parameters <- function(file_path){
   cfg <- config::get(file = file_path)
 
-  if (!is.null(cfg$paths$data_location_ext)) {
-    data_location_ext <<- cfg$paths$data_location_ext
-  }
+  proj_data_location_ext <<- cfg$paths$data_location_ext
 
   project_report_name <<- cfg$reporting$project_report_name
   display_currency <<- cfg$reporting$display_currency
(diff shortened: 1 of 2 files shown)


0_global_functions.R
```diff
@@ -46,9 +46,7 @@ set_col_types <- function(grouping_variables, fixed_col_types) {
 set_project_parameters <- function(file_path){
   cfg <- config::get(file = file_path)
 
-  if (!is.null(cfg$paths$data_location_ext)) {
-    data_location_ext <<- cfg$paths$data_location_ext
-  }
+  proj_data_location_ext <<- cfg$paths$data_location_ext
```
AlexAxthelm, line 49: This will write `NULL` if it doesn't exist in the config file?

AlexAxthelm (approved):
one open question, but looks good if that resolves.

0_global_functions.R
cjyetman: exactly

cjyetman:
```r
list()$test_nonexistant_name_element
# NULL
```

AlexAxthelm: let's ship it.

jacobvjk:
@cjyetman we are not assigning anything to `data_location_ext`, so I would expect this to be fine from our side

cjyetman:
> @cjyetman we are not assigning anything to `data_location_ext`, so I would expect this to be fine from our side

thanks @jacobvjk , but to be clear, I'm more concerned about `data_location_ext` being used to read in data rather than `analysis_inputs_path`... e.g.
```r
read_rds(paste0(data_location_ext, "/masterdata_ownership_datastore_technology_type_view.rda"))
```
instead of
```r
read_rds(paste0(analysis_inputs_path, "/masterdata_ownership_datastore_technology_type_view.rda"))
```

I haven't found anything like that in the stress testing code, but I have at least in the documentation in this repo... and it's good to know for the future.

Also FYI, stress testing has a version of `set_analysis_inputs_path()` that is now out of sync with the version here...
https://github.com/2DegreesInvesting/r2dii.climate.stress.test/blob/master/0_global_functions_st.R#L121-L130

PR #508, opened 2021-11-16
Add function to pull start date from holdings date

AlexAxthelm:
if `start_date` not defined in project parameters, then use one from the
`holdings_date` in portfolio parameters, with `2020L` fallback if neither is
defined.


cjyetman (approved):
to be honest, I'm worried about the long term impact of this, but we gotta do what we gotta do to get things working today

I'd much rather see `start_year` inferred from the data

AlexAxthelm:
agreed.

jacobvjk:
same here, which is also how I understand PACTA for Banks to function? Just saying that I am more than happy to adjust the ST to use a data based `start_year` in the web tool

PR #509, opened 2021-11-17
Add infrastructure for HTML user error reports

AlexAxthelm:
This commit adds a lot to solve a relatively small problem, but opens up
a better line of communication with users in the future.

Generally, this adds `log_user_errors()` and `prepare_error_report()`,
which  allow us to identify user-level errors, which they can fix
before running a new analysis.

Specifically, this checks that the `port_holdings_date` in the
portfolio parameters file is a scalar, not an array (which can happen if
the user groups portfolios with different holding dates.)

Input very welcome.


cjyetman:
Wow, this is impressively sophisticated and feature-rich compared to what I was imagining! 🤣 bravo!

Other than the R-CMD-check error, this looks pretty good, but I'd like to actually test it out before I approve it, which may be a few hours from now.... is that ok @AlexAxthelm?

AlexAxthelm:
> I'd like to actually test it out before I approve it

please do. The checks failing is surprising  to me, since it looks like it's complaining about a lot of stuff that I didn't touch 😆

0_global_functions.R
```diff
@@ -393,6 +393,9 @@ write_log <- function(msg, file_path = log_path, ...) {
 }
 
 convert_quarter_to_year <- function(quarter_string){
+  # see function definition for more info, but this give the user a useful
+  # error if length > 1 (grouped portfolio from multiple timestamps)
```
maurolepore, line 397: Nit: If anywhere I would move this comment to the function's body. Here the risk is high that the comment will fall out of sync with the implementation.

R/user_error_checks.R
```diff
@@ -0,0 +1,45 @@
+# This function exists to check if a portfolio has a single holdings
+# date, or multiple. If there is a single date, then everything is good.
+# but on the platform, if a user groups multiple  portfolios, and some of
+# thse have different holding dates, then the parameters file will
```
maurolepore, line 4: Nit: `[-thse-]{+these+}`

R/user_error_checks.R
```diff
@@ -0,0 +1,45 @@
...
+#
+# grouped portfolio with different dates:
+# holdings_date: [2019Q4, 2020Q4]
+#
+# `config` will read this in as a character vector, so all we need to do
+# here is check for length > 1
+
+check_grouped_portfolio_years <- function(
+  port_holdings_date
+  ) {
```
maurolepore, line 23: Not gonna get into style :-)

R/user_error_checks.R
```diff
@@ -0,0 +1,45 @@
...
+      "Delete this grouping, and create a new one with portfolios",
+      "with the same holdings date before running the analysis again."
+    )
+    log_user_errors(
+      error_name = error_name,
+      suggested_action = action,
+      description = desc,
+      immediate = TRUE
+    )
+    stop("port_holdings_date length > 1")
```
maurolepore, line 43: Nit: Maybe move this to the top so it's one case less we need to think about. Then we can focus on the longer happy path.

Nit: Maybe use `call. = FALSE` as that info clutters the console and can be retrieved easily with `traceback()`.

> Use stop(call. = FALSE), rlang::abort(), Rf_errorcall(R_NilValue, ...) to avoid cluttering the error message with the name of the function that generated it. That information is often not informative, and can easily be accessed via traceback() or an IDE equivalent.

https://style.tidyverse.org/error-messages.html

cjyetman:
ok, this wasn't working the way I expected it to, so let me clarify your intent...

If a `start_year` is set in the project parameters, then this will never run? That's because we allow a project to set a specific `start_year`, and that way even if users within that project group portfolios with different `start_year`s, we force it to assume all grouped portfolios are of the `start_year` set in the project parameters. That's basically to protect COP users/projects.

is that correct?

R/user_error_checks.R
maurolepore: Nit. This (and other functions here) seems to be called primarily for its side effects. Then it's nice to return the first argument invisibly so we can pipe it is we choose so.

https://design.tidyverse.org/out-invisible.html

R/user_error_infrastructure.R
```diff
@@ -0,0 +1,107 @@
...
+  immediate = FALSE
+  ) {
+  # check for strings
+  stopifnot(is.character(error_name))
+  stopifnot(is.character(suggested_action))
+  stopifnot(is.character(description))
+  # check that directory exists, create if not
+  if (!dir.exists(dirname(error_file))) {
+    dir.create(dirname(error_file), recursive = TRUE)
+  }
```
maurolepore, line 42: Nit. The comment is redundant and, as any other comment, is at risk of eventually falling out of sync with the code.

R/user_error_checks.R
```diff
@@ -0,0 +1,45 @@
+# This function exists to check if a portfolio has a single holdings
+# date, or multiple. If there is a single date, then everything is good.
+# but on the platform, if a user groups multiple  portfolios, and some of
+# thse have different holding dates, then the parameters file will
+# contaain a yaml array with the distinct dates from each grouped
+# portfolio.
+# Examples:
```
maurolepore, line 7: Nit. Maybe add an example showing how we should call this function?

maurolepore (approved):
LGTM. See a PR that extends this one -- hopefully fixing R CMD check issues.

cjyetman:
ok, so one edge (maybe?) case is if both portfolios have the same `holdings_date` and they both get added to the portfolio/s parameter file, e.g. `holdings_date: ["2020Q4", "2020Q4"]`, in which case this will error out the script and create the user error report including "The portfolios in this group have holding dates of: 2020Q4, 2020Q4".

@AlexAxthelm is this something worth being defensive about?

cjyetman:
@AlexAxthelm can you maybe add a call somewhere to `write_log()` (from [here](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/0_global_functions.R#L385-L393)) so that we'll also log this error to the standard log file where other problems are recorded?

0_global_functions.R
AlexAxthelm: this is in the function body, but I clarified which function I'm referring to.

R/user_error_checks.R
AlexAxthelm: The "unhappy" path (where there is an error) is the long one.

I generally prefer structures where it's along the lines of 
```r
x <- function() {
  if (condition) {
    stop("condition")
  }
  return(value)
}
```

over constructions like

```r
x <- function() {
  if (!condition) {
    return(value)
  }
  stop("condition")
  # it feels weird to have the last line of the function not be return()
}
```

There are cases where the second construction is better, but in general, if I'm looking for what the function gives me in the happy path (no `stop()`s or similar errors), the ffirst place I'm going to look is the last line of the function.

R/user_error_infrastructure.R
AlexAxthelm: agreed.

AlexAxthelm:
> If a start_year is set in the project parameters, then this will never run? That's because we allow a project to set a specific start_year, and that way even if users within that project group portfolios with different start_years, we force it to assume all grouped portfolios are of the start_year set in the project parameters. That's basically to protect COP users/projects.

Correct, @cjyetman . basically, if `port_holdings_date` doesn't get used, then none of this machinery gets called. Note that the only path into this code right now is in `set_project_parameters()`, where it's running the fallback check to use the holding date from portfolio if it isn't in the project params. (line 61 is where this kicks off).

https://github.com/2DegreesInvesting/PACTA_analysis/blob/9d7d09b8f82ab592250cbe6facc80af608283a8d/0_global_functions.R#L58-L65

> ok, so one edge (maybe?) case is if both portfolios have the same `holdings_date` and they both get added to the portfolio/s parameter file...

definitely worth getting defensive about. I've added 2 `unique()`s: one in the string conversion function, and another in the validation function.

> can you maybe add a call somewhere to write_log()

Done. Note that passing a vector (or list) to `write_log()` works fine, but writes multiple lines.

AlexAxthelm:
The failing test is expected, since the stress testing repo is currently not supporting the webtool. Should be fixed Soon™️

maurolepore:
https://github.com/2DegreesInvesting/r2dii.climate.stress.test/pull/232 should restore support for the webtool. If urgent touch base with Mirja.

R/user_error_checks.R
maurolepore: Oh, I see -- this `check_*()` function writes logs before it aborts. Then it makes sense what you say. I guess I wouldn't have missed if the  details were extracted away the the two main actions are clearer: Log and abort:

```r
f <- function(first_argument) {
  if (something_happens) {
    log()
    abort()
  }

  invisible(first_argument)
}
```

maurolepore (commented):
Thanks for the responses.

cjyetman (approved):

maurolepore:
@AlexAxthelm FYI all CI checks are now green.

PR #510, opened 2021-11-18
feature/html user error reports mll

maurolepore:
This PR merges into `feature/html-user-error-reports`. It aims to fix the issues in R CMD check.

- Use package jsonlite
- Add non-standard dir to .Rbuildignore
- FIXME: Define global variables


AlexAxthelm (approved):

AlexAxthelm:
Good fixes, thanks @maurolepore

PR #512, opened 2021-11-19
pass tdm data to create interactive report

cjyetman:
depends on 2DegreesInvesting/create_interactive_report/pull/444
depends on #511 
depends on #506 
depends on 2DegreesInvesting/pacta-data/pull/48

@AlexAxthelm @jdhoffa @MonikaFu

PR #513, opened 2021-11-20
Pass config files

cjyetman:
depends on 2DegreesInvesting/create_interactive_report/pull/443
depends on #511 
depends on #512


jdhoffa (approved):
lgtm

Issue #466, opened 2021-07-02
ensure that devtools is installed in 1_portfolio_check_initialization


andybue:
One of the first comments of the script (in line6) is:

`devtools::load_all()`

In the collobartion with Sweden,  we tried to run the code offline.  we realized, that, if devtools, it not installed, the code does not return explicitly, that devtools is not installed, but complains about missing function in a later points.

It would be great, if a line is added, that asserts that devtools is installed, and either direclty installs it or suggests to install it before line6

cjyetman:
While I sympathize with the intent here, there are many required packages for PACTA to run beyond just `devtools`. See for instance https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/deduplicate/load-and-attach-r-packages.R

So if any such assertions should be made, it should probably be for a full list of required packages.

Also of note... as long as this repo is not fully transformed into a true package (which would then manage all these things in a more automatic way), I would love to rip out all this developer/package style stuff because I think it just way over-complicates things, e.g. using `devtools` to load scripts in the `R/` directory, this weird `deduplicate/` directory with some other utility functions, these deeply nested utility functions that you have to follow down 3 or 4 levels before you can figure out what it is actually doing, etc.

AlexAxthelm:
I agree with CJ on this point. we've not prioritized making this into a package, which would take care of a lot of these concerns for us. If we're planning on having more users running this on their own in the future, then we can reprioritize that, but that would be a pretty big chunk of work. 

Since this package iis under active development, keeping track of the dependencies is not impossible, but it could get tricky. Installing common packages is a pretty common issue when working in R, so I'd expect users who have installed software that's still in development to be used to this kind of issue.

cjyetman:
Also, after thinking about this more.... I suspect that what's happening is that the people are selecting portions of the code in the editor and clicking run or command-enter, which means an "assertion" will not stop this problem either. The call to a function that is not in scope because the package is not installed should already serve the purpose of an assertion, because it too will cause an error and halt if the script is run as a complete script, e.g. using `source()`. But it's a very common problem that R users manually run chunks of code from the editor, and the problem is that RStudio will run each selected line in sequence regardless of whether the previous line errors or not, so a failed assertion will also not prevent the following lines from running, which causes a lot of confusion when people don't realize that. I even wrote a package to deal with this problem: https://github.com/cjyetman/safe.run

[...]

PR #514, opened 2021-11-22
begin very minimal auto-testing of 0_portfolio_input_check_functions.R

cjyetman:
This is a barebones test to begin the process of testing functions defined within `0_portfolio_input_check_functions.R` so that we can start to refactor them without so much worry. The test data frames are likely to be changed very rapidly and frequently as I build out more tests for more functions, but this is meant to lay the groundwork so I have something to build from.


cjyetman:
need to figure out how to get this to pass R CMD check.... need to be able to source a file in the root directory.... `rprojroot::find_package_root_file()` works in `{testthat}` tests but not R CMD check.

jdhoffa:
I know you're nervous about this but.... it looks like this just tests the function `add_fin_data()`, and looking at that function, it could easily be extracted into the `PACTA.analysis` package (literally just by copying and pasting it), and you wouldn't have to worry about sourcing

jdhoffa:
Doing so would remove the `rprojroot` dependency, and the hoops that are necessary to jump through to get `testthat` working

cjyetman:
Some of the "old"/"offline"/"multiple"/whatever you wanna call it code will very likely break if we do that. Maybe that's not such a terrible thing. Older scripts do not always rely/expect `devtools::load_all()` to have been called.

cjyetman:
There are also multiple places in external repos that source this file directly, so they would lose access to `add_fin_data()`, e.g. https://github.com/search?q=org%3A2DegreesInvesting+0_portfolio_input_check_functions.R&type=code

jdhoffa:
Yup that makes sense. What we can do is a `ctrl + shift + F` for all instances of `add_fin_data()`, and replace with `PACTA.analysis::add_fin_data()`. 

For the offline/old/blah scripts, this would just require that the user has `PACTA.analysis` installed. 

We could then try to notify whoever might still be using them (e.g. MFM), that they just need to add the step of installing the `main` version of `PACTA.analysis`.

jdhoffa:
Just read your second comment. Yes this would also be an issue. We would basically be adding `PACTA.analysis` as a hard dependency (which I personally feel more or less ok with, since the repo itself is `PACTA_analysis`), but I recognize the decision should be considered more carefully.

jdhoffa:
Are we opposed to adding a `devtools::load_all()` at the beginning of whichever script(s) call `add_fin_data()`? 
This would add `devtools` as a hard dependency to our root-level scripts, but I think the benefits would be pretty big, and we could start to more easily move things into `R/` (at least some particular functions that we want to test more extensively)

cjyetman:
If I do my own cost-benefit analysis of the two options, I definitely lean towards...
- don't break anything, but finally get some of these things tested

versus...
- definitely break a bunch of things, here, in other repos, and also potentially in code that is not in any of our repos and we don't even know about, but get this function into a different directory so that it looks more like a package

I would support your suggested strategy if it were done in a broad, very-well considered, coordinated effort to make this type of change universally, with clear documentation about when and where the changes were made, their consequences, and how to work around them if your code breaks, but even then I would expect a serious cost-benefit analysis to be part of "well-considered".

Believe me, I find this utterly frustrating as well, but making safe incremental steps is a smaller pill to swallow.

jdhoffa:
Yup, I totally get that. I'm also just wondering if there is a way to get the best of both worlds.

cjyetman:
> Yup, I totally get that. I'm also just wondering if there is a way to get the best of both worlds.

At first I was thinking that maybe the in-between would be to add `devtools::load_all()` to the top of `0_portfolio_input_check_functions.R` that way anyone that sources it will necessarily have run `devtools::load_all()`, but that doesn't work when you're in another repo. 🤔 

Maybe we could add `library(PACTA.analysis)` to the top of `0_portfolio_input_check_functions.R`, and then anyone would error out early and get a message that the library is not installed and hopefully figure out how to install it. Ironically, I would have to install the "PACTA.analysis" package for the first time ever. I'm not sure that's super realistic for every user since they would have to use the build tools. Maybe we could use `require()` and if the package is not installed give a message saying:

> PACTA.analysis package is not installed. please run `devtools::install_github("2DegreesInvesting/PACTA_analysis")`

cjyetman:
probably good to get @AlexAxthelm 's take on this

jdhoffa:
> At first I was thinking that maybe the in-between would be to add `devtools::load_all()` to the top of `0_portfolio_input_check_functions.R` that way anyone that sources it will necessarily have run `devtools::load_all()`, but that doesn't work when you're in another repo. 🤔

This is exactly what I was thinking, and I hadn't considered that detail.

jdhoffa:
> Maybe we could add `library(PACTA.analysis)` to the top of `0_portfolio_input_check_functions.R`, and then anyone would error out early and get a message that the library is not installed and hopefully figure out how to install it. Ironically, I would have to install the "PACTA.analysis" package for the first time ever. I'm not sure that's super realistic for every user since they would have to use the build tools. Maybe we could use `require()` and if the package is not installed give a message saying:
> 
> > PACTA.analysis package is not installed. please run `devtools::install_github("2DegreesInvesting/PACTA_analysis")`

This to me sounds like a viable possibility. And definitely, we will need to get @AlexAxthelm's sign-off.

AlexAxthelm:
I'm liking the discussion, and the options you're coming up with, but what do we do for folks who can't run `install_GitHub`?

cjyetman:
> I'm liking the discussion, and the options you're coming up with, but what do we do for folks who can't run `install_GitHub`?

To be honest, they're kinda lost already because half of PACTA is in package form, and other than sourcing every file in the `R/` manually (maybe that's not such a terrible idea?), we're already locked in to using `devtools::load_all()` (this was always one of my fears of making PACTA a pseudo-package but not diving all the way in).

cjyetman:
Using `library(PACTA.analysis)` will also require the Docker image for transitionmonitor.com to have the pseudo PACTA.analysis package installed, which it doesn't. 😵‍💫

jdhoffa:
> Using `library(PACTA.analysis)` will also require the Docker image for transitionmonitor.com to have the pseudo PACTA.analysis package installed, which it doesn't. 😵‍💫

I imagine this part is very easily solvable? since Docker already has a fresh clone of the repo AND has to have `devtools` installed anyway.

cjyetman:
> > Using `library(PACTA.analysis)` will also require the Docker image for transitionmonitor.com to have the pseudo PACTA.analysis package installed, which it doesn't. 😵‍💫
> 
> I imagine this part is very easily solvable? since Docker already has a fresh clone of the repo AND has to have `devtools` installed anyway.

yes, but running `library(PACTA.analysis)` and running `devtools::load_all()` are two alternate solutions, each with their own set of unfortunate consequences

jdhoffa:
Yes i wasn't suggesting `devtools::load_all()`, I was suggesting `devtools::install()`. If TM already has `devtools` as a dependency, we should be able to build that package in docker. (And if we're prudent that all R CMD checks pass, this should be relatively safe)

cjyetman:
This is what I mean about doing a cost-benefit analysis.... jumping through all these hoops, and potentially causing some stuff to fail, just so that we can test these functions in one way versus testing them in some other way doesn't make sense to me.

cjyetman:
> > Yes i wasn't suggesting `devtools::load_all()`, I was suggesting `devtools::install()`. If TM already has `devtools` as a dependency, we should be able to build that package in docker. (And if we're prudent that all R CMD checks pass, this should be relatively safe)
> 
> but then that won't work for when this code is called from other repos

Maybe instead we could make the Docker build process always run `devtools::install()` so that we know a fresh up-to-date version of `{PACTA.analysis}` is installed, and then use the `library(PACTA.analysis)` solution?

jdhoffa:
> > Yes i wasn't suggesting `devtools::load_all()`, I was suggesting `devtools::install()`. If TM already has `devtools` as a dependency, we should be able to build that package in docker. (And if we're prudent that all R CMD checks pass, this should be relatively safe)
> 
> but then that won't work for when this code is called from other repos

I'm a little confused by this? 

For TM, in the dockerfile we can just add `devtools::install()` to build the package
For downstream repo users, we ask them to install the package however it would be easiest for them?

jdhoffa:
> > > Yes i wasn't suggesting `devtools::load_all()`, I was suggesting `devtools::install()`. If TM already has `devtools` as a dependency, we should be able to build that package in docker. (And if we're prudent that all R CMD checks pass, this should be relatively safe)
> > 
> > 
> > but then that won't work for when this code is called from other repos
> 
> Maybe instead we could make the Docker build process always run `devtools::install()` so that we know a fresh up-to-date version of `{PACTA.analysis}` is installed, and then use the `library(PACTA.analysis)` solution?

ya exactly, this is what i was saying.

cjyetman:
relevant conversation related to relying on `devtools::load_all()` in #466

jdhoffa:
Also a relevant conversation regarding us being more transparent about dependencies!

cjyetman:
> Also a relevant conversation regarding us being more transparent about dependencies!

By that you mean that it's difficult for one to know what packages are needed? There are quite a number of rather easy ways of determining that.

If we did have a complete list, where could we put it that would satisfy your definition of "more transparent"? Presumably, that should be a very easy thing to do.

We could also reduce these dependencies quite a lot I believe...
```r
sort(unique(renv::dependencies()$Package))
Finding R package dependencies ... Done!
 [1] "assertthat"     "base"           "cli"           
 [4] "config"         "conflicted"     "devtools"      
 [7] "dplyr"          "DT"             "fs"            
[10] "fst"            "ggplot2"        "glue"          
[13] "here"           "janitor"        "jsonlite"      
[16] "knitr"          "magrittr"       "PACTA.analysis"
[19] "plyr"           "purrr"          "R"             
[22] "r2dii.utils"    "readr"          "renv"          
[25] "rlang"          "rmarkdown"      "rstudioapi"    
[28] "snakecase"      "testthat"       "tibble"        
[31] "tidyr"          "tidyverse"      "tools"         
[34] "utils"          "withr"          "yaml"    
```

cjyetman:
https://github.com/2DegreesInvesting/PACTA_analysis#software-requirements--r-environment
```
Github software

For example GitHub Desktop https://desktop.github.com/

Integrated Development Environment (IDE):f

RStudio preferably version 1.1.456 https://rstudio.com/products/rstudio/

R Environment:

Ensure you have the required libraries installed on your computer.

Further, the evaluation will require a bunch of package. It is recommanded install the package during the first run of the code. (assess approx. 30 minutes for the installation process during the first usage of the code).

Necessary packages are: tidyr, dplyr, scales, reshape2, tidyverse, readxl, tidyselect (and dependencies)

For report generation:

MiKTeX - a distribution of the LaTeX typesetting system ; includes TeXworks preferably version 2.9

Further R packages needed: grid, ggplot2, ggthemes, dplyr, reshape2, gridExtra, scales, stringr, extrafont, tidyr, knitr, RColorBrewer, matrixStats, rworldmap, ggmap, cowplot, ggrepel, readxl, tidyverse, ggforce, sitools , countrycode

```

cjyetman:
probably the most minimal but complete set of dependencies...
```r
base_r <- c("R", "base", "utils", "tools", "stats")
self <- "PACTA.analysis"
tidyverse_imports <- tidyverse::tidyverse_packages(include_self = FALSE)
repo_deps <- sort(unique(renv::dependencies()$Package))
repo_deps <- repo_deps[!repo_deps %in% c(base_r, self, tidyverse_imports)]
cat(paste(repo_deps, collapse = ", "))
#> assertthat, config, conflicted, devtools, DT, fs, fst, glue, here, janitor, 
#> knitr, plyr, r2dii.utils, renv, rmarkdown, snakecase, testthat, tidyverse, 
#> withr, yaml
```

jdhoffa:
I just mean, in the issue you posted, the issue was that they didn't have `devtools` installed. And there wasn't an easy way for the user to understand what these packages are, or how to install them easily. 

(Obviously as an experience R user, you know how to figure it out, but we could make it way more explicit/ obvious lol)

cjyetman:
ugh, this is never going to work with R CMD check because it will never allow these scripts in the root directory, even if I remove them from being explicitly disallowed in `.Rbuildignore`.

Maybe we should remove the R CMD check in CI since it's not testing anything worthwhile that a `testthat` test is/can not test?

jdhoffa:
I like having R CMD Check around (for example it was what caught [this](https://github.com/2DegreesInvesting/PACTA_analysis/pull/519) which `testthat` obviously would not have caught). 

I think it's clear by now, but my opinion is the more we stray away from the infrastructure that exists to catch things we may not think of, the more we may eventually have to think of our own way to re-write that infrastructure, which feels like a waste of time. But I won't block you, you can do whatever you think is best.

jdhoffa:
> ugh, this is never going to work with R CMD check because it will never allow these scripts in the root directory, even if I remove them from being explicitly disallowed in `.Rbuildignore`.
> 
> Maybe we should remove the R CMD check in CI since it's not testing anything worthwhile that a `testthat` test is/can not test?

I think it checks some worthwhile things, for example, that files are UTF-8 encoded, that files have platform-agnostic names (ie. all file names will function across windows, linux, mac), non-ASCII characters popping up in code, no binary files etc. That said, these checks only really apply cleanly to code that is in the `R/` folder... I bring up the UTF-8 encoding one in particular because that is a hell you went through personally and wrote your own custom action to check it, it's something we get for free in the R CMD check + R/ folder world (obviously it doesn't work for code the is `.Rbuildignore`d in the root)

My honest opinion is that I still prefer the solution I proposed already. It also allows us to use `testthat` in a more standard way and avoid other potential workarounds we might need to think up.

cjyetman:
> that files are UTF-8 encoded
there's already a [test for that](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/.github/workflows/pr-check-file-encodings.yml)

> non-ASCII characters popping up in code
we need these frequently

> no binary files
we also need these sometimes

> My honest opinion is that I still prefer the solution I proposed already. It also allows us to use `testthat` in a more standard way and avoid other potential workarounds we might need to think up.

I still disagree. There are consequences to forcing everything to be an R package. Pros **and** cons. Having done so has substantially stagnated development in this repo. There's numerous open issues related to things that only exist in this repo because we forced it to be an R package. R package development has always been a bit of a mess. `tidyverse` didn't change that or make it better, it just abstracts a bunch of the mess so it's "easier" to deal with, but it doesn't clean up the mess.

cjyetman:
#527 demonstrates how we can easily switch to using `testthat` for unit testing **and** be able to test all of the important code in our repo, regardless if it is in the `R/` directory or otherwise complies with R package structure. For me, that's a very big win, and I'll finally be able to test critical code that lives in the root directory without having to move it around and break other code.

jdhoffa:
> > that files are UTF-8 encoded
> 
> there's already a [test for that](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/.github/workflows/pr-check-file-encodings.yml)

Of course, my point here is that there are useful things that come baked in with R CMD check, that could eventually remove some of the infrastructure we have to maintain for this repo. 

> 
> > non-ASCII characters popping up in code
> 
> we need these frequently

in `create_interactive_report`, sure, but hopefully not in `PACTA.analysis` itself, or?

> 
> > no binary files
> 
> we also need these sometimes

I didn't realize that, where do we use these? That could make running a "local PACTA" quite difficult/ could pose a security hazard. 

> 
> > My honest opinion is that I still prefer the solution I proposed already. It also allows us to use `testthat` in a more standard way and avoid other potential workarounds we might need to think up.
> 
> I still disagree. There are consequences to forcing everything to be an R package. Pros **and** cons. Having done so has substantially stagnated development in this repo. There's numerous open issues related to things that only exist in this repo because we forced it to be an R package. R package development has always been a bit of a mess. `tidyverse` didn't change that or make it better, it just abstracts a bunch of the mess so it's "easier" to deal with, but it doesn't clean up the mess.

Fair enough. We disagree with one another, it is what it is. I'm just giving my honest review. I totally agree with the Pros **and** cons point, but each time I see a Pro, and extra work necessary to achieve the same thing, I have a hard time not pointing it out. I think "Having done so has substantially stagnated development in this repo" is not strictly true. To me, the stagnation seems to be coming from not committing to one or the other, and that comes from a lack of a clear vision about what this project is and what it should be. Without that, each of us is earnestly trying to clean it up, and we're all pulling it in different directions which creates a mess. 

> R package development has always been a bit of a mess

Not really relevant to this discussion, but that hasn't been my experience.

cjyetman:
> Of course, my point here is that there are useful things that come baked in with R CMD check, that could eventually remove some of the infrastructure we have to maintain for this repo.

I have demonstrated that things that R CMD check do can easily be done otherwise, and in some case with more/better features, e.g. testing code that lives outside the `R/` directory.

> > > non-ASCII characters popping up in code
> > 
> > we need these frequently
> 
> in `create_interactive_report`, sure, but hopefully not in `PACTA.analysis` itself, or?

a quick check shows 8 files in this repo that contain non-ASCII characters

> > > no binary files
> > 
> > we also need these sometimes
> 
> I didn't realize that, where do we use these? That could make running a "local PACTA" quite difficult/ could pose a security hazard.

RDA and RDS files are binary

> I think "Having done so has substantially stagnated development in this repo" is not strictly true. To me, the stagnation seems to be coming from not committing to one or the other, and that comes from a lack of a clear vision about what this project is and what it should be. Without that, each of us is earnestly trying to clean it up, and we're all pulling it in different directions which creates a mess.

No offence to anyone that has contributed to this repo, but for the past 6 months I have worked on this repo [substantially more](https://github.com/2DegreesInvesting/PACTA_analysis/graphs/contributors) than anyone else and I am the de facto maintainer. I acknowledge that it may not impact everyone in the same way, but it does impact me, and therefore it has stagnated progress in this repo.

> > R package development has always been a bit of a mess
> 
> Not really relevant to this discussion, but that hasn't been my experience.

R package development imposes rather strict, often unnecessary, and often quite unintuitive rules on a repo.

jdhoffa:
> > Of course, my point here is that there are useful things that come baked in with R CMD check, that could eventually remove some of the infrastructure we have to maintain for this repo.
> 
> I have demonstrated that things that R CMD check do can easily be done otherwise, and in some case with more/better features, e.g. testing code that lives outside the `R/` directory.
> 
> > > > non-ASCII characters popping up in code
> > > 
> > > 
> > > we need these frequently
> > 
> > 
> > in `create_interactive_report`, sure, but hopefully not in `PACTA.analysis` itself, or?
> 
> a quick check shows 8 files in this repo that contain non-ASCII characters
> 
> > > > no binary files
> > > 
> > > 
> > > we also need these sometimes
> > 
> > 
> > I didn't realize that, where do we use these? That could make running a "local PACTA" quite difficult/ could pose a security hazard.
> 
> RDA and RDS files are binary
> 
> > I think "Having done so has substantially stagnated development in this repo" is not strictly true. To me, the stagnation seems to be coming from not committing to one or the other, and that comes from a lack of a clear vision about what this project is and what it should be. Without that, each of us is earnestly trying to clean it up, and we're all pulling it in different directions which creates a mess.
> 
> No offence to anyone that has contributed to this repo, but for the past 6 months I have worked on this repo [substantially more](https://github.com/2DegreesInvesting/PACTA_analysis/graphs/contributors) than anyone else and I am the de facto maintainer. I acknowledge that it may not impact everyone in the same way, but it does impact me, and therefore it has stagnated progress in this repo.
> 
> > > R package development has always been a bit of a mess
> > 
> > 
> > Not really relevant to this discussion, but that hasn't been my experience.
> 
> R package development imposes rather strict, often unnecessary, and often quite unintuitive rules on a repo.

I obviously have thoughts on all these points, but I'm hesitant to derail this PR further. My opinion is what I explained above, but it is just an opinion. As the de-facto maintainer, you are of course free to merge whatever you like.

cjyetman:
After a whole lot of experimentation, I finally came up with a solution...

When `R CMD` runs, it sets an environment variable `R_CMD` to the path to the `Rcmd` binary. This `R_CMD` environment variable should never exist otherwise. So if I skip the tests that need to access a script in the root level of the repo only when it runs during `R CMD`, then I can still run those tests when using e.g. `devtools::test()` or `testthat::test_local()`, but still be able to run the rest of the tests when running e.g. `R CMD build` or `dev tools::check()` without any unnecessary errors.

So adding this to my tests allows us to have and use both testing infrastructures working simultaneously... 
```r
skip_if(nzchar(Sys.getenv("R_CMD")))
```

Does that seem reasonable @jdhoffa @AlexAxthelm?

AlexAxthelm (approved):

PR #516, opened 2021-11-22
Add a script for testing, can use local repos

AlexAxthelm:
This script runs the container like Constructiva does, but allows for
the uses of local repos (using flags to specify which repo is which).
not particularly elegant, but I've found it usesful in my testing work
so far.

### DIFF
--- a/transitionmonitor_docker/run-like-constructiva-flags.sh
+++ b/transitionmonitor_docker/run-like-constructiva-flags.sh
@@ -61,7 +61,8 @@ if [ -n "${verbose}" ]; then
 fi
 
 if [ -z "${docker_command}" ]; then
-  docker_command="/bound/bin/run-r-scripts \"$portfolioIdentifier\""
+  docker_command="/bound/bin/run-r-scripts"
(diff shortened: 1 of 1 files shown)


AlexAxthelm:
@cjyetman This is an extension of your `run-like-wim.sh`, but uses flags, rather than positional arguments, and then allows the user to also override the repos in the container with local ones (using `--mount`s if needed)

cjyetman (approved):
cool! have you looked at #498 recently... as far as I know, that's the most up-to-date, most correct version that we have of what Constructive actually does

AlexAxthelm:
It's not working perfectly yet.

AlexAxthelm:
I hate parsing arguments in bash

cjyetman:
glad it's not me LOL

AlexAxthelm:
I've got it working great, except that when I do the last pass to `docker run`, it doesn't pick up the last argument as something that should be passed to the command

AlexAxthelm:
A change that may be worth making is taher than having to do `-r bash -i`, jusst have `-i` imply that we want to run bash

PR #517, opened 2021-11-23
remove all the confusing extra example files

cjyetman:
All the extra example files under `working_dir/` add more confusion than any help. This removes all of them but the 2 files that are essential for the "default" example to work (based on `TestPortfolio_Input` being explicitly [set as the default here](https://github.com/2DegreesInvesting/PACTA_analysis/blob/0936c1d71462443179a875156b587b7fba795b0a/web_tool_script_1.R#L12) and in the other web tool scripts). The existing [.gitignore](https://github.com/2DegreesInvesting/PACTA_analysis/blob/0936c1d71462443179a875156b587b7fba795b0a/.gitignore#L36-L43) file already ignores any new files that might be added to those directories so they cannot be committed by accident. The directory structure underneath `working_dir/` is required, so those sub-directories are kept/protected by containing a `.gitkeep.txt` file, like[ this one](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/working_dir/00_Log_Files/.gitkeep.txt). The only two files left then under `working_dir/` are those that are essential for PACTA to run:
[working_dir/10_Parameter_File/TestPortfolio_Input_PortfolioParameters.yml](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/working_dir/10_Parameter_File/TestPortfolio_Input_PortfolioParameters.yml)
[working_dir/20_Raw_Inputs/TestPortfolio_Input.csv](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/working_dir/20_Raw_Inputs/TestPortfolio_Input.csv)

### DIFF
--- a/working_dir/10_Parameter_File/0_Indices_Input_PortfolioParameters.yml
+++ b/working_dir/10_Parameter_File/0_Indices_Input_PortfolioParameters.yml
@@ -1,5 +1,0 @@
-default:
-    parameters:
-        portfolio_name_in: test_portfolio
-        investor_name_in: Test
-
--- a/working_dir/10_Parameter_File/BondsTest_Input_PortfolioParameters.yml
+++ b/working_dir/10_Parameter_File/BondsTest_Input_PortfolioParameters.yml
@@ -1,5 +1,0 @@
-default:
-    parameters:
-        portfolio_name_in: test_portfolio
-        investor_name_in: Test
-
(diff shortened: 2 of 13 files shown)


jdhoffa (approved):
YAY

jdhoffa:
Can we also consider removing the `documentation` folder at some point?

cjyetman:
why remove documentation?

PR #518, opened 2021-11-23
Add scenario benchmark calculation to `calculate_tdm()` function

jdhoffa:
This PR adds a new calculation to `calculate_tdm()` which effectively calculates the Transition Disruption Metric of the scenario against itself. 

This can be used as a "scenario benchmark". 

Functionally, the output gains one column, `tdm_metric`, which indicates if the values stored in `tdm_*_value` correspond to the `portfolio` or to the `scenario`). 

FYI. @MonikaFu


NAMESPACE
```diff
@@ -39,5 +39,6 @@ importFrom(rlang,abort)
 importFrom(rlang,inform)
 importFrom(rlang,warn)
 importFrom(tibble,tibble)
+importFrom(tidyr,pivot_longer)
```
jdhoffa, line 42: I chose to export this, since `pivot_wider` is already exported, and the names are pretty unique to `tidyr`. Also happy to go the other direction, and explicitly namespace all `pivot_*`s

R/calculate_tdm.R
```diff
@@ -20,10 +20,12 @@
...
 #' @return A tibble with the columns specified in the `additional_groups` input
-#'   as well as `tdm_technology_value`: the technology level transition disruption
-#'   metric, `tdm_sector_value`: the sector level transition disruption metric,
-#'   `tdm_portfolio_value`: the portfolio level transition disruption metric, as well
-#'   as `tdm_t0`, `tdm_delta_t1` and `tdm_delta_t2`: corresponding to the input
+#'   as well as `tdm_technology_value`: the technology level transition
+#'   disruption metric, `tdm_sector_value`: the sector level transition
+#'   disruption metric, `tdm_portfolio_value`: the portfolio level transition
+#'   disruption metric, `tdm_metric`: a flag that indicates what each `*_value`
+#'   is being calculated against (e.g. "portfolio" or "scenario"),  as well as
```
jdhoffa, line 27: I know the wording here is a bit awkward, if you can think of a clearer way to explain the purpose of that column, that would be cool.

R/calculate_tdm.R
```diff
@@ -188,23 +190,37 @@ add_tdm <- function(data, groups) {
...
+        max(0, (.data$.numerator_portfolio / .data$.denominator) * .data$monotonic_factor) * 2
       ),
-      .numerator = NULL,
+      tdm_technology_value_scenario = ifelse(
+        .data$.denominator == 0,
+        0,
+        max(0, (.data$.numerator_scenario / .data$.denominator) * .data$monotonic_factor) * 2
+      ),
+      .numerator_portfolio = NULL,
+      .numerator_scenario = NULL,
```
jdhoffa, line 207: This just doubles up the calculation in a DRY-ish way. `denominator` is the same for both, only the `numerator` changes between `portfolio` and `scenario`.

tests/testthat/test-calculate_tdm.R
```diff
@@ -51,9 +60,11 @@ test_that("outputs the expected names", {
 test_that("outputs 0 for `tdm_technology_value` and `tdm_sector_value`, if `scen_alloc_wt_tech_prod
           doesn't change over 10 year period`", {
   data <- fake_tdm_data(year = c(2020, 2025, 2030))
-  out <- calculate_tdm(data, 2020)
-  expect_true(out$tdm_technology_value == 0)
-  expect_true(out$tdm_sector_value == 0)
+  out <- calculate_tdm(data, 2020) %>%
+    split(.$tdm_metric)
```
jdhoffa, line 64: Most tests had to gain a line similar to `split(.$tdm_metric)` to account for the change in the structure of the output tibble.

R/calculate_tdm.R
cjyetman: should it say

> what each `tdm_*_value` is being calculated against

?

cjyetman:
out of curiosity (and please excuse my ignorance).... I thought that the TDM was defined as matching the scenario at a specific number (2?), was that changed, or just my misunderstanding?

R/calculate_tdm.R
```diff
@@ -188,23 +190,37 @@ add_tdm <- function(data, groups) {
...
       .denominator = NULL
     ) %>%
-    select(.data$tdm_technology_value, all_of(groups)) %>%
+    select(.data$tdm_technology_value_portfolio, .data$tdm_technology_value_scenario, all_of(groups)) %>%
     distinct() %>%
+    pivot_longer(
+      cols = tidyr::starts_with("tdm_technology_value"),
```
cjyetman, line 213: should this be (with a "_" at the end?
```r
cols = tidyr::starts_with("tdm_technology_value_"),
```

cjyetman: never mind, I see you used `names_prefix` below

cjyetman: actually that doesn't even matter either.... I just would have been slightly more specific in my `starts_with` selection, but no worries

R/imports.R
```diff
@@ -3,14 +3,15 @@
...
-#' @importFrom tidyr pivot_wider unnest
+#' @importFrom tidyr pivot_wider pivot_longer unnest
 #' @importFrom rlang abort warn inform %||%
 #' @importFrom fs path dir_create dir_delete dir_ls
 NULL
 
 globalVariables(
   c(
     ".data",
+    ".env",
```
cjyetman, line 14: 👍🏻

R/calculate_tdm.R
jdhoffa: 👍

cjyetman (commented):
I was surprised to not see changes to `fake_tdm_data()`... did that adapt to this new struture with no problems? does it give the same values for `tdm_metric == portfolio` and `tdm_metric == scenario` rows?

jdhoffa:
`fake_tdm_data()` is input data! the input is the same, just the output changes.

jdhoffa:
> out of curiosity (and please excuse my ignorance).... I thought that the TDM was defined as matching the scenario at a specific number (2?), was that changed, or just my misunderstanding?

I don't fully understand this, but I THINK you're referring to the factor of 2 in the metric calculation? 

This is just what t2/(t2-t1) reduces to if t2 = 10 and t1 = 5. 

Or are you talking about something else?

jdhoffa:
> I was surprised to not see changes to `fake_tdm_data()`... did that adapt to this new struture with no problems? does it give the same values for `tdm_metric == portfolio` and `tdm_metric == scenario` rows?

In terms of output, I had to change the tests a bit. Mostly, I had to add either `filter(tdm_metric == "scenario")` or `split(.$tdm_metric)`, to pick out the appropriate `portfolio` values to test against. 

Also note: the test coverage against `scenario` values is still pretty low. Only one test actually touches those values (`"outputs the expected tibble"`)

cjyetman:
> > I was surprised to not see changes to `fake_tdm_data()`... did that adapt to this new struture with no problems? does it give the same values for `tdm_metric == portfolio` and `tdm_metric == scenario` rows?
> 
> In terms of output, I had to change the tests a bit. Mostly, I had to add either `filter(tdm_metric == "scenario")` or `split(.$tdm_metric)`, to pick out the appropriate `portfolio` values to test against.
> 
> Also note: the test coverage against `scenario` values is still pretty low. Only one test actually touches those values (`"outputs the expected tibble"`)

yeah, this is what I was questioning.... how do you know, e.g. `expect_equal(out$scenario$tdm_technology_value, 1L)`?

jdhoffa:
> > > I was surprised to not see changes to `fake_tdm_data()`... did that adapt to this new struture with no problems? does it give the same values for `tdm_metric == portfolio` and `tdm_metric == scenario` rows?
> > 
> > 
> > In terms of output, I had to change the tests a bit. Mostly, I had to add either `filter(tdm_metric == "scenario")` or `split(.$tdm_metric)`, to pick out the appropriate `portfolio` values to test against.
> > Also note: the test coverage against `scenario` values is still pretty low. Only one test actually touches those values (`"outputs the expected tibble"`)
> 
> yeah, this is what I was questioning.... how do you know, e.g. `expect_equal(out$scenario$tdm_technology_value, 1L)`?

Ya that particular value I just calculated by hand. In that test, the dataset has scen(2020) = 1, scen(2025) = 2 and scen(2030) = 3. So: 
(3-2/ 3-1) * 2 = (1/2) * 2 = 1

jdhoffa:
> > out of curiosity (and please excuse my ignorance).... I thought that the TDM was defined as matching the scenario at a specific number (2?), was that changed, or just my misunderstanding?
> 
> I don't fully understand this, but I THINK you're referring to the factor of 2 in the metric calculation?
> 
> This is just what t2/(t2-t1) reduces to if t2 = 10 and t1 = 5.
> 
> Or are you talking about something else?

Oh fuck! I didn't even do this (even though I said I did). That's a bug, going to fix now

cjyetman:
> > out of curiosity (and please excuse my ignorance).... I thought that the TDM was defined as matching the scenario at a specific number (2?), was that changed, or just my misunderstanding?
> 
> I don't fully understand this, but I THINK you're referring to the factor of 2 in the metric calculation?
> 
> This is just what t2/(t2-t1) reduces to if t2 = 10 and t1 = 5.
> 
> Or are you talking about something else?

🤷🏻 I think I had it in my mind that...
- if your portfolio's TDM value was 0, then you expected little to no disruption
- if your portfolio's TDM value was 2, then you expected the same disruption IPR's scenario predicted
- if your portfolio's TDM value was >2, then you expected even more disruption than IPR's scenario predicted

or something like that.... I didn't realize the benchmark moves... but then the color scale stays the same. Obviously I don't really understand this metric.

cjyetman:
this depends on some change in `create_interactive_report()`, otherwise it breaks, correct?

jdhoffa:
> > > out of curiosity (and please excuse my ignorance).... I thought that the TDM was defined as matching the scenario at a specific number (2?), was that changed, or just my misunderstanding?
> > 
> > 
> > I don't fully understand this, but I THINK you're referring to the factor of 2 in the metric calculation?
> > This is just what t2/(t2-t1) reduces to if t2 = 10 and t1 = 5.
> > Or are you talking about something else?
> 
> 🤷🏻 I think I had it in my mind that...
> 
> * if your portfolio's TDM value was 0, then you expected little to no disruption
> * if your portfolio's TDM value was 2, then you expected the same disruption IPR's scenario predicted
> * if your portfolio's TDM value was >2, then you expected even more disruption than IPR's scenario predicted
> 
> or something like that.... I didn't realize the benchmark moves... but then the color scale stays the same. Obviously I don't really understand this metric.

Ah yes, this was just incorrect. I went through the expected values and interpretations of different value ranges in detail in the `vignettes` article.

jdhoffa:
> this depends on some change in `create_interactive_report()`, otherwise it breaks, correct?

Yes I believe so.

cjyetman:
> > this depends on some change in `create_interactive_report()`, otherwise it breaks, correct?
> 
> Yes I believe so.

so technically this should not be merged until we have the infrastructure on `create_interactive_report()` ready to accept it... just verified that this will break the current charts

jdhoffa:
> > > this depends on some change in `create_interactive_report()`, otherwise it breaks, correct?
> > 
> > 
> > Yes I believe so.
> 
> so technically this should not be merged until we have the infrastructure on `create_interactive_report()` ready to accept it... just verified that this will break the current charts

It's sorta hard to say what should come first no? 

If we change `create_interactive_report` first, it will be expecting a dataset that doesn't exist yet?

jdhoffa:
and also, since we haven't built a new docker image/ bumped the version number, the last version should still be stable? I think its potato vs potato over what gets merged first.

R/calculate_tdm.R
```diff
@@ -184,27 +186,43 @@ add_monotonic_factor <- function(data, t0, delta_t1, delta_t2, groups) {
...
+      .numerator_scenario = .data$scenario_production_plus_delta_t2 - .data$scenario_production_plus_delta_t1,
       .denominator = .data$scenario_production_plus_delta_t2 - .data$scenario_production_t0,
-      # TODO: This case we
-      tdm_technology_value = ifelse(
+      .time_factor = (delta_t2 / (delta_t2 - delta_t1)),
+      tdm_technology_value_portfolio = ifelse(
         .data$.denominator == 0,
         0,
-        max(0, (.data$.numerator / .data$.denominator) * .data$monotonic_factor) * 2
+        max(0, (.data$.numerator_portfolio / .data$.denominator) * .data$monotonic_factor) * .time_factor
```
cjyetman, line 200: `.data$time_factor`?

R/calculate_tdm.R
```diff
@@ -184,27 +186,43 @@ add_monotonic_factor <- function(data, t0, delta_t1, delta_t2, groups) {
...
       ),
-      .numerator = NULL,
-      .denominator = NULL
+      tdm_technology_value_scenario = ifelse(
+        .data$.denominator == 0,
+        0,
+        max(0, (.data$.numerator_scenario / .data$.denominator) * .data$monotonic_factor) * .time_factor
```
cjyetman, line 205: `.data$time_factor`?

cjyetman:
> > so technically this should not be merged until we have the infrastructure on `create_interactive_report()` ready to accept it... just verified that this will break the current charts
> 
> It's sorta hard to say what should come first no?
> 
> If we change `create_interactive_report` first, it will be expecting a dataset that doesn't exist yet?

actually, you're write... this should come first because `create_interactive_report()` is designed to not error out but to just show a message if this data is not available.... and it's harder to work on `create_interactive_report()` when you don't know what the output from PACTA is going to be

cjyetman (approved):
cool, works for me

PR #519, opened 2021-11-24
Remove global env variables from `tdm_utility_functions`

jdhoffa:
I noticed (when running R CMD Check) that some of the functions in `tdm_utility_functions.R` depend on variables in the global environment (ie. variables that are not explicitly defined as function arguments). 

I know this is how a lot of the legacy PACTA code runs, but i think we should be prudent that all new code doesn't rely on this.

### DIFF
--- a/R/tdm_utility_functions.R
+++ b/R/tdm_utility_functions.R
@@ -6,15 +6,15 @@ analysis_inputs_with_tdm_scenarios <- function() {
   c("../pacta-data/2020Q4")
 }
 
-data_includes_tdm_scenarios <- function() {
+data_includes_tdm_scenarios <- function(analysis_inputs_path) {
   analysis_inputs_path %in% analysis_inputs_with_tdm_scenarios()
 }
 
-tdm_conditions_met <- function() {
-  project_code == "GENERAL" && data_includes_tdm_scenarios()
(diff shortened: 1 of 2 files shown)


cjyetman (dismissed):
definitely appreciate that strategy 👍🏻

cjyetman:
test fails?

cjyetman (commented):
reverting approval until we figure out why the test fails

jdhoffa:
Very confused as to why the check is failing. 

Are these functions called in other repos?

cjyetman:
> Are these functions called in other repos?

yup.... https://github.com/2DegreesInvesting/create_interactive_report/blob/af5a1c91659da022aa0c086b29631f37bb0a67ba/general_en_template/index.Rmd#L45

jdhoffa:
Should be solved now in 2DegreesInvesting/create_interactive_report#464
Unfortunately, the check here will still fail until that PR is merged (and I think actually this PR should be merged first, followed by the downstream PR). 

But you can check out locally and see that it works.

cjyetman (approved):
^agreed

lgtm

cjyetman:
re-ran tests and they all pass now the 2DegreesInvesting/create_interactive_report#464 is merged

PR #520, opened 2021-11-24
Always evaluate tdm conditions as unmet

AlexAxthelm:
this is a temporary workaround to allow a new docker image to be built
and deployed, without rolling out TDM to users prematurely.


cjyetman (approved):
sure, why not

PR #522, opened 2021-11-24
update README with current dependencies

cjyetman:
https://github.com/2DegreesInvesting/PACTA_analysis/pull/514#issuecomment-977877197

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,9 @@ Ensure you have the required libraries installed on your computer.
 Further, the evaluation will require a bunch of package. It is recommanded install the package during the first run of the code. (assess approx. 30 minutes for the installation process during the first usage of the code).
 
 Necessary packages are: 
-tidyr, dplyr, scales, reshape2, tidyverse, readxl, tidyselect (and dependencies)
+assertthat, config, conflicted, devtools, DT, fs, fst, glue, here, janitor, 
+knitr, plyr, r2dii.utils, renv, rmarkdown, snakecase, testthat, tidyverse, 
+withr, yaml
     
 **For report generation:**
 


jdhoffa:
So, this LGTM, but could we potentially have the dependencies in a file that's easy to source? Like `imports.R` or something? And then we could have `README.Rmd` that sources that, and shows the dependencies to humans. And also the `imports.R` could be sourced directly by docker when installing everything? 

Just not sure if there's a reason against doing this.

cjyetman:
> Just not sure if there's a reason against doing this.

more infrastructure and dependencies to maintain and worry about?

jdhoffa:
But how is that any more infrastructure than making sure this `README` is up to date...? You would just make sure `imports.R` is up to date? 

And it's a gain because it can be called easily from a script.

cjyetman:
a `README.Rmd` file that uses this code could easily do this programmatically, but then we'd have to either remember to knit it whenever it is edited (or create a new action or CI process to do it automatically), and remember to edit the `README.Rmd` not the `README.md`...
```r
base_r <- c("R", "base", "utils", "tools", "stats")
self <- "PACTA.analysis"
tidyverse_imports <- tidyverse::tidyverse_packages(include_self = FALSE)
repo_deps <- sort(unique(renv::dependencies()$Package))
repo_deps <- repo_deps[!repo_deps %in% c(base_r, self, tidyverse_imports)]
cat(paste(repo_deps, collapse = ", "))
```

personally, I'd rather just update it in one place if it ever changes

jdhoffa:
> a `README.Rmd` file that uses this code could easily do this programmatically, but then we'd have to either remember to knit it whenever it is edited (or create a new action or CI process to do it automatically), and remember to edit the `README.Rmd` not the `README.md`...
> 
> ```r
> base_r <- c("R", "base", "utils", "tools", "stats")
> self <- "PACTA.analysis"
> tidyverse_imports <- tidyverse::tidyverse_packages(include_self = FALSE)
> repo_deps <- sort(unique(renv::dependencies()$Package))
> repo_deps <- repo_deps[!repo_deps %in% c(base_r, self, tidyverse_imports)]
> cat(paste(repo_deps, collapse = ", "))
> ```
> 
> personally, I'd rather just update it in one place if it ever changes

We could also just keep the `README.md` and under dependencies say "dependencies can be seen in `imports.R`". The `.Rmd` wasn't the most important part of what i was getting at

jdhoffa:
> personally, I'd rather just update it in one place if it ever changes

definitely agree with this ;)

cjyetman:
I don't find the structure of having two separate files, both with necessarily very specific names, and then requiring some dependent software running manually or even another very specific file running it automagically somehow easier than just keep the one list up-to-date.

We don't.... or at least we shouldn't be... updating the package dependencies here very often.

cjyetman:
> We could also just keep the `README.md` and under dependencies say "dependencies can be seen in `imports.R`". The `.Rmd` wasn't the most important part of what i was getting at

isn't tucking the dependencies away in code in some file less "transparent" than just putting it in the README?

jdhoffa:
> I don't find the structure of having two separate files, both with necessarily very specific names, and then requiring some dependent software running manually or even another very specific file running it automagically somehow easier than just keep the one list up-to-date.
> 
> We don't.... or at least we shouldn't be... updating the package dependencies here very often.

The benefit is that any new user (be it an actual user, docker, whatever) can easily install dependencies, and it is explained how to do this. The list should definitely only exist in one place. and to me the list should be easily formatted such that a computer can parse it and install dependencies. 

We already have it in at least two places: (`README.md`,  https://github.com/2DegreesInvesting/docker/blob/master/r-packages/Dockerfile), and a new user can not necessarily easily install all the packages listed in either of these locations.

jdhoffa:
> > We could also just keep the `README.md` and under dependencies say "dependencies can be seen in `imports.R`". The `.Rmd` wasn't the most important part of what i was getting at
> 
> isn't tucking the dependencies away in code in some file less "transparent" than just putting it in the README?

I feel like saying "All dependencies are in `imports.R` is pretty transparent... more so than listing the "dependencies" manually in the `README`, and it not being up to date hahaha.

cjyetman:
> I feel like saying "All dependencies are in `imports.R` is pretty transparent... more so than listing the "dependencies" manually in the `README`, and it not being up to date hahaha.

why would it not be up-to-date.... you would update it at the same time anyone would have updated `imports.R`?

weren't we talking about making it more transparent for unexperienced R users? how is putting it in code somewhere more transparent for an unexperienced R user than putting it right up front in the README?

jdhoffa:
(also im obviously not saying it has to be `imports.R` or `DESCRIPTION` or any other "specially named filed". I just feel like it should be in some file that is easy to source by a computer)

jdhoffa:
> > I feel like saying "All dependencies are in `imports.R` is pretty transparent... more so than listing the "dependencies" manually in the `README`, and it not being up to date hahaha.
> 
> why would it not be up-to-date.... you would update it at the same time anyone would have updated `imports.R`?
> 
> weren't we talking about making it more transparent for unexperienced R users? how is putting it in code somewhere more transparent for an unexperienced R user than putting it right up front in the README?

No tests or checks will fail if `README` isn't updated, so it would be incredibly easy to forget to add a dependency, and it would never be caught. If the `docker` sources the same `imports.R` that the `README` references, we will immediately know if we forget to define a dependency

cjyetman:
1. to me any "list" maintained is solely useful for a human user.... any program, Docker build, whatever is better off determining the dependencies programmatically, which is easy to do
2. I don't believe this repo should in any way install packages for a user automatically in their environment... personally I find that super annoying and dangerous.... I would only support that if we did so in a completely repo contained environment, e.g. using `{renv}`

cjyetman:
> No tests or checks will fail if `README` isn't updated, so it would be incredibly easy to forget to add a dependency, and it would never be caught. If the `docker` sources the same `imports.R` that the `README` references, we will immediately know if we forget to define a dependency

that's only true if what you said above is not true.... either you have a lot more infrastructure (scripts, files, test, ci, etc.) or not

jdhoffa:
> > I feel like saying "All dependencies are in `imports.R` is pretty transparent... more so than listing the "dependencies" manually in the `README`, and it not being up to date hahaha.
> 
> why would it not be up-to-date.... you would update it at the same time anyone would have updated `imports.R`?
> 
> weren't we talking about making it more transparent for unexperienced R users? how is putting it in code somewhere more transparent for an unexperienced R user than putting it right up front in the README?

Well, I'm trying to highlight gains in multiple different aspects, so I'm talking about "unexperienced users" and "CI" and "docker" somewhat simultaneously, and trying to find the simplest 1 solution that could satisfy each of these use-cases.

jdhoffa:
> > No tests or checks will fail if `README` isn't updated, so it would be incredibly easy to forget to add a dependency, and it would never be caught. If the `docker` sources the same `imports.R` that the `README` references, we will immediately know if we forget to define a dependency
> 
> that's only true if what you said above is not true.... either you have a lot more infrastructure (scripts, files, test, ci, etc.) or not

How? hahahaha:
* You add the `imports.R` file
* You change the part of the `dockerfile` that manually lists dependencies to source a file (``imports.R`)
* You write a static line in README that references the deps explicitly, and make it say Look at `imports.R`

It actually removes infrastructure, because we don't need to maintain BOTH `README` and `Dockerfile`, we just maintain `imports.R`

AND CI will then automatically check if anything breaks

cjyetman:
you don't consider CI infrastructure?

jdhoffa:
> 1. to me any "list" maintained is solely useful for a human user.... any program, Docker build, whatever is better off determining the dependencies programmatically, which is easy to do
> 2. I don't believe this repo should in any way install packages for a user automatically in their environment... personally I find that super annoying and dangerous.... I would only support that if we did so in a completely repo contained environment, e.g. using `{renv}`

1. Disagree. Explicitly managing dependencies allows us to make sure we don't add a bunch of unnecessary ones
2. Agree, but what I'm saying doesn't do this? It just points the user to the `imports.R` file (and maybe suggests a useful line they can run to install the packages). Still doesn't ever do anything automatically

jdhoffa:
> you don't consider CI infrastructure?

I don't consider it "adding" infrastructure, since all the checks already exist and are being run

jdhoffa:
> > you don't consider CI infrastructure?
> 
> I don't consider it "adding" infrastructure, since all the checks already exist and are being run

This is the check: 
https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/.github/workflows/source-web-tool-scripts.yaml

cjyetman:
I'm not saying it adds infrastructure, I'm saying it requires infrastructure... it requires a complex infrastructure that includes numerous files and systems that all have to work together in a perfect symphony just to put a list of dependencies in the README solely for the purpose of being "more transparent" for some hypothetical unexperienced R users.

jdhoffa:
> I'm not saying it adds infrastructure, I'm saying it requires infrastructure... it requires a complex infrastructure that includes numerous files and systems that all have to work together in a perfect symphony just to put a list of dependencies in the README solely for the purpose of being "more transparent" for some hypothetical unexperienced R users.

Ok, that is fair. Then I would ask two questions: 
* Do you think changing extracting the list of dependencies to one universal file (whatever that is), and having each of the different potential ways people use PACTA all referring to that file adds significant complexity to the already existing infrastructure?
* Do you think the pros of having all of these things defined in one place outweigh the cons of the added complexity above?

cjyetman:
> * Do you think changing the status quo, by extracting the list of dependencies to one universal file (whatever file that may be), and having each of the different potential ways people use PACTA (users, docker, web tool) all referring to that file adds significant complexity to the already existing infrastructure?

Yes... if the goal is to make this "more transparent" for some hypothetical unexperienced R users, then I believe it is more complex to try to squeeze that feature out of a single file that is also meant to feed programmatic tools... it's trying to squeeze too much out of one thing, so you end up with less than ideal solution for everything.

> * Do you think the pros of having all of these things defined in one place outweigh the cons of whatever perceived complexity is added above?

I don't believe there's many pros of having these things defined in one place because the use cases are different and there are better options for both.

jdhoffa:
> I'm not saying it adds infrastructure, I'm saying it requires infrastructure... it requires a complex infrastructure that includes numerous files and systems that all have to work together in a perfect symphony just to put a list of dependencies in the README solely for the purpose of being "more transparent" for some hypothetical unexperienced R users.

Also, the users mentioned aren't hypothetical. One is here: https://github.com/2DegreesInvesting/PACTA_analysis/issues/466

> > * Do you think changing the status quo, by extracting the list of dependencies to one universal file (whatever file that may be), and having each of the different potential ways people use PACTA (users, docker, web tool) all referring to that file adds significant complexity to the already existing infrastructure?
> 
> Yes... if the goal is to make this "more transparent" for some hypothetical unexperienced R users, then I believe it is more complex to try to squeeze that feature out of a single file that is also meant to feed programmatic tools... it's trying to squeeze too much out of one thing, so you end up with less than ideal solution for everything.
> 
> > * Do you think the pros of having all of these things defined in one place outweigh the cons of whatever perceived complexity is added above?
> 
> I don't believe there's many pros of having these things defined in one place because the use cases are different and there are better options for both.

You keep saying "hypothetical user", as though we haven't already run into multiple users that can't easily run PACTA? Specifically where dependencies are concerned. The call we had with the documentarian is one example. https://github.com/2DegreesInvesting/PACTA_analysis/issues/466 this is one example, Walter (RMI) is one example (he had no idea how to get started), although admittedly that was more a holistic issue of no documentation.

jdhoffa:
> > * Do you think changing the status quo, by extracting the list of dependencies to one universal file (whatever file that may be), and having each of the different potential ways people use PACTA (users, docker, web tool) all referring to that file adds significant complexity to the already existing infrastructure?
> 
> Yes... if the goal is to make this "more transparent" for some hypothetical unexperienced R users, then I believe it is more complex to try to squeeze that feature out of a single file that is also meant to feed programmatic tools... it's trying to squeeze too much out of one thing, so you end up with less than ideal solution for everything.
> 
> > * Do you think the pros of having all of these things defined in one place outweigh the cons of whatever perceived complexity is added above?
> 
> I don't believe there's many pros of having these things defined in one place because the use cases are different and there are better options for both.

And to me, the use-case is clear. What dependencies are required to run the `web_tool_scripts`?

jdhoffa:
> > > * Do you think changing the status quo, by extracting the list of dependencies to one universal file (whatever file that may be), and having each of the different potential ways people use PACTA (users, docker, web tool) all referring to that file adds significant complexity to the already existing infrastructure?
> > 
> > 
> > Yes... if the goal is to make this "more transparent" for some hypothetical unexperienced R users, then I believe it is more complex to try to squeeze that feature out of a single file that is also meant to feed programmatic tools... it's trying to squeeze too much out of one thing, so you end up with less than ideal solution for everything.
> > > * Do you think the pros of having all of these things defined in one place outweigh the cons of whatever perceived complexity is added above?
> > 
> > 
> > I don't believe there's many pros of having these things defined in one place because the use cases are different and there are better options for both.
> 
> And to me, the use-case is clear. What dependencies are required to run the `web_tool_scripts`?

Regardless of if it's `docker`, or an R "user" or whoever, they will require the same dependencies if they want to run it in a consistent way.

jdhoffa:
anyway, this PR is approved. You can merge whenever.

cjyetman:
> You keep saying "hypothetical user", as though we haven't already run into multiple users that can't easily run PACTA? Specifically where dependencies are concerned. The call we had with the documentarian is one example. #466 this is one example, Walter (RMI) is one example (he had no idea how to get started), although admittedly that was more a holistic issue of no documentation.

1. these would be resolved by using `{env}`
2. now that [this has been resolved](https://github.com/2DegreesInvesting/create_interactive_report/issues/422), all the dependency issues we had with the documentarian don't exist
3. #466 is about users that are so inexperienced that they don't read errors and messages and follow the advice they give and/or have no idea how to interpret `there is no package called ‘devtools’`... this is the "hypothetical" user I'm talking about, and I call them "hypothetical" because I'm not sure it's reasonable to cater to them.... there are many other opportunities for a user like that to get confused and not be able to interpret the results etc.... installing packages is like the least of their problems
4. honestly, if a user can't install git, R, RStudio and a few packages on their own, this repo is never going to be for them... I don't mean to be mean about it, it's just there is realistically some minimal level of user knowledge/experience required to use this code.... if they don't have that, that's what we have the free, publicly available website for

jdhoffa:
ok

jdhoffa (approved):

PR #526, opened 2021-11-28
move web_tool_script CI to rocker/verse image

cjyetman:
depends on #528


cjyetman:
closing because I moved this effort directly into #528 since it's critical to have our CI testing working on the same or similar infrastructure as the stuff intended for deployment

PR #527, opened 2021-11-28
replace R CMD check CI with simple testthat

cjyetman:


cjyetman:
closing because I found a way to keep both the `devtools::check()` CI tests that run `R CMD check` and the tests that run `devtools::test()` or `testthat::test_local()` to test scripts in the repo's root directory without causing errors on the `R CMD check` test. https://github.com/2DegreesInvesting/PACTA_analysis/pull/514#issuecomment-987501082

PR #528, opened 2021-11-28
switch to r-base base image for transitionmonitor Docker

### LINKED ISSUE #454: transition to using a rocker image as the base
This has been discussed a few times elsewhere, but this is meant to converge the discussion into one place and maybe/hopefully get the ball rolling on this eventually.

Right now, we use [2dii/r-packages](https://hub.docker.com/repository/docker/2dii/r-packages) (which is maintained at [2DegreesInvesting/docker/r-packages](https://github.com/2DegreesInvesting/docker/tree/master/r-packages)) as the base image for the Docker image that is used for transtionmonitor.com. It may be advantageous to instead use a popular, well-maintained Docker image (such as [rocker/verse](https://hub.docker.com/r/rocker/verse)) that includes R and many of the necessary packages already as the base image to which we add just the most use case specific changes needed.

There are a few very specific changes made in 2dii/r-packages, and we will need to ascertain if they are necessary for the final images we make for the web server...
https://github.com/2DegreesInvesting/docker/blob/9efe76454b91a45f47e32da5d5d7d440d6a8a644/r-packages/Dockerfile#L50-L53

We will also have to consider the list of installed packages and make sure that we add any that are not already installed in rocker/verse...
https://github.com/2DegreesInvesting/docker/blob/9efe76454b91a45f47e32da5d5d7d440d6a8a644/r-packages/Dockerfile#L9-L49

I do not take this change lightly... obviously we will need to do this very cautiously, and with a lot of testing.

cjyetman:
closes #454 
depends on #536


jdhoffa:
What are the advantages of doing this? You mention in #454 that there may be many advantages, but was just wondering if you could elaborate. 

One potential con I see of "using a Docker image that already includes many useful packages" is that we might accidentally depend on a package that we don't explicitly keep track of. What I mean is, if we add a dependency on a package that is already included in `rockerverse`, but don't add it to `README.md` or anywhere else, we wouldn't necessarily notice because the Docker image wouldn't complain on any CI checks. 

Not sure how big of a deal this is, but just wanted to flag it.

cjyetman:
advantages:
1. We benefit from using an image that is maintained by an organization that has the resources to do robust setup and testing (as opposed to using our in-house r2dii/system and 2dii/r-packages images)
2. We benefit from using a base image that is regularly updated, so that e.g. it's trivial to move to a new/different version of R if we need/want
3. We benefit by allowing some other organization that has suitable resources to deal with the chaos of getting LaTeX setup properly

I'm not sure what to say about adding package dependencies. The base image that we currently use also has a number of packages pre-installed, so I suppose that's already a possibility. I don't think it should be acceptable to add a dependency without properly documenting it, but that's more of a developer behavior issue.

jdhoffa:
I don't think it should be acceptable either, but many of these "behaviour things" can be pretty easy to miss/ forget, and checking annoying details that people often forget is sorta the whole point of CI. 

Also, we're going to have a dependency on an image that we maintain at some point no matter what, it's just a matter of where we make the "cut". `r2dii/system` I don't think is super necessary for us to maintain, and happy to pull from a maintained image that already has that infrastructure. But assuming we ever plan on depending on a package that isn't in `rocker`, we'll always need some form of `r2dii/packages`. Either we maintain it separately (as you guys do now), or it's a line in the main `docker` file that installs all dependencies from scratch, but the content of that file will always be something we need to keep track of. 

Totally agree with the Latex point. 

(To be clear, I'm not against using `rocker` at all, just exploring all the details of the decision and the reasoning behind it)

cjyetman:
> it's a line in the main `docker` file that installs all dependencies from scratch

that is what this PR does: [Dockerfile](https://github.com/2DegreesInvesting/PACTA_analysis/blob/db679ca8a9c4edd1de3dd19cd54ccd1c41df799f/transitionmonitor_docker/Dockerfile#L19-L25)
I think it's preferable to have that right up front, here in this repo, rather than maintained in a separate repo.

but importantly, it installs packages using a date tagged RSPM repo so that we always get the same package versions
https://github.com/2DegreesInvesting/PACTA_analysis/blob/db679ca8a9c4edd1de3dd19cd54ccd1c41df799f/transitionmonitor_docker/Dockerfile#L8

cjyetman:
> Also, we're going to have a dependency on an image that we maintain at some point no matter what, it's just a matter of where we make the "cut"

I would prefer the "cut" to be a fully functioning OS with a fully functioning R installation and a fully functioning LaTeX system. Installing/managing dependent packages is easy to do, and probably better for us to do so that we control the versioning. But I would say it's not precisely correct that we will "have a dependency on an image that we maintain at some point no matter what", at least not in the way that I understood you meant it.... in this PR we rely on only the primary base image (rocker/verse) and the package management is done directly in the Dokcerfile.

cjyetman:
Yes... it's also already in the [DESCRIPTION](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/DESCRIPTION) file (which we could maybe do away with if we stopped forcing this repo to be an R package)

jdhoffa:
> > it's a line in the main `docker` file that installs all dependencies from scratch
> 
> that is what this PR does: [Dockerfile](https://github.com/2DegreesInvesting/PACTA_analysis/blob/db679ca8a9c4edd1de3dd19cd54ccd1c41df799f/transitionmonitor_docker/Dockerfile#L19-L25) I think it's preferable to have that right up front, here in this repo, rather than maintained in a separate repo.
> 
> but importantly, it installs packages using a date tagged RSPM repo so that we always get the same package versions
> 
> https://github.com/2DegreesInvesting/PACTA_analysis/blob/db679ca8a9c4edd1de3dd19cd54ccd1c41df799f/transitionmonitor_docker/Dockerfile#L8

It looks like we have lists of dependencies:
- here: https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/DESCRIPTION
- here: https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/README.md
- here: https://github.com/2DegreesInvesting/PACTA_analysis/blob/db679ca8a9c4edd1de3dd19cd54ccd1c41df799f/transitionmonitor_docker/Dockerfile#L8
- and here: https://github.com/2DegreesInvesting/docker/blob/master/r-packages/Dockerfile

do you know of other files that contain a list of dependencies?

jdhoffa:
> > Also, we're going to have a dependency on an image that we maintain at some point no matter what, it's just a matter of where we make the "cut"
> 
> I would prefer the "cut" to be a fully functioning OS with a fully functioning R installation and a fully functioning LaTeX system. Installing/managing dependent packages is easy to do, and probably better for us to do so that we control the versioning. But I would say it's not precisely correct that we will "have a dependency on an image that we maintain at some point no matter what", at least not in the way that I understood you meant it.... in this PR we rely on only the primary base image (rocker/verse) and the package management is done directly in the Dokcerfile.

Yup that makes sense. Then I tend to agree, use rocker as base, manage dependencies explicitly directly in the repo.

jdhoffa:
All I mean "we will have to maintain an image" is like... `r2dii/packages` is just a Dockerfile. We can add it to an existing Dockerfile, and then two files become one file, but we are still maintaining the same "code". Not sure if I'm explaining myself well.

jdhoffa:
> > > Also, we're going to have a dependency on an image that we maintain at some point no matter what, it's just a matter of where we make the "cut"
> > 
> > 
> > I would prefer the "cut" to be a fully functioning OS with a fully functioning R installation and a fully functioning LaTeX system. Installing/managing dependent packages is easy to do, and probably better for us to do so that we control the versioning. But I would say it's not precisely correct that we will "have a dependency on an image that we maintain at some point no matter what", at least not in the way that I understood you meant it.... in this PR we rely on only the primary base image (rocker/verse) and the package management is done directly in the Dockerfile.
> 
> Yup that makes sense. Then I tend to agree, use rocker as base, manage dependencies explicitly directly in the repo.

The bit about it coming with some R packages baked in makes me a bit nervous, and I think of as a down-side, but I can get over that.

cjyetman:
> All I mean "we will have to maintain an image" is like... `r2dii/packages` is just a Dockerfile. We can add it to an existing Dockerfile, and then two files become one file, but we are still maintaining the same "code". Not sure if I'm explaining myself well.

yeah, I get what you mean (I think?)... but compare this chain of Dockerfiles (two of which are outside of this repo), which is what we currently rely on...
https://github.com/2DegreesInvesting/docker/blob/master/system/Dockerfile
https://github.com/2DegreesInvesting/docker/blob/master/r-packages/Dockerfile
https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/transitionmonitor_docker/Dockerfile

to the one Dockerfile in this PR...
https://github.com/2DegreesInvesting/PACTA_analysis/blob/db679ca8a9c4edd1de3dd19cd54ccd1c41df799f/transitionmonitor_docker/Dockerfile

I think the latter is much cleaner and easier to maintain, plus the added advantages I stated above

cjyetman:
> do you know of other files that contain a list of dependencies?

no, I think that's it

- one of those would be eliminated by this PR
- one could be eliminated if we weren't forcing this to be an R package
- one is there because you wanted it to be more "transparent"

jdhoffa:
> > All I mean "we will have to maintain an image" is like... `r2dii/packages` is just a Dockerfile. We can add it to an existing Dockerfile, and then two files become one file, but we are still maintaining the same "code". Not sure if I'm explaining myself well.
> 
> yeah, I get what you mean (I think?)... but compare this chain of Dockerfiles (two of which are outside of this repo), which is what we currently rely on... https://github.com/2DegreesInvesting/docker/blob/master/system/Dockerfile https://github.com/2DegreesInvesting/docker/blob/master/r-packages/Dockerfile https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/transitionmonitor_docker/Dockerfile
> 
> to the one Dockerfile in this PR... https://github.com/2DegreesInvesting/PACTA_analysis/blob/db679ca8a9c4edd1de3dd19cd54ccd1c41df799f/transitionmonitor_docker/Dockerfile
> 
> I think the latter is much cleaner and easier to maintain, plus the added advantages I stated above

Yup, 100% agree with this!

cjyetman:
> * the quotations around that word make it seem you don't think that's important? If we expect people to "install" this repo, do you not think it's important to clarify to them how to install the dependencies? (and by clarify I mean more clear than pointing them to a few lines baked into a Dockerfile, or asking them to run a series of commands to programmatically do it)

Just because that was your word, not mine. Personally I felt like the DESCRIPTION file was transparent enough.

cjyetman:
> do you not think it's important to clarify to them how to install the dependencies?

No, I think there's some reasonable, minimum bar for knowledge and experience using R that can be expected, and explaining to people how to install packages is below that bar. Realistically, this repo is not of any use to someone that does not already know how to do that.

jdhoffa:
Ahahaha oh man, then this was a miscommunication. My preference was also to only specify all dependencies in `DESCRIPTION`

cjyetman:
> * Any docker image that runs this repo, installs each of the dependencies DIRECTLY from that file (ie. doesn't have them written out like in `r2dii/packages`)

Can you write a Dockerfile that properly parses a R package DESCRIPTION file? I'm a bit daunted by that task to be honest. Unless you mean add a line like 
```Dockerfile
RUN Rscript -e 'devtools::install_deps(".")'
```
I'm a bit more comfortable with 
```Dockerfile
RUN install2.r --skipinstalled --error [list_of_packages]
```
because I know it skips packages that were already installed, and it doesn't, yet again, force us to conform to some particular R package structure

jdhoffa:
> > * Any docker image that runs this repo, installs each of the dependencies DIRECTLY from that file (ie. doesn't have them written out like in `r2dii/packages`)
> 
> Can you write a Dockerfile that properly parses a R package DESCRIPTION file? I'm a bit daunted by that task to be honest. Unless you mean add a line like
> 
> ```dockerfile
> RUN Rscript -e 'devtools::install_deps(".")'
> ```
> 
> I'm a bit more comfortable with
> 
> ```dockerfile
> RUN install2.r --skipinstalled --error [list_of_packages]
> ```
> 
> because I know it skips packages that were already installed, and it doesn't, yet again, force us to conform to some particular R package structure

Ya, in the past I have just done this: 
```dockerfile
RUN Rscript -e 'install.packages("remotes")'
COPY DESCRIPTION 
RUN Rscript -e 'remotes::install_deps("/bound", dependencies = TRUE)'
```

The only trouble for "double installing" packages, I guess would be installing those that `rockerverse` already has? But in any case, the `DESCRIPTION` file shouldn't change often, so Docker's caching system should mean that the installation process doesn't get re-run super frequently.

cjyetman:
that's pretty meta... creating a new package dependency to install the package dependencies 🤔

jdhoffa:
> that's pretty meta... creating a new package dependency to install the package dependencies 🤔

Haha I mean `devtools::install_deps` is just `remotes::install_deps`. I just usually do `remotes` in my other repos to not require the added `devtools` dependency unless it's necessary.

cjyetman:
Based on the [rocker/verse Dockerfile](https://hub.docker.com/r/rocker/verse/dockerfile), it also appear to be possible to lock-in specific versions of the TeX packages based on a fixed date with something like (though I'm not sure how to determine the appropriate date for our purposes)...
```Dockerfile
# Version-stable CTAN repo from the tlnet archive at texlive.info
ARG CTAN_REPO=${CTAN_REPO:-https://www.texlive.info/tlnet-archive/2021/11/18/tlnet}
ENV CTAN_REPO=${CTAN_REPO}
```

cjyetman:
@jdhoffa one critical piece of information I forgot in this discussion of managing (and documenting) dependencies...

The Docker image that this PR is meant to build is for the TM website, and therefore, it requires all dependencies  of the entire "family of PACTA repos", which will probably always be larger than the list of dependencies needed for just PACTA_analyis. We could inflate the list of this repo's dependencies just so that it covers all the other repos as well, but this already breaks your desire for a single point of truth about the required dependencies (the other repos might change their dependencies).

Given the intent for this Dockerfile, I think it makes sense for it to maintain its own set of dependencies, distinct and separate from that of this repo.

jdhoffa:
> @jdhoffa one critical piece of information I forgot in this discussion of managing (and documenting) dependencies...
> 
> The Docker image that this PR is meant to build is for the TM website, and therefore, it requires all dependencies of the entire "family of PACTA repos", which will probably always be larger than the list of dependencies needed for just PACTA_analyis. We could inflate the list of this repo's dependencies just so that it covers all the other repos as well, but this already breaks your desire for a single point of truth about the required dependencies (the other repos might change their dependencies).
> 
> Given the intent for this Dockerfile, I think it makes sense for it to maintain its own set of dependencies, distinct and separate from that of this repo.

I mean that's exactly why I wouldn't keep track of the dependencies in the Dockerfile haha. If the stress testing tool changes it's dependencies and we don't notice and update the Dockerfile, it's gonna break. If each tool manages it's dependencies in the `DESCRIPTION` file, then the Dockerfile can just install directly from that and always be up to date. 

(ie. if we want to use the tools `PACTA_analysis` and `create_interactive_report`, and they conveniently manage their dependencies in the `DESCRIPTION` file, then all we need to do in the Dockerfile is install from each of those `DESCRIPTION` files, and we don't need to keep track of the dependencies of either of those repos in the `Dockerfile` at all. It's just a step to better modularize each repo)

cjyetman:
`create_interactive_report` is not a package and does not have a `DESCRIPTION` file

jdhoffa:
Yes I recognize that hahaha, it doesn't need to be specifically a `DESCRIPTION` file, just a self-contained way for a tool to manage it's own dependencies (in python `requirements.txt` is frequently used, etc.), really doesn't matter what file is actually used. 

It just doesn't make sense to have a completely separate Dockerfile living somewhere, that needs to be updated every time the dependencies of a downstream repo change. 

It makes more sense for each repo to manage it's own dependencies in a single file, and then if a Dockerfile wants to use that tool, it just has to install the up-to-date list of dependencies for each repo used. Does that make sense? The "R package" stuff doesn't matter at all, it's just a matter of better modularizing these tools.

jdhoffa:
Whenever I say `DESCRIPTION` file, feel free to replace it conceptually with `imports.R` or `requirements.txt` or any other dependency management file that fits most given the specific context

AlexAxthelm:
@cjyetman this is going to take a while to review, but I'll get to it soon. Anything in particular I should be testing?

cjyetman:
- The Dockerfile should probably be reviewed conceptually. It's precise, well-formatted, and minimized, so you should be able to read through it and determine if it makes sense.
- Each of the 3 shell scripts I made substantial improvements and adjustments to. Maybe you can read through them and make sense of them. For me, reading shell scripts is very foreign so... I can only really test them and see if they work. They all do at least have a usage function and description of the options, which I copied from you, so that should at least help in understanding the intended functionality.
- I made substantial modifications to the README. I definitely recommend reading that thoroughly. It will also help with understanding some of the new intended features.
- Maybe the most important thing for you to test is that this new Dockerfile will build and work properly in the Azure pipelines you've been working on.

The big changes to the Dockerfile are...
- a completely new, completely self-contained Dockerfile
- the base is r-base:4.1.2 which is [Docker Hub's "official" r-base image](https://hub.docker.com/_/r-base) (though as I understand, it is maintained by Rocker)
  - upgrades us to R v4.1.2
  - is multi-arch, and therefore allows for building our Docker image targeting different platforms, e.g. an M1 Mac
  - because it's maintained by Rocker and R version tagged, we will have access to frequently built base images for specific versions of R, making it trivial for use to switch to a different version of R
  - this means we no longer need to maintain, use, or rely on our self-made Docker images in the [2DegreesInvesting/docker](https://github.com/2DegreesInvesting/docker) repo
- it installs xelatex and a minimal set of dependencies (fonts)
  - our current version installs LuaTeX and texlive, and then relies on texlive R functions to do the install and setup tlmgr, which is problematic because that's intended to do user level installs
  - xelatex is the preferred LaTeX engine accdording to our [executive summary Rmd files](https://github.com/2DegreesInvesting/create_interactive_report/blob/fd7ada58b5cfaf8dcc0f2fd4a49026f9b8bcbd9a/swiss_en_exec_summary/template.Rmd#L5)
- it installs only the necessary TeX packages
  - it uses tlmgr directly to do this directly
  - it uses a CTAN repo that is fixed to a certain date so that we have stable versions of the TeX packages
- it installs the R packages that are needed, and the system dependencies of those R packages
  - I've gone to great lengths to minimize our dependencies and only install those that are needed, which makes the Docker build process much more efficient
  - it uses a CRAN repo that is fixed to a certain date so that we have stable versions of the R packages
  - it uses the `{remotes}` package to install the R packages, which is much more efficient than base R's `install.packages()`
- the order of the Docker layers have been optimized and minimized to make Docker's caching system work as well as possible, meaning that build times can be very quick if you have already built previous version on your machine
- I realize that we maybe/hopefully are moving towards using an Azure pipeline for building and testing the Docker images, however, my goal was to have a single Dockerfile that can be used in an Azure pipeline or locally on any of our machines to facilitate building, testing, and experimenting it multiple contexts. I believe this will be advantageous for us in the long run.

The big changes with the shell scripts are...
- The build script can now target different platforms, e.g. I can now build a Docker image for "linux/arm64" (M1 Mac) and run the test much more efficiently on my Mac than running a "linux/x86_64" version in emulation.
- The build script gained typical shell command options modeled after the ones you made for the test scrips, including a usage function to describe the usage of the options.
- The build script no longer requires you to delete all previous version of the 2dii_pacta Docker image before running. You can build incremental versions and leave them loaded in your local Docker system (as long as the tags do not clash).
- The build script no longer exports/saves the Docker image as a .tar.gz file by default, unless you explicitly set the `-s` option. The export process is often very long, sometimes even longer than the build process, so this significantly decreases the amount of time needed to build a new image and test it.
- The build script now upon successful completion lists a number of useful commands for possible tasks that a user might want to do.
- I included the run-all-tests.sh script that we have had in our test zipfle, but had not previously been in this repo. I have improved it a bit and aligned it with the options of the other scripts, e.g. adding the `-x` option to target a docker image for a specific platform ("linux/x86_64" is always the default, because that is mandatory for the TM website).
- I also added some options and made some improvements to the run-like-constructiva-flags.sh script

Outside of this PR, I've managed to squelch almost all of the common warnings, message, errors, that clutter the console output from running the Docker image and instead send those message to the log file where they belong. As it is, I can run all of the tests in pacta_web_template_v0.0.14.zip in less than 30 minutes (assuming I build a "linux/arm64" image), and the output has only 1 message across all test portfolios. I believe this will substantially help in finding problems and fixing problems in the future, as we will not be utterly exhausted with the output that we eventually start skimming through because we're used to seeing so much of it.

transitionmonitor_docker/Dockerfile
```diff
@@ -1,23 +1,133 @@
...
+# https://hub.docker.com/_/r-base
+# https://www.texlive.info/tlnet-archive
+# https://packagemanager.rstudio.com/client/#/repos/1/overview
 
-RUN echo "options(tinytex.tlmgr.path = '/root/.TinyTeX/bin/x86_64-linux/tlmgr')" > .Rprofile \
+FROM r-base:4.1.2
+
+# install system dependencies
+ARG SYS_DEPS="\
+    git \
```
AlexAxthelm, line 10: does `git` really need to be installed inside the container?

cjyetman: ideally, yes...

1. in our/my current process, we tag the repos from within the docker container (which doesn't necessarily have to be the case, but we need a robust alternative)
2. the environment_info.json export includes the local git tag and git hash (this is for a backup/comparison to the environment variable `build_version`, so it's not totally necessary, but I thought it was a good backup)

PR #530, opened 2021-11-29
rename `portfolio_name_in` to `portfolio_name` everywhere

### LINKED ISSUE #485: change `portfolio_name_in` to `portfolio_name` in portfolio parameter files
In the existing example portfolio parameter files, `portfolio_name_in` and `investor_name_in` keys are used. However, the code that reads in the portfolio parameter file looks for keys `portfolio_name` and `investor_name`, which works anyway by accident, as far as I can tell, because Extract (`$`) does partial matching.

https://github.com/2DegreesInvesting/PACTA_analysis/blob/897433f71c18edb6ba2cf6798f4c961158f27b0d/0_web_functions.R#L94-L95

I propose to universally change the keys `portfolio_name_in` and `investor_name_in` to `portfolio_name` and `investor_name` everywhere in this repo, which at the moment only appears in the existing example portfolio parameter files, and here...
https://github.com/2DegreesInvesting/PACTA_analysis/blob/897433f71c18edb6ba2cf6798f4c961158f27b0d/workflow-pacta-cop.Rmd#L490-L491

@AlexAxthelm do you agree?

cjyetman:
closes #485

### DIFF
--- a/indices_benchmarks/combine_indices.R
+++ b/indices_benchmarks/combine_indices.R
@@ -34,8 +34,8 @@ for (portfolio in portfolios) {
     list(
       default = list(
         parameters = list(
-          investor_name_in = investor_name,
-          portfolio_name_in = portfolio_name,
+          investor_name = investor_name,
+          portfolio_name = portfolio_name,
           language = "EN",
           project_code = investor_name
         )
--- a/workflow-pacta-cop.Rmd
+++ b/workflow-pacta-cop.Rmd
@@ -487,8 +487,8 @@ yaml_data <-
   list(default = list(
     parameters =
       list(
-        portfolio_name_in = portfolio_name,
-        investor_name_in = investor_name,
+        portfolio_name = portfolio_name,
+        investor_name = investor_name,
         peer_group = peer_group_type,
         language = language_to_print,
         project_code = project_code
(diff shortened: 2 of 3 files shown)


AlexAxthelm (approved):
I get _why_ this works (partial name matching, because we use `$` rather than `[["` as  the accessor everywhere, but I still feel uneasy that the checks are all passing on the first time around. Do you think it's worth some further testing? :lolsob:

cjyetman:
The intent here is to eliminate exactly this reliance on partial matching. We currently have a mix of `portfolio_name_in` and `portfolio_name` in our config files, and in our R scripts we have only `portfolio_name` (to be precise, the R script that is changed here is programmatically creating a config file with the wrong key name `portfolio_name_in`, so I count that with the bogus config files). The only reason those existing mis-matches work right now is because partial matching accidentally saves us.

So, if anything, I would expect this to make things more robust rather than less, but really it should make no difference... at least until someone does write code like `cfg$parameters["portfolio_name"]`, in which case our existing code will break. The only reason I want to get this done now is because our single example config file uses the wrong name, so that will easily propagate to more stuff eventually, and then it will be an even bigger nightmare to fix.

PR #535, opened 2021-11-30
quit wiith 0 status code, rather than error

AlexAxthelm:
the Constructiva server fails quietly (infinite spinner) if the docker
container errors, but it will show the index.html if it returns status
code `0`. So for user-correctable errors, we'll do that, and give them
an informative message.

### DIFF
--- a/R/user_error_checks.R
+++ b/R/user_error_checks.R
@@ -50,7 +50,11 @@ check_grouped_portfolio_years <- function(
       description = desc,
       immediate = TRUE
     )
-    stop("port_holdings_date contains multiple distinct values", call. = FALSE)
+    # using message and quit, rather than stop(), because server silently
+    # fails if docker container exits with anything other than a 0 status
+    # code
+    message("port_holdings_date contains multiple distinct values")
+    quit(status = 0L)
   }
   return(invisible(port_holdings_date))
 }
--- a/R/user_error_checks.R
+++ b/R/user_error_checks.R
@@ -54,7 +54,7 @@ check_grouped_portfolio_years <- function(
     # fails if docker container exits with anything other than a 0 status
     # code
     message("port_holdings_date contains multiple distinct values")
-    quit(status = 0L)
+    quit(save = "no", status = 0L, runLast = FALSE)
   }
   return(invisible(port_holdings_date))
 }


R/user_error_checks.R
```diff
@@ -50,7 +50,11 @@ check_grouped_portfolio_years <- function(
       description = desc,
       immediate = TRUE
     )
-    stop("port_holdings_date contains multiple distinct values", call. = FALSE)
+    # using message and quit, rather than stop(), because server silently
+    # fails if docker container exits with anything other than a 0 status
+    # code
+    message("port_holdings_date contains multiple distinct values")
+    quit(status = 0L)
```
cjyetman, line 57: might be safer to use `save = "no"` here? I think the default is probably "yes" or "ask"?

actually (emphasis mine)...

> The default is to ask in interactive use but may be overridden by command-line arguments **(which must be supplied in non-interactive use)**.

cjyetman (approved):
can we also use [`write_log()`](https://github.com/2DegreesInvesting/PACTA_analysis/blob/21187adf25697c10caaf3b9a517f2607dcb7594a/0_global_functions.R#L393-L404) here?

otherwise.... good riddance problem

R/user_error_checks.R
cjyetman: > Some error status values are used by R itself. The default error handler for non-interactive use effectively calls q("no", 1, FALSE) and returns error status 1. Error status 2 is used for R ‘suicide’, that is a catastrophic failure, and other small numbers are used by specific ports for initialization failures. It is recommended that users choose statuses of 10 or more.

so, maybe `quit(save = "no", status = 0, runLast = FALSE)`? your call

I'm not sure why it returns `0` instead of `0L` by default 🤷🏻

AlexAxthelm: Good points @cjyetman . I use `0L`, instead of `0`, because I want to make explicit that it's an integer return code.

AlexAxthelm:
> can we also use [`write_log()`]

That gets called as part of `log_user_errors()` on line 47, unless you want to also add a message along the lines of "ending analysis early"

cjyetman:
> That gets called as part of `log_user_errors()` on line 47, unless you want to also add a message along the lines of "ending analysis early"

nope, once is enough ;-)

PR #536, opened 2021-12-01
add build version as ENV var inside image and display it when running

cjyetman:
prototype for burning the Docker build version tag into the image as an environment variable when it is built, and then printing that, if it's available, along with the message that shows what script is running.

@AlexAxthelm you know what this is for 😬

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -157,7 +157,7 @@ expect_no_message <- function(object, regexp = NA, ...) {
   testthat::expect_message(object = object, regexp = regexp, ...)
 }
 
-get_build_version <- function(env_var = "$build_version") {
-  build_version <- system2("echo", args = env_var, stdout = TRUE)
+get_build_version <- function(env_var = "build_version") {
+  build_version <- Sys.getenv(env_var)
   ifelse(nzchar(build_version), paste0("v", build_version), NA_character_)
 }
--- a/R/utils.R
+++ b/R/utils.R
@@ -156,3 +156,8 @@ mark_end <- function() {
 expect_no_message <- function(object, regexp = NA, ...) {
   testthat::expect_message(object = object, regexp = regexp, ...)
 }
+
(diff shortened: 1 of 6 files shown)


cjyetman:
would love to get this in the next build @AlexAxthelm

R/utils.R
```diff
@@ -156,3 +156,8 @@ mark_end <- function() {
 expect_no_message <- function(object, regexp = NA, ...) {
   testthat::expect_message(object = object, regexp = regexp, ...)
 }
+
+get_build_version <- function(env_var = "$build_version") {
+  build_version <- system2("echo", args = env_var, stdout = TRUE)
```
AlexAxthelm, line 161: is there a reason to not use `Sys.getenv()` here?

AlexAxthelm (approved):

R/utils.R
cjyetman: because as I understood, `Sys.getenv()` returns R environment variables, but not system variable, right?
```r
Sys.getenv("$HOME")
#> [1] ""
system2("echo", args = "$HOME", stdout = TRUE)
#> [1] "/Users/cj2dii"
```

to be honest, that's always been a bit confusing to me, so I could be not getting it right

AlexAxthelm: Don't include the `$`.
```r
R> Sys.getenv("HOME")
[1] "/Users/alexaxthelm"
R> Sys.getenv("$HOME")
[1] ""
R> system2("echo", args = "$HOME", stdout = TRUE)
[1] "/Users/alexaxthelm"
```

cjyetman: changed to use `Sys.getenv()`... if that works for you, let's merge

cjyetman: I restarted tests.... first round failed, but it was on the checkout of pacta-data repo, so should be not related to this change

cjyetman: 2nd round of tests passed

PR #544, opened 2021-12-08
test `set_portfolio_parameters()`

cjyetman:
test `set_portfolio_parameters()`  (requires `{yaml}`, so I added it to `Suggests`)
This does not run during `R CMD check` because `set_portfolio_parameters()` is defined in [0_web_functions.R](https://github.com/2DegreesInvesting/PACTA_analysis/blob/6efa32bf616d6fa747b7757dd87b9854b504cf56/0_web_functions.R#L91-L101) which can not be found during `R CMD check`.

This was prompted by a regression caused by #540 which had to be reverted by #543. That unset keys in the parameter file end up as objects in the global environment with  a value of `NULL` is now [tested here](https://github.com/2DegreesInvesting/PACTA_analysis/blob/3d9006d0b5dd44689dc4cf67077adc4e58150471/tests/testthat/test-0_web_functions.R#L52-L71), however, the desired effect of #540 was that any key that is set should always end up as an object in the global environment with  a **character** value. Once that issue (#545) is properly addressed, a test [should be added here](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/tests/testthat/test-0_web_functions.R) to ensure that if a key is set to a numeric value, that the matching global environment object's value is coerced to be a character value. e.g.
```yaml
investor_name: 2534
```
should create `investor_name == "2534"`


AlexAxthelm (approved):

maurolepore (approved):
LGreatTM!

It looks like we may benefit from a [local helper](https://testthat.r-lib.org/articles/test-fixtures.html#local-helpers) `local_yaml()`.

cjyetman:
> It looks like we may benefit from a [local helper](https://testthat.r-lib.org/articles/test-fixtures.html#local-helpers) `local_yaml()`.

I didn't know about that one.... or do you mean a custom one?

cjyetman:
@maurolepore I was also wondering if I made the right choice with `withr::with_environment(emptyenv(), { XXX})`. If I understand correctly, if I used `withr::local_environment()` then the temporary/local environment would not clear until all the tests in ```test_that("`set_portfolio_parameters()` works as expected", { ...``` had finished, which would be bad because I need the temp/test environment to be cleared between each test... does that make sense. Otherwise, I could have split each test into their own `test_that()` function, but I think you lean now towards having all expectations for a given function in one `test_that()` function, correct?

🤦🏻 testing functions that set values in global is also a mess

cjyetman:
does this look right?
```r
local_yaml <- function(parameters,
                       filename = "portfolio_config.yml",
                       env = parent.frame()) {
    yaml::write_yaml(list(default = list(parameters = parameters)), filename)
    withr::defer(unlink(filename, recursive = TRUE), envir = env)
}
```

maurolepore:
@cjyetman, 

As you now suspect, yes, I mean to write a custom `local_yaml()`. That would DRY this PR a bit and reduce the pain you and I note in testing functions that rely on global state.

Your implementation looks good. The most tricky thing I think you already got right: To use `env = parent.frame()` in the signature and pass it to `withr::defer()` or a friend. Among the friends I see `withr::local_file()` which compared to `withr::defer()` seems like a better fit for this case. I'm not sure it'll work might be worth trying. I note the argument that takes the environment is not `envir` but `.local_envir` so you would need to adapt your implementation accordingly.

Here is my suggestion. Note:

- `filename` moved to position 1 so you can inspect the result with `readLines(local_yaml())` or similar.
- `parameters` expects a list. In your implementation I'm not sure how you would pass an arbitrary number of unpredictable `key = value` pairs. But if yours works then go ahead.

``` r
local_yaml <- function(filename, parameters, env = parent.frame()) {
  withr::local_file(.file = filename, .local_envir = env)
  yaml::write_yaml(list(default = parameters), filename)
  
  invisible(filename)
}


path <- "my.yml"
params <- list(key1 = "value1", key2 = "value2")

local({
  local_yaml(path, params)
  file.exists(path)
})
#> [1] TRUE

file.exists(path)
#> [1] FALSE
```

<sup>Created on 2021-12-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

maurolepore:
RE the scope of the state-modification:

If you use `withr::with_*(something, {...})` friends the scope is limited to the contends of `{...}`. 

If you use `withr::local_*()` friends or your custom version then the scope is that of the environment that calls it -- you can see this in my previous comment, where I used `local({...})`, but I could have also used `f <- function() {...}` or `thestthat::test_that("title", {...})`.

Answering your questions more directly, the local yaml file will be available to all expectations within `{...}` in `test_that("title", {...})`. You may need to organize your tests and expectations accordingly. And while it's a pain to write tests that rely in global state, the new `local_yaml()` would reduce the friction and you'd be motivated to add fewer expectations per test -- if that's best.

cjyetman:
> * `filename` moved to position 1 so you can inspect the result with `readLines(local_yaml())` or similar.

you mean `return` the filename, right? because this would do the same thing...
```r
local_yaml <- function(parameters, filename, env = parent.frame()) {
  withr::local_file(.file = filename, .local_envir = env)
  yaml::write_yaml(list(default = parameters), filename)
  
  invisible(filename)
}

path <- "my.yml"
params <- list(key1 = "value1", key2 = "value2")

local({
  print(readLines(local_yaml(params, path)))
  file.exists(path)
})
#> [1] "default:"       "  key1: value1" "  key2: value2"
#> [1] TRUE


file.exists(path)
#> [1] FALSE
```

cjyetman:
ok... this is what I came up with. Ironically, I feel like the tests are more complicated than the thing they're testing, which makes me feel like I need some tests for the tests. 🤣 

I don't want to always set the filename, so I leave that as the 2nd argument with a default value. Returning the filename invisibly allows me to capture whatever the filename was set to, default or otherwise, so that I can use it downstream. I modeled the argument names after those in `withr::local_file()` for consistency.

So yeah, my understanding was correct, I need to use `withr::with_environment()` here because e.g. if the first test sets `portfolio_name == "TestPortfolio"` in the local environment of the `test_that()` function block, then further tests (particularly the one that expects `portfolio_name == NULL`) are liable to not work as expected if I don't make sure the local environment of the `test_that()` function block is cleared and `exists("portfolio_name") == FALSE`, so it's just easier/safer to run each test in a child `withr::with_environment()` environment and not worry about it.

The `parameters`, now `.params`, object needs to be a named list (it could have nested named lists). And that list needs to be an element named "parameters" of a list, which is itself an element named "default" of a list. 🤯 Weird, but 100% critical... that's what `yaml::write_yaml()` and `yaml::read_yaml()` expect.

```r
local_yaml <- function(.params,
                         .file = "portfolio_config.yml",
                         .local_envir = parent.frame()) {
    withr::local_file(.file = .file, .local_envir = .local_envir)
    yaml::write_yaml(list(default = list(parameters = .params)), .file)
    invisible(.file)
  }

  # test that standard parameter names are read in to
  # the proper object names in the global environment
  withr::with_environment(
    emptyenv(),
    {
      parameters <-
        list(
          portfolio_name = "TestPortfolio",
          investor_name = "TestInvestor",
          peer_group = "test_group",
          language = "DE",
          user_id = "2345",
          project_code = "GENERAL",
          holdings_date = "2020Q4"
        )
      filename <- local_yaml(parameters)

      set_portfolio_parameters(filename)
      expect_equal(portfolio_name, "TestPortfolio")
      expect_equal(investor_name, "TestInvestor")
      expect_equal(peer_group, "test_group")
      expect_equal(language_select, "DE")
      expect_equal(user_id, "2345")
      expect_equal(project_code, "GENERAL")
      expect_equal(port_holdings_date, "2020Q4")
    }
  )
```

cjyetman:
ack! no, that does not work the way I expected it to...
``` r
library(testthat)

test_that("`set_portfolio_parameters()` works as expected", {
  withr::with_environment(emptyenv(), { portfolio_name <- "something" })
  print(exists("portfolio_name"))
  
  withr::with_environment(emptyenv(), { rm(portfolio_name) })
  print(exists("portfolio_name"))
})
#> [1] TRUE
#> [1] FALSE
#> ── Skip (???): `set_portfolio_parameters()` works as expected ──────────────────
#> Reason: empty test
```

cjyetman:
putting this on pause, because I don't understand how this works...
```r
test <- function() {
  withr::with_environment(new.env(), { portfolio_name <- "something" })
  portfolio_name
}
test()
#> [1] "something"
```
and if an object created inside `withr::with_environment(new.env(), {`...here...` })` is still around after that, then these tests are not testing what they're meant to test.

cjyetman:
ok, I completely rewrote these tests here [tests/testthat/test-0_web_functions_v2.R](https://github.com/2DegreesInvesting/PACTA_analysis/blob/e538c656edcc136ec3540518b48e775f207c8ffb/tests/testthat/test-0_web_functions_v2.R)

now they actually do what I wanted to save

maurolepore:
Fair enough. This idea of test fixtures is weird enough and the `withr::with_*()` family is quite puzzling to me too.

While we're here I share some thoughts:

1. Why do you use `emptyenv()`? I think by definition you can't assign in it.

``` r
try(assign("x", 1, envir = emptyenv()))
#> Error in assign("x", 1, envir = emptyenv()) : 
#>   cannot assign values in the empty environment
```

The `with_*()` family seems to NOT work as I expect but the `local_*()` family does:

``` r
x <- 1
e <- new.env()
local({
  withr::local_environment(e)
  x <- 9
  x
})
#> [1] 9

x
#> [1] 1
```

However it may not work in this case because the `<<-` in `set_portolio_paremeters()` would break the seal of the local environment, all the way up to the global environment.

``` r
e <- new.env()
x <- 1
local({
  withr::local_environment(e)
  x <<- 9
  x
})
#> [1] 9

x
#> [1] 9
```

Not sure if this helps but `callr` can help you run some code in a separate process and thus a separate global environment -- which you could then return as a list and test if the elements of that list meet your expectations.

``` r
# Maybe callr can help you get a separate global environment?
# You can return it as a list and query it in your tests
out <- callr::r(function() {
  source("0_web_functions.R")
  file_path <- "working_dir/10_Parameter_File/TestPortfolio_Input_PortfolioParameters.yml"
  set_portfolio_parameters(file_path)
  
  as.list.environment(.GlobalEnv)
})

testthat::expect_equal(out$portfolio_name, "TestPortfolio_Input")
testthat::expect_equal(out$investor_name, "Test")
```

cjyetman:
> 1. Why do you use `emptyenv()`? I think by definition you can't assign in it.

That was not a good idea. I realized that and changed to `new.env()` in the [later example](https://github.com/2DegreesInvesting/PACTA_analysis/pull/544#issuecomment-989033797).

The problem with my understanding of the intent of `withr::local_environment()` and `withr::with_environment()` was this (besides that they are basically undocumented and I was trying to infer their purpose based on analogous inference from the names and descriptions of the similar `withr::local_*()` and `withr::with_*()` functions)...

As you've said, the function I'm trying to test uses `<<-` 😬. I was expecting the `withr::*_environment()` pair of functions to use a specified environment as the pseudo-global environment of the code you applied it to, so I was expecting that if I create a new, empty environment and specified that as the desired environment then all the paired code would function as if that environment was the global environment, and after it as finished, `{withr}` would remove that environment for me and I would have avoided messing with anything in global.

After reading the [source for `withr::with_environment()`](https://github.com/r-lib/withr/blob/7725d4a36e3b6fe28b4001fdd0c224b876a67602/R/namespace.R#L72-L79), and in particular the default `pos = 2L` argument, it became clear to me what the intent is... it's intended to **simulate** having a global environment with desired objects in it, which is a totally understandable task, just far from what I was trying to get out of it.

So that's why my new code has a bunch of explicit `env = .GlobalEnv` type arguments. And yeah, once I realized `withr::with_environment()` is not what I wanted, I though about using `{callr}`, but I guess the sheer exhaustion of working through the misunderstanding of `{withr}` , I was loathe to give a go at another rather complicated package that I'm not an experienced user of.

cjyetman:
- reimplemented using `{call}` and now it's much cleaner
- moved to `test-set_portfolio_parameters.R` because there are too many functions in [0_web_functions.R](https://github.com/2DegreesInvesting/PACTA_analysis/blob/master/0_web_functions.R) to test in one script
- need to use [skip_if_R_CMD_check()](https://github.com/2DegreesInvesting/PACTA_analysis/blob/7da3180306251897646147644e188181148e11fc/R/skip_if_R_CMD_check.R#L1-L7) when #547 merges instead of the custom skip
- need to uncomment [the last test](https://github.com/2DegreesInvesting/PACTA_analysis/blob/1d38144277ecf5f1ded525041d3f67769cb068bc/tests/testthat/test-set_portfolio_parameters.R#L80-L95) once #545 has been fixed

PR #546, opened 2021-12-09
improve efficiency reducing and reordering layers

cjyetman:


AlexAxthelm:
I like these changes, mostly.

I've also been playing with the dockerfile on the `azure-pipelines` branch. 

basically, those changes boil down to:
we should add the repos in order of how big & hw often they change, so we can use caches better:
1. PACTA-data
2. stress test data
3. stress test
4. interactive report
5. PACTA_analysis

and the `chmod` can all be one command.

AlexAxthelm (approved):

PR #549, opened 2021-12-11
refactor add_fin_data()

cjyetman:
depends on #548 

`add_fin_data()` is one of the bootle neck functions that takes a lot of time, and it also chews up memory. It seems very easy to fix, but we have to strongly consider all the possibilities, and why someone did it the current way in the first place. Having tests will help, but I haven't thought of any great tests yet, so... let's see what other people think.


jdhoffa (approved):
lgtm

also... how would this function even work in the way it was written previously? wouldn't the `left_join` in line 540 add columns that don't exist in `portfolio_no_isin`? and wouldn't that mess up the `bind_rows`... odd.

jdhoffa:
Ah, nevermind, I see that `bind_rows` will just drop the columns that aren't shared by both data-frames... in which case, I am very very confused as to what this function is meant to do?

cjyetman:
Yeah, this is one of those things that I don't really get and was therefore afraid to guess what was intended so I just leave it that way, but I'm over it now. 😂

I think it just worked by accident... ["When row-binding, columns are matched by name, and any missing columns will be filled with NA."](https://dplyr.tidyverse.org/reference/bind.html)

PR #551, opened 2021-12-11
refactor clean_unmatched_holdings() from 0_portfolio_inoput_check_fun…

cjyetman:
depends on #550
refactor [clean_unmatched_holdings()](https://github.com/2DegreesInvesting/PACTA_analysis/blob/ad066dc1d78075779238b14dc44057f2f1e0792b/0_portfolio_input_check_functions.R#L1467-L1482) in [0_portfolio_input_check_functions.R](https://github.com/2DegreesInvesting/PACTA_analysis/blob/ad066dc1d78075779238b14dc44057f2f1e0792b/0_portfolio_input_check_functions.R)

This is another one that has been causing some pain. Looks really simple, but afraid to touch it without some tests like in #550 (though I think those should be expended)


jdhoffa (approved):
lgtm

PR #552, opened 2021-12-16
catch errors in stress test and log a message but continue

cjyetman:
Full stop errors in the stress testing leads to The Infinite Spinning Cog™ on the platform, which is not a good user experience, and is especially difficult to debug given the current constraints we have with access to logs and notifications of problems. It also seems to be one of the more common problems that we face.

This PR wraps the sourcing of the stress test code in a `tryCatch()` so that if/when an error code occurs, the report generation will continue and not cause a full stop error, which will lead to the user being able to view their report, though some of the stress testing charts/data may not be available.

Ideally, this PR should be paired with a PR in [create_interactive_report](https://github.com/2DegreesInvesting/create_interactive_report) that would notify the user if there is no data available to plot any of the stress testing charts with an informative message that would ideally prompt the user to let us know that something went wrong with their stress testing section. I imagine such a feature could work very similar to what was implemented in [this PR](https://github.com/2DegreesInvesting/create_interactive_report/pull/291). @MonikaFu

### DIFF
--- a/web_tool_script_3.R
+++ b/web_tool_script_3.R
@@ -45,12 +45,12 @@ options("ST_DATA_PATH" = stress_test_data_location)
 # run 2dii stress test
 tryCatch(
   source(file.path(stress_test_path, "web_tool_stress_test.R")),
-  error = function(e) simpleError("an error in web_tool_stress_test.R occurred")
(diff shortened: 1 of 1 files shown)


cjyetman:
FYI.... we may need to reach out to the PACTA team in general to determine if this is a suitable/acceptable solution.

AlexAxthelm (approved):

PR #557, opened 2021-12-17
Remove `{assertthat}` dependency

cjyetman:


jdhoffa (approved):
lgtm

maurolepore (approved):
LGTM

Recently I discovered `vctrs::vec_assert()` which may help do the same. Because vctrs is a dependency of packages we already use, it wouldn't add any new dependency overall.

``` r
vctrs::vec_assert(1L, ptype = double())
#> Error: `1L` must be a vector with type <double>.
#> Instead, it has type <integer>.
vctrs::vec_assert(1L, ptype = integer(), size = 2L)
#> Error: `1L` must have size 2, not size 1.
```

<sup>Created on 2021-12-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

PR #559, opened 2021-12-17
Remove `{rstudioapi}` dependency

cjyetman:
`{rstudioapi}` is only used in a few locations, and is only used if `rstudioapi::isAvailable()` is `TRUE`, so it is not a hard dependency and therefore should not be forced to be included. Starting these tests with `require(rstudioapi) && ` should prevent them from running unless `{rstudioapi}` is available.

https://github.com/2DegreesInvesting/PACTA_analysis/blob/11ea54588f9b618c6f1b09c5f712146f5f0628f3/1_portfolio_check_initialisation.R#L12-L16

https://github.com/2DegreesInvesting/PACTA_analysis/blob/21187adf25697c10caaf3b9a517f2607dcb7594a/0_global_functions.R#L34-L39

https://github.com/2DegreesInvesting/PACTA_analysis/blob/21187adf25697c10caaf3b9a517f2607dcb7594a/0_global_functions.R#L264-L269

related to #137

### DIFF
--- a/0_global_functions.R
+++ b/0_global_functions.R
@@ -32,11 +32,7 @@ is_blank_na <- function(x) {
 }
 
 set_location <- function() {
-  if (require(rstudioapi) && rstudioapi::isAvailable()) {
-    working_location <- dirname(rstudioapi::getActiveDocumentContext()$path)
-  } else {
-    working_location <- getwd()
-  }
+  working_location <- getwd()
 
   working_location <- paste0(working_location, "/")
 
@@ -262,11 +258,7 @@ set_project_paths <- function(project_name, twodii_internal, project_location_ex
 }
 
 set_git_path <- function() {
-  if (require(rstudioapi) && rstudioapi::isAvailable()) {
-    git_path <- dirname(rstudioapi::getActiveDocumentContext()$path)
-  } else {
-    git_path <- getwd()
-  }
+  git_path <- getwd()
 
   git_path <- gsub("?", "", git_path)
   git_path <- paste0(git_path, "/")
--- a/0_global_functions.R
+++ b/0_global_functions.R
@@ -32,7 +32,7 @@ is_blank_na <- function(x) {
 }
 
 set_location <- function() {
-  if (rstudioapi::isAvailable()) {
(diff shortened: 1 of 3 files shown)


jdhoffa:
I'm a bit less comfortable with this. Correct me if I am misunderstanding, but basically, there is the potential that these scripts could have different behaviors if run in the exact same setting, just depending on if `rstudioapi` is installed or not? 

ie. If `rstudioapi` is installed, the working directory would be defined by the active file path, and if it is not it would be defined by `getwd()`? What would happen if we just remove the `rstudioapi` calls entirely, and only define paths with `getwd()`? Why did we start using it in the first place?

I don't like the idea of potentially introducing conflicting behaviour based on package availability, and also not calling that package a dependency, that seems sketchy.

cjyetman:
That's a good point @jdhoffa, but here's the thing...
1. I believe that when you install RStudio, it automatically installs `{rstudioapi}`
2. both these files are in the root directory

so unless the user is doing something really weird (e.g. installing `{rstudioapi}` after installing RStudio, moving these files to some other directory, or screwing around with their working directory after opening this project), the result should always be equivalent to `getwd()`, in which case I would support your second suggestion, just getting rid of this stuff altogether and setting `working_location <- getwd()` and `git_path <- getwd()`

I don't know why this was implemented in the first place.

maurolepore:
I think in some case `rstudioapi::isAvailable()` was used to achieve what is best achieved with `interactive()`. If that's the case, then 100% up remove {rstudioapi}.



```r
interactive()
callr::r(function() interactive())

rstudioapi::isAvailable()
callr::r(function() rstudioapi::isAvailable())
```

jdhoffa:
get rid of it!

jdhoffa:
:-)

cjyetman:
cool.... moving to draft until I implement the advice here

cjyetman:
I think @maurolepore is right, `rstudioapi::isAvailable()` was used as a one-to-one replacement for `base::interactive()`. However, I don't know of any `base` equivalent of `rstudioapi::getActiveDocumentContext()$path`, so I'm leaning toward just replacing these bits completely with either `working_location <- getwd()` or `git_path <- getwd()`

PR #560, opened 2021-12-17
depend on individual pkgs rather than all of `{tidyverse}`

cjyetman:
Remove dependency on all of `{tidyverse}` and instead depend on and load/install only the individual underlying packages actually required.

### DIFF
--- a/.github/workflows/testthat-test_local.yml
+++ b/.github/workflows/testthat-test_local.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Install dependencies
         shell: Rscript {0}
         run: |
-          install.packages(c("assertthat", "cli", "config", "conflicted", "countrycode", "devtools", "DT", "fs", "fst", "glue", "here", "janitor", "knitr", "plyr", "r2dii.utils", "remotes", "renv", "rmarkdown", "snakecase", "testthat", "tidyverse", "withr", "yaml", "zoo"))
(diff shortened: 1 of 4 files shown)


jdhoffa (approved):
YAY

maurolepore (approved):
LGTM.

Beyond the scope of this PR but related: To more easily keep docs and code in sync it would be nice remove the explicit name of the depencencies from the docs. Instead we could point with a URL to the file that actually introduces those dependencies, e.g. DESCRIPTION.

cjyetman:
Thanks @maurolepore. In principal, I agree with you, however... we have to keep in mind that at the moment, at least for the transitionmonitor stuff, we need to track all of the dependencies for 5 different repos. That doesn't conflict with what you're saying, but it does mean that what we have in DESCRIPTION here is and very likely will remain an incomplete set of dependencies for what's needed on transitionmonitor.

jdhoffa:
@maurolepore this relates to many conversations Cj and I have been having across a few PRs. I back it. 

We are not there yet, but I definitely feel dependencies should be tracked per repo (so any repo keeps track of only ITS own dependencies), and then workflows that depend on multiple repos (ie. transitionmonitor) just install dependencies of whatever tools they make use of. 

Not there yet, but I think we all agree that we should try to get there :-)

PR #561, opened 2021-12-17
remove `{DT}` as a hard dependency

cjyetman:
Remove `{DT}` as a hard dependency since it is only use in this file [pacta_data_qa.Rmd](https://github.com/2DegreesInvesting/PACTA_analysis/blob/11ea54588f9b618c6f1b09c5f712146f5f0628f3/pacta_data_qa.Rmd)


jdhoffa (approved):
yay

PR #563, opened 2021-12-18
`{withr}` is only used as a development tool

cjyetman:
`{withr}` is only used as a development tool, so can be moved to Suggests

does that sound right to you @jdhoffa @maurolepore


maurolepore (approved):
LGTM.

Yeah, it that sounds right. A search for "withr" shows it's used only in tests and thus can be declared in DESCRIPTION > Suggests.

PR #562, opened 2021-12-17
remove `{plyr}` dependency

cjyetman:
remove `{plyr}` dependency by replacing `plyr::mapvalues()` with `dplyr::recode()`

this still requires testing, so putting in draft mode

related to #137 
related to #184

PR #565, opened 2021-12-18
remove `{conflicted}` dependency

cjyetman:
The purpose/use of `{conflicted}` appears to be only here...

Presumably, the intention is to address the conflict between `dplyr::filter` and `dplyr::lag` and the `{stats}` version of those functions. Since `{stats}` always loads with base R before any other packages are explicitly loaded, I think it's extremely common that when a user loads `{dplyr}`, which itself is extremely common, they understand that `dplyr::filter` and `dplyr::lag` are what will be used. I have verified that there are no current usages of `stats::filter` or `stats::lag`, and frankly I can't imagine why there ever will/would be.

The rest of the functions here, `dplyr::mutate`, `dplyr::here`, `dplyr::rename`, `dplyr::summarise`, and `dplyr::arrange`, are presumably here to avoid a conflict with the `{plyr}` versions of theses functions. I have verified that none of the `{plyr}` versions of those functions are currently in use... the only existing usage of `{plyr}` is two `plyr::mapvalues()`commands, which have already been replaced with `dplyr::recode()` in #562. `{plyr}` itself is on it's way out too with #562, so once that has landed, this usage of `{conflicted}` will also be completely irrelevant.

### DIFF
--- a/.github/workflows/testthat-test_local.yml
+++ b/.github/workflows/testthat-test_local.yml
@@ -36,7 +36,7 @@ jobs:
 
       - name: Install dependencies
         shell: Rscript {0}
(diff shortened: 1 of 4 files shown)


maurolepore (approved):
LGTM

PR #566, opened 2021-12-19
remove `{countrycode}`

cjyetman:
`{countrycode}` does not appear to be used anywhere in this repo, so don't load it or list it as a dependency (it is currently a dependency for the transitionmonitor Docker image)


maurolepore (approved):
LGTM

PR #568, opened 2021-12-20
add `remove_if_exists()` and use it

cjyetman:
To be honest, this feels entirely over-engineered, but the web tool scripts need to clean up after themselves by removing unnecessary objects that might be rather large, however, they don't always exist (e.g. if a portfolio does not have any Bonds), so this is intended to have and use a standardized, consistent way of removing objects without throwing a bunch of warnings if they don't exist.


maurolepore (approved):
LGTM. 

Nit. I would pass a string -- not symbol. This results in a simpler body and a return value that follows the common pattern by which a side-effect function returns the first argument invisibly, so you can use it in a pipe.


```r
remove_if_exists <- function(name) {
  if (exists(name, envir = parent.frame())) {
    rm(list = name, envir = parent.frame())
  }
  invisible(name)
}
```

AlexAxthelm:
@maurolepore Except this function should _not_ be part of a chain.

AlexAxthelm (approved):

PR #569, opened 2021-12-20
add imports for root level scripts

cjyetman:
import some package functions that are used by scripts in the root level so that we don't have to load the packages during the tests to make them work


maurolepore (approved):
LGTM.

Nit. I'm used to importing functions in R/imports.R and leave R/packagename-package.R untouched -- once created with `usethis::use_package_doc()`. I nice side effect is that I can update R/packagename-package.R with a new call to `use_package_doc()` and not have to account for any import from other packages.

AlexAxthelm (approved):

cjyetman:
@maurolepore I used e.g. `usethis::use_import_from("tibble", "as_tibble")` and that's what it creates 🤷🏻 

can I just remove `R/packagename-package.R`?

maurolepore:
Cool! I didn't know of `usethis::use_import_from()`. Then I prefer your approach to mine. This function seems to be quite clever and sorts packages and functions alphabtically. Up to you though.

maurolepore:
Yes, thanks!

PR #579, opened 2021-12-30
only report invalid holdings that are directly held

cjyetman:
This PR removes all indirect holdings from the `invalidsecurities.csv`

Background:

Including indirect holdings (i.e. holdings that are added by merging our fund data) in the `invalidsecurities.csv` causes a lot of confusion with users. First, whatever value is in the `isin` column for an indirect holding will not be found in the original portfolio CSV, so a user is left to understand somehow that it must be some component of a fund in their portfolio (which, frankly, is unrealistic that they will understand that). Secondly, our fund data requires each fund to have a full 100% weight, so if the original data from Lipper, or other source, does not have the full composition of a fund (very common) then we add a line with "MISSINGWEIGHT" or "MISSINGVALUE" or some variation of that to make up for the difference in the total weight of the fund, so it is very common for a user's `invalidsecurities.csv` to be full of lines with an `isin` of "MISSINGWEIGHT" and they have no realistic way of understanding what that means or why it is there (according to @catarinabrg this is a **VERY** frequent problem with COP participants and causes an enormous amount of questions, feedback, and general angst). Thirdly, the original data from Lipper, or other source, often contains fund components that are not ISINs, but rather one of "Cash", "Tax", "Fees", "Gold", etc. (also very common) and so the user also typically gets a `invalidsecurities.csv` full of lines with these variations as well in the `isin` column, which likewise causes a lot of confusion and questions. Fourthly, the original data from Lipper, or other source, can contain ISINs that are not valid, either because it's bogus data or because it's an ISIN that we don't know about in our financial data, leading to the `invalidsecurities.csv` having rows with ISINs that do not appear in the user's original portfolio CSV, causing confusion with the user because they don't understand why these seemingly random holdings have been added to their portfolio.

Originally, @catarinabrg and @cjyetman had considered adding columns to the `invalidsecurities.csv`, like `direct_holding` and/or `fund_isin`, with the goal of maybe making it more obvious to an astute user that those questionable rows are components of one the funds that they directly hold and was included in their portfolio CSV, and hopefully assume that they should not worry about it and not submit angsty questions about why there are all these weird holdings that we've added to their portfolio. After testing that out a bit, and also realizing how many variations of indirect holdings can show up in the `invalidsecurities.csv`, I don't believe adding these columns is really going to improve the situation.

The goal of the `invalidsecurities.csv` is to allow the user to quickly see the holdings that they submitted that we were unable to process for one reason or another, and give them the opportunity to review that and possibly resubmit their portfolio with fixes if possible. There is nothing that a user will ever be able to do about or fix an "invalid" security that is an indirect holding, because indirect holdings are only ever added by merging in our fund data, so they will never be in the user's original portfolio CSV. So I have come to the conclusion that the sensible thing to do is simply not show any indirect holdings in the `invalidsecurities.csv` at all. They only add confusion, and there's nothing a user will ever be able to do with that information. Additionally, the `audit.csv` will continue to have these holdings, along with more columns/data/information to contextualize what they are, e.g. the `direct_holding` which will be `FALSE`.

I do however suspect that this could be a controversial decision, so I think that we should ask as many people on the PACTA team as is feasible before merging this.

For example, this is the current `invalidsecurities.csv` that is generated for one of our test portfolios...
| isin                          | marketValues | currency | flag                                 |
| ------------------ | -------------  | --------- | ---------------------- |
| WeightGapMarker | 477                  | EUR         | Invalid or missing ISIN |
| NA                           | 477                  | EUR         | Invalid or missing ISIN |
| NA                           | 477                  | EUR         | Invalid or missing ISIN |
| NA                           | 477                  | EUR         | Invalid or missing ISIN |
| NA                           | 477                  | EUR         | Invalid or missing ISIN |
| NA                           | 477                  | EUR         | Invalid or missing ISIN |
| NA                           | 477                  | EUR         | Invalid or missing ISIN |
| NA                           | 477                  | EUR         | Invalid or missing ISIN |
| NA                           | 477                  | EUR         | Invalid or missing ISIN |
| MISSINGWEIGHT  | 477                  | EUR         | Invalid or missing ISIN |

### DIFF
--- a/0_json_functions.R
+++ b/0_json_functions.R
@@ -352,6 +352,7 @@ export_audit_graph_json <- function(audit_file__, export_path_full) {
 
 export_audit_invalid_data <- function(portfolio_total_, export_path_full) {
   portfolio_total_ <- portfolio_total_ %>% subset(flag %in% c("Missing currency information", "Negative or missing input value", "Invalid or missing ISIN"))
+  portfolio_total_ <- filter(portfolio_total_, direct_holding == TRUE)
   portfolio_total_ <- portfolio_total_ %>% select("isin", "market_value", "currency", "flag")
   portfolio_total_ <- portfolio_total_[order(-portfolio_total_$market_value), ]
 


jacobvjk:
seems reasonable to me

Antoine-Lalechere:
Make also sense for me as we have also recently the question from Peruvian Pension funds. 
To be sure that we are on the same page now, the expected behaviour for a portfolio investing in 2 funds FR0000000 and FR00000001 where we only have information for fund FR0000000 should be : 
image
And then about the flag "Not in Bloomberg database" we won't give some more context if it's because the whole fund is not in Bloomberg database or if it's because too much holding from that fund are not in the fund because in both case, the user can't act on it?
In that case, make total sense for me

cjyetman:
> To be sure that we are on the same page now, the expected behaviour for a portfolio investing in 2 funds FR0000000 and FR00000001 where we only have information for fund FR0000000 should be :
> image

to be exact, the `invalidsecurities.csv`, would look like...

| isin                          | marketValues | currency | flag                                 |
| ------------------ | -------------  | --------- | ---------------------- |
| FR00000001          | 100                  | EUR         | Holding not in Bloomberg database |

it would not include FR0000000 because that is not an "invalid security" (assuming it were true that it was flagged as "Included in the analysis"). However, the audit file would include the values shown in your screenshot.

---

> And then about the flag "Not in Bloomberg database" we won't give some more context if it's because the whole fund is not in Bloomberg database or if it's because too much holding from that fund are not in the fund because in both case, the user can't act on it?

If the flag is "Holding not in Bloomberg database", that means that we don't know what that ISIN is. So we don't know if it is a fund or anything else, so we can not process it any further. Technically, that means that...
1. the ISIN was not matched to an ISIN in our financial data from Bloomberg
2. and also... it does not match a know fund ISIN in our fund data from Lipper

AlexAxthelm (approved):

PR #584, opened 2022-01-03
add `is_valid_currency_code()`

cjyetman:
adds an `is_valid_currency_code()` vectorized function that validates a vector of currency codes with `TRUE` or `FALSE`

Valid currency codes must follow the [ISO 4217](https://www.six-group.com/en/products-services/financial-information/data-standards.html#scrollTo=maintenance-agency) alpha code specification. This function leverages the [data built-in](https://github.com/vincentarelbundock/countrycode/blob/main/R/codelist.R) to the `{countrycode}` package which is pulled from the original source periodically.

This also creates a dependency on the `{countrycode}` package, so the DESCRIPTION is updated here (imports are not needed because this function uses a dataset from `{countrycode}` but not a function from it). `{countrycode}` is already a downstream dependency, for instance in [create_interactive_report](https://github.com/2DegreesInvesting/create_interactive_report/blob/e011d87601c7cd016777b4dabc8831e5c6558c94/create_interactive_report.R#L61).


maurolepore (approved):
LGTM :-)

PR #586, opened 2022-01-03
add `is_file_accessible()`

cjyetman:
adds an `is_file_accessible()` vectorized function that validates a vector of filenames or filepaths with `TRUE` or `FALSE`

Filenames and filepaths are expanded to their full, absolute path. They must...
1. be a file (not a directory)
2. exist
3. be readable
4. not be empty (this can be the case if a Dropbox file is not downloaded)

This function has a dependency on the `{fs}` package, which is already a dependency of this package/repo.


cjyetman:
Seems like `Sys.chmod()` must work differently on Windows, or not work at all, hence the fail on the Windows CI test. Will have to look into alternatives for Windows or a way to check if the file permissions can be set properly and if not skip that test. Open to any ideas/suggestions.

R/is_file_accessible.R
```diff
@@ -0,0 +1,28 @@
...
+#' empty files, and they will not pass this validation.
+#'
+#' @param filepaths A character vector
+#'
+#' @return A logical vector the same length as `filepaths`.
+#' @export
+is_file_accessible <- function(filepaths) {
+  if (is.data.frame(filepaths) && identical(length(filepaths), 1L)) {
+    filepaths <- filepaths[[1L]]
+  }
```
maurolepore, line 15: Just curious, why do we expect/accept a dataframe input and not only vectors?

maurolepore (commented):
LGTM

R/is_file_accessible.R
cjyetman: Because tidyverse decided to use a different behavior than base R (for valid reasons that I agree with) when a tibble is subset using `[`.... base R by default "simplifies" to a vector if a data frame is subset to only one column, so it has become very common for users to provide as an argument to a function that is expecting a vector something like `mydata[1]`... but if `mydata` is a tibble, either intentionally or not, then `mydata[1]` will be a data frame with one column. Often users don't even know that their data frame has become a tibble by passing it through a dplyr chain or other common processes. I have dealt with countless questions and support requests related to this over the years, so now I habitually coerce one column data frames to a vector if a vector is what was expected.

cjyetman:
I wasted a bunch of time trying to create a test file with no read access on Windows with variations like...
```
fs::file_create(bar, mode = "000")
fs::file_create(bar, mode = "u-r")
fs::file_chmod(bar2, "000")
fs::file_chmod(bar2, "u-r")
#> ... etc.
```
and nothing seems to work. I always get at least read access on the file on Windows.

Then I found in `{fs}` testing infrastructure skips of read write access ability on Windows here...
https://github.com/r-lib/fs/blob/555821f3e7164a09ceb133a310b0af196b582977/tests/testthat/test-access.R#L11-L21
```r
    it("checks for file read and write ability", {
      skip_on_os(c("windows", "solaris"))
      if (Sys.info()[["effective_user"]] == "root") skip("root user")
      file_chmod("foo/bar", "-rw")
      expect_equal(file_access("foo/bar", "read"), c("foo/bar" = FALSE))
      expect_equal(file_access("foo/bar", "write"), c("foo/bar" = FALSE))


      file_chmod("foo/bar", "+rw")
      expect_equal(file_access("foo/bar", "read"), c("foo/bar" = TRUE))
      expect_equal(file_access("foo/bar", "write"), c("foo/bar" = TRUE))
    })
```

so, I'm gonna take that as a legitimate reason to skip testing no read access on Windows in the tests here.

cjyetman:
Although, I'm not entirely ok with losing the functionality of testing whether a file is readable on Windows... maybe I need to implement something like...
```r
is_readable_file <- function(filepaths) {
  !identical(
    tryCatch(
      expr = {
        con <- suppressWarnings(file(filepaths, open = "r"))
        on.exit(close(con))
      }, 
      error = function(cond) return(FALSE)
    ), 
    FALSE
  )
}
```

cjyetman:
ok, now we have a more robust, cross-platform compatible (i.e. should work on Windows) `is_readable_file()` function, and the tests have been adapted to only test `is_file_accessible()` on files with no read permission on POSIX platforms because otherwise I don't know how to create a test file with no read permission

PR #587, opened 2022-01-04
add `is_binary_file()`

cjyetman:
depends on #586 

adds an `is_binary_file()` vectorized function that checks if files in a vector of filenames or filepaths are binary with `TRUE` or `FALSE`... if the file is inaccessible, it will return `NA` for that element

This function has a dependency on the `{fs}` package, which is already a dependency of this package/repo.


cjyetman:
closed in favor of #591

PR #588, opened 2022-01-04
add `guess_file_encoding()`

cjyetman:
depends on #586 
depends on #591 

adds a `guess_file_encoding()` that will guess the file encoding of a vector of filenames or filepaths and return the file encoding as a string. It primarily uses `stringi::stri_enc_detect()` to guess the encoding. Additionally, it searches for known CP850 and CP1252 characters and will return the appropriate encoding if found, because ICU/stringi cannot detect them. If a file is a binary file, it will return `"binary"`. If a file is inaccessible it will return `NA` for that element.

This PR includes basic tests to verify that the function always returns a character vector under various potential use cases and even when non-string vectors are passed to it. It also includes tests to verify that it correctly identifies a number of files in different encodings.

This function has a dependency on the `{stringi}` package, which is already an implicit dependency of this package/repo.


cjyetman:
I still haven't figured out how to run a few of the tests at the bottom, but I leave them commented out for now.

I'm a bit torn on whether I should return `"binary"`  if it detects a binary/non-text file, or return `NA` same as if the file doesn't exist or is otherwise inaccessible.

R/guess_file_encoding.R
```diff
@@ -0,0 +1,52 @@
...
+      if (!is_file_accessible(filepath)) { return(NA_character_) }
+      if (!is_text_file(filepath)) { return("binary") }
+
+      lines <- stringi::stri_read_raw(filepath)
+
+      if (all(stringi::stri_enc_isascii(lines))) { return("ascii") }
+
+      # Swiss - CP850
+      # u-umlaut - 0x81 - "\x81" - iconv("\x81", "cp850", "UTF-8") - validUTF8("\x81")
+      # o-umlaut - 0x94 - "\x94" - iconv("\x94", "cp850", "UTF-8") - validUTF8("\x94")
```
maurolepore, line 35: :love_you_gesture:

maurolepore (approved):
LGTM. 

Not sure why those tests don't work as you expect but the first, naive thing that comes to mind would be to try `expect_equal()` instead of `*_identical()`.

cjyetman:
> Not sure why those tests don't work as you expect but the first, naive thing that comes to mind would be to try `expect_equal()` instead of `*_identical()`.

The problem with those tests is that I don't know how to create the tempfiles in the appropriate encoding, so the expectations will fail.

PR #589, opened 2022-01-04
add `guess_file_mime_type()`

cjyetman:
depends on #586 

adds a `guess_file_mime_type()` vectorized function that guesses the [MIME type](https://en.wikipedia.org/wiki/Media_type) of each file in a vector of filenames or filepaths and returns a string for each element with the name of the MIME type, or if the file is inaccessible, it will return `NA` for that element

This PR includes basic tests to verify that the function always returns a character vector under various potential use cases and even when non-string vectors are passed to it.

This function has a dependency on the `{wand}` package, which is a new dependency for this package/repo, so DESCRIPTION has been updated and an import was added. `{wand}` maintains an extensive lookup table of file extensions and MIME types, and some heuristics to figure out strange cases, which would be difficult and frankly unreasonable to maintain ourselves here.


cjyetman:
closing in favor of a combination of...
#586 `is_file_accessible()`
#591 `is_text_file()`
#590 `guess_delimiter()`

PR #591, opened 2022-01-05
add `is_text_file()`

cjyetman:
depends on #586 
depends on 2DegreesInvesting/docker/pull/57

supercedes #587 (found a more robust way of testing if a file is text or not, which was the original intent for `!is_binary_file()`

adds an `is_text_file()` vectorized function that checks if files in a vector of filenames or filepaths are text files and returns `TRUE` or `FALSE` for each

It primarily uses the system's `file -b --mime-type` to guess the content type, but if `file` is not found on the system's path (e.g. on Windows without Rtools installed) and/or if the `find` path is not executable it will fall back on using `wand::get_content_type()`, which is R-only and cross platform compatible but less robust.

This PR includes basic tests to verify that the function always returns a logical vector under various potential use cases and even when non-string vectors are passed to it. It also tests that it maintains basic/limited functionality if the system's `file` is not found.

This function has a dependency on the `{fs}` package, which is already a dependency of this package/repo, and the `{wand}` package, which is a new dependency, so it has been added to DESCRIPTION and the appropriate imports have been made.

Because `{wand}` is a new dependency that has not been added to our TM docker image yet, or more precisely to the `2dii/pacta` base Docker image (which is in turned based on `2dii/r-packages` Docker image), the source-web-tool-scripts CI will fail because `{wand}` will not be available when `devtools::load_all()` is called (which attempts to load all packages listed as dependencies even though they may not actually be used/called anywhere in the code we intend to use). I temporarily circumvented this in this PR by adding a install command for `{wand}` in the source-web-tool-scripts CI so that the next to last commit in this PR can pass all tests. Then the final commit is only removing this change so that it won't merge to master, and hence the source-web-tool-scripts CI tests will fail. This won't be a problem anymore once we either add `{wand}` to the `2dii/r-packages` Docker image or we migrate to the new Docker image I've been working on in #528.


cjyetman:
I had to add a check if the file is accessible when the system's `file` command is not available and it has to fallback to using `wand::get_content_type()` because `wand::get_content_type()` will fail with an error if it cannot access the file. Therefore, this PR now depends on #586 for the new `is_file_accessible()` function.

I also added return value tests for both when the `file` command is and is not available.

maurolepore (approved):
LGTM. It looks very robust.

R/is_text_file.R
```diff
@@ -0,0 +1,46 @@
...
+    file_command <- Sys.which("file")
+    if (!is_file_accessible(filepath)) {
+      NA_character_
+    } else if (fs::file_access(file_command, mode = "execute")) {
+      system2(
+        command = "file",
+        args = c("-b", "--mime-type", shQuote(filepath)),
+        stdout = TRUE
+      )
+    } else {
```
AlexAxthelm, line 33: Can we change this to also `require("wand")`, and then throw a warning if it isn't installed? Maybe it should error if it can't check, but won't we see the error in the file failing to open anyways?

AlexAxthelm (changes requested):

R/is_text_file.R
cjyetman: I prefer to have these low level functions just work and not throw a bunch of errors/warnings/messages over bad input. I could add one more `require(wand)` here and then add one more `else` that just returns `NA_character`. 

The intent was for all of these `is_*()` style functions to **always* return `TRUE` or `FALSE` for anything you pass to it.... if it can't definitively guarantee that something "is" what it is supposed to be, it should return `FALSE`, regardless of why it could not verify it.

cjyetman: but.... do you mean use `require(wand)` here and then remove `{wand}` as a dependency? I could do that, but.... honestly `{wand}` is a rather obscure package, so if we don't import it, there's an extremely high probability it won't be installed, so.... might as well just remove it completely.... thoughts?

cjyetman: actually, I'm rather opposed to that now that I think about it because the only likely way a Windows user will have access to the `file` command is if Rtools is installed, which is not too uncommon because it's need to build packages from source, but I suspect there's a substantial chunk of R users that don't have it installed since it's a separate download and install and requires a bit of setup.

AlexAxthelm: > remove `{wand}` as a dependency

I was thinking it could be a `suggests`.

cjyetman: Yeah, I get it. But since `Suggests` packages are not installed by default with `base::install.packages()`, I have to assume most/many Windows users will not have either of the facilities to make this function work (`files` with Rtools, or `{wand}`), and therefore I assume for most/many Windows users this function will return `FALSE` for anything you give it, making it rather useless.

Maybe I should merge into this repo some of the file type guessing based on file extension? e.g. if none of the other tools are available, assume any path that ends in `.csv`, `.tsv`, `.txt`, `.R`, `.Rmd`, `.md`, `.yml`, `.yaml`, etc. is a text file?

cjyetman:
I was under the impression that when we made a change to the [Dockerfile 2dii/r-packages](https://github.com/2DegreesInvesting/docker/blob/master/r-packages/Dockerfile), as I did in [this merged PR](https://github.com/2DegreesInvesting/docker/pull/57), that Docker hub would automatically build it, but that [doesn't seem to be the case](https://hub.docker.com/r/2dii/r-packages/builds). Maybe @maurolepore knows how to trigger that?

This is one of the reasons I'd eventually prefer to move to a fully self-contained Dockerfile in this repo, as in #528

maurolepore:
@cj it seems that the feature is now limited to PRO accounts so we lost it.

> This feature requires a Docker subscription -- https://docs.docker.com/docker-hub/builds/

cjyetman:
> @cj it seems that the feature is now limited to PRO accounts so we lost it.
> 
> > This feature requires a Docker subscription -- https://docs.docker.com/docker-hub/builds/

ugh, thanks... so I guess we now **have to** migrate to a fully self-contained Dockerfile in this repo @AlexAxthelm

AlexAxthelm:
Good to know. Doesn't seem to be a big issue though.

R/is_text_file.R
cjyetman: I've completely reimplemented this function by checking for nul bytes in the beginning of the file, which has no package dependencies and no dependency on the system's `file` command.

cjyetman:
I've completely reimplemented this function by checking for nul bytes in the beginning of the file, which has no package dependencies and no dependency on the system's file command.

Sorry for the run-around, but I think this is ready to go now.

R/is_text_file.R
```diff
@@ -0,0 +1,36 @@
...
+#' column data.frame to the `filepaths` argument, this function will give an
+#' error.
+#'
+#' @param filepaths A character vector
+#'
+#' @return A logical vector the same length as `filepaths`.
+#' @export
+is_text_file <- function(filepaths) {
+  filepaths <- simplify_if_one_col_df(filepaths)
+  stopifnot("`filepaths` must be a character vector" = typeof(filepaths) == "character")
```
maurolepore, line 22: Interesting!

maurolepore (approved):
LGTM

PR #597, opened 2022-01-10
add `rds_or_rda()`

cjyetman:
This function will determine if a file is a `.rds` or `.rda`/`.RData` file and return `"rds"` or `"rda"` respectively, or return `NA` if it is neither or if the file is inaccessible. It uses the magic bits of the file to make a determination, so it is not dependent on the file name extension, and it automatically adapts to whether the file was saved as binary or ascii and if it is compressed or not.

rda magic bits were inferred from [R source code](https://github.com/wch/r-source/blob/79298c499218846d14500255efd622b5021c10ec/src/library/base/R/load.R#L22-L41)
rds magic bits were determined through experimentation


maurolepore (approved):
LGTM. 

I note two small things. I frame them as questions but I need no answer -- it's just a way to spark reflexion:

1. Is `NA_character` more useful than say, "none"? (Something like "none" seems more stable, and might be easier to work with in for example a `switch()`).
2. How easy is to read a test failure if it is caused by, say "rda_ascii"? (While having one test per case would add duplication if might result in more readable failures).

cjyetman:
> 1. Is `NA_character` more useful than say, "none"? (Something like "none" seems more stable, and might be easier to work with in for example a `switch()`).

That's a good question, and I'm not sure which is better. I leaned towards `NA_character` because `NA` is an ubiquitous response in R when the function doesn't have an appropriate result with the inputs you gave (but didn't error), so if a user passes a vector of 1000 file paths to this function and there's a few `NA`s in the result, I would guess it's likely that they will notice that and intuitively understand what that means. Whereas if the 1000 element character vector returned has all non-`NA` values in it, and just one or two `"none"` values, they would likely not notice it, believe that everything went normal, and they would have no way of knowing how to check for it or filter them out without reading the documentation, because it could also just as easily be "nothing"`, `"undetermined"`, "`unknown`", etc.?

cjyetman:
> 2. How easy is to read a test failure if it is caused by, say "rda_ascii"? (While having one test per case would add duplication if might result in more readable failures).

Do you mean having it like this instead? I actually prefer that style, but I thought the current trend/recommendation was to put expectations together in a single `test_that()`. Is that not true? Maybe @jdhoffa told me that?

```r
test_that("returns expected output for ascii rds", {
   rds_ascii <- withr::local_tempfile()
   saveRDS("xyz", file = rds_ascii, ascii = TRUE, compress = FALSE)
  
   expect_equal(rds_or_rda(rds_ascii), "rds")
 })

test_that("returns expected output for binary rds", {
   rds_binary <- withr::local_tempfile()
   saveRDS("xyz", file = rds_binary, ascii = TRUE, compress = FALSE)
  
   expect_equal(rds_or_rda(rds_binary), "rds")
 })
```

maurolepore:
I read books that recommend opposite things. In tests I tend to prefer readability of the test and the test-failure message over DRY code. It's personal I guess.

cjyetman:
> I read books that recommend opposite things. In tests I tend to prefer readability of the test and the test-failure message over DRY code. It's personal I guess.

awesome, I'm gonna do that then

PR #598, opened 2022-01-10
namespace `testthat::skip()` to avoid R CMD warn

cjyetman:


maurolepore (approved):
LGTM. (If testthat was in Suggests now we would need it in Imports).

cjyetman:
> LGTM. (If testthat was in Suggests now we would need it in Imports).

That's what I thought, but apparently not, and honestly I'm not exactly sure how/why?
https://community.rstudio.com/t/why-are-tests-testthat-helper-files-discouraged-in-testthat/85253/6?u=cjyetman

Maybe that's how some packages have exported functions with "optional" behavior *if* you have a package installed?

cjyetman:
Also, all the R CMD checks passed here 🤷🏻

PR #603, opened 2022-01-16
add `read_portfolio_csv()` and underlying utility functions

cjyetman:
this is still a work in progress

This PR adds new direct dependencies {vctrs} and {progress}, though both are already indirect dependencies, so they've been added to [DESCRIPTION](https://github.com/2DegreesInvesting/PACTA_analysis/blob/read_portfolio_csv()/DESCRIPTION#L42-L50) and the TM [Dockerfile](https://github.com/2DegreesInvesting/PACTA_analysis/blob/read_portfolio_csv()/transitionmonitor_docker/Dockerfile#L99-L109).


cjyetman:
Still waiting for the {fstcore} problem to be resolved so the Linux tests can pass, but otherwise I think this is ready to go.

cjyetman:
{fstcore} issue has resolved itself and Linux tests are now passing again

cjyetman:
force merging because this has been up for review for more than a week

PR #605, opened 2022-01-18
Feature/azure pipelines build only

AlexAxthelm:
include  all the build (not  testing) steps  from Azure Pipelines (so we can at  least be making use of that while we sort out the tests.

This will trigger builds on updates to `master`, `release/*` and PRs to the same, as well as `develop` (in case we start  using git-flow or something similar)


DESCRIPTION
```diff
@@ -36,6 +36,7 @@ Imports:
     fst,
     glue,
     here,
+    jsonlite,
```
cjyetman, line 39: this has been reordered in master, so it's just below.... this is probably unnecessary

tests/testthat/test-0_portfolio_input_check_functions.R
```diff
@@ -0,0 +1,26 @@
+portfolio <- tibble::tribble(
```
cjyetman, line 1: this while file is from somewhere else, no?

DESCRIPTION
AlexAxthelm: I tried to merge in master, and these two files are what came through after using `theirs` as  the merge strategy. I can clean these up.

tests/testthat/test-0_portfolio_input_check_functions.R
AlexAxthelm: should this file be gone entirely?

cjyetman: no... I split them out into tests per function.... maybe you merged master with an old version of master?

cjyetman: that wasn't clear.... yes, this file should be completely gone because it has been superseded by 4 individual test files
- test-add_fin_data.R
- test-add_fund_portfolio.R
- test-calculate_fund_portfolio.R
- test-clean_unmatched_holdings.R

transitionmonitor_docker/Dockerfile
```diff
@@ -118,9 +118,9 @@ RUN Rscript -e "\
 # copy in PACTA repos
 COPY pacta-data /pacta-data
 COPY r2dii.stress.test.data /r2dii.stress.test.data
-COPY PACTA_analysis /bound
-COPY create_interactive_report /create_interactive_report
 COPY r2dii.climate.stress.test /r2dii.climate.stress.test
+COPY create_interactive_report /create_interactive_report
+COPY PACTA_analysis /bound
```
AlexAxthelm, line 123: @cjyetman since these are the most active, I put them at the end, hoping to cache the other 3.

AlexAxthelm:
@cjyetman I think this one is ready. if you want to have a walkthrough on how to access the outputs/results, I can show you no problem.

cjyetman:
> if you want to have a walkthrough on how to access the outputs/results, I can show you no problem.

Definitely! 😁

PR #609, opened 2022-01-27
fix docs of `check_grouped_portfolio_years()` to avoid R CMD check warning

cjyetman:
Ultimately, I "fixed" the documentation so that it is proper, however I disabled the generation of the docs (with `@noRd`) because `check_grouped_portfolio_years()` is not exported and therefore the examples will fail when being tested by R CMD check


jdhoffa (approved):
lgtm

PR #610, opened 2022-01-31
Meta report data creator

jakobwes:
Fixed some minor things when combining the outputs including:

- On portfolio-level when merging the meta portfolio with the rest an error occurred, because portfolio_name for an individual portfolio was type double, but for the meta portfolio was type character
- The naming of meta investor and portfolio was standardized to "Meta Investor" and "Meta Portfolio" to be in agreement with the standard 2dii-naming. This also fixed that in the meta portfolio case often the meta_investor - column was NA.

### DIFF
--- a/meta_report_data_creator/combine_portfolios.R
+++ b/meta_report_data_creator/combine_portfolios.R
@@ -391,7 +391,7 @@ lapply(data_filenames, function(data_filename) {
   all_result_filepaths <- setNames(all_result_filepaths, sub(paste0("^", project_prefix, "_port_"), "", basename(dirname(all_result_filepaths))))
 
   all_results <-
-    map_df(all_result_filepaths, readRDS, .id = "portfolio_id") %>%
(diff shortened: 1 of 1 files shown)


cjyetman:
This reveals two bugs in PACTA that would ideally be fixed rather than hard-coding in workarounds downstream, but I get why this had to be done to get this meta-report thing working in the meantime.

Do you wanna open issues for these?..
- when reading in a portfolio, the investor name and portfolio name are not enforced to be character vectors, so outputs can end up having different columns with different types
- when reading in a portfolio, if it uses new style column names (e.g. "investor_name" instead of "Investor.Name"), new columns will be added instead of modifying the existing investor and portfolio columns (https://github.com/2DegreesInvesting/PACTA_analysis/blob/959ab2cc6c1845908a47782aec19317bc52fcc6d/0_web_functions.R#L178-L184)

cjyetman (approved):

jakobwes:
Created an issue for the [first problem](https://github.com/2DegreesInvesting/pacta.portfolio.analysis/issues/18). I don't really get the second one to be honest. If you could elaborate that would be great!

PR #618, opened 2022-02-15
Fix the save option

### LINKED ISSUE #616: Save/export switch in build_with_tag.sh doesn't seem to work
The "save" switch/option never seems to work for me as expected when I do e.g. `./build_with_tag.sh -s -t 0.3.3`. Any ideas @AlexAxthelm @maurolepore?

https://github.com/2DegreesInvesting/PACTA_analysis/blob/dfa27446159a649fb5b840eff1da7d128d875d1a/transitionmonitor_docker/build_with_tag.sh#L12-L13

https://github.com/2DegreesInvesting/PACTA_analysis/blob/dfa27446159a649fb5b840eff1da7d128d875d1a/transitionmonitor_docker/build_with_tag.sh#L22

https://github.com/2DegreesInvesting/PACTA_analysis/blob/dfa27446159a649fb5b840eff1da7d128d875d1a/transitionmonitor_docker/build_with_tag.sh#L172-L180

cjyetman:
closes #616 

tbh, I didn't test this fully, I just tested a minified example... I guess I/we'll see next time we build a docker image

### DIFF
--- a/transitionmonitor_docker/build_with_tag.sh
+++ b/transitionmonitor_docker/build_with_tag.sh
@@ -2,7 +2,7 @@
 
 # Examples:
 # # The tag is mandatory
-# ./build_with_tag -t 0.1.1
+# ./build_with_tag.sh -t 0.1.1
 
 usage() {
     echo "Usage: $0  -t <docker image tag>" 1>&2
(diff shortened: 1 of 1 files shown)


AlexAxthelm (approved):
also not testeed, but looks good in theory.

cjyetman:
FYI... this seems to work as expected in practice 🎉 
actually speeds up the process quite a bit by saving directly rather than exporting later

PR #623, opened 2022-02-22
Add PA2022CH parameters file

cjyetman:
Adding project parameters file for PA2022CH, based on project parameters file from PA2021NO. As usual, this is being added with no direction whatsoever from the decision makers for this project, so it's questionable if it has the proper settings. Additionally, this parameter file is currently set to use 2020Q4 data, which is known to be incorrect, but is necessary to enable audits to run before the 2021Q4 is available.


catarinabrg (approved):

catarinabrg:
Looks good

Issue #625, opened 2022-03-02
Portfolio with no PACTA relevant data causes error in TDM code


cjyetman:
If you have a portfolio with equity holdings, but none that are in PACTA relevant sectors and none that have PACTA relevant assets, the TDM code will fail with an error and stop everything.

```r
Error in `check_crucial_years()` at PACTA_analysis/R/calculate_tdm.R:278:2:
! Data must contain all crucial years, by group:
x c(2021, 2026, 2030)
ℹ Are all crucial years present in input data?
Run `rlang::last_error()` to see where the error occurred.
```

Here is an example portfolio that will cause this error...
```
Investor.Name,Portfolio.Name,ISIN,MarketValue,Currency
test,test,LU0061462528,1000000,EUR
```

jdhoffa:
@cjyetman what would be the desired behaviour here? 

Perhaps pass an empty dataframe, with the same structure as the typical output of `calculate_tdm()`, along with a warning that no useful data was found?

cjyetman:
I haven't looked into it too deeply to know if that would prevent problems downstream, but yes, that sounds about right.

PR #626, opened 2022-03-08
Add `quit_if_no_pacta_relevant_data()` add use in `web_tool_script_2.R`

cjyetman:
The goal of this is to stop the process during the report generation phase that occurs when a user on TM clicks the "View report" or "View Exec Summary" button, which triggers the running of web_tool_script_2.R and web_tool_script_3.R. It's intended to use the same infrastructure that @AlexAxthelm created to stop and display an error if there are different holding dates in a grouped portfolio. To achieve its goal on the TM server, it needs to make R quit but with a exit code of 0, and it also needs to drop an HTML page in the right location.

Doing this check as soon as possible in web_tool_script_2.R seems like the right thing to do, to avoid to any of the potential heavy processing of matching ABCD etc., though I suppose if the portfolio doesn't have any PACTA relevant holdings it won't take long to process. However, at least currently because of [this issue](https://github.com/2DegreesInvesting/PACTA_analysis/issues/625), the TDM processing will error out, so at least at the moment, this break needs to happen early in the process before any TDM is calculated.

### DIFF
--- a/R/quit_if_no_pacta_relevant_data.R
+++ b/R/quit_if_no_pacta_relevant_data.R
@@ -1,0 +1,30 @@
+quit_if_no_pacta_relevant_data <- function(portfolio) {
+  if (!any(portfolio$has_asset_level_data, na.rm = TRUE)) {
+    error_name <- "Portfolio does not have any holdings with PACTA relevant data"
+
+    desc <- paste(
+      "This portfolio does not have any direct or indirect investments in ",
+      "companies for which we have PACTA relevant, asset based company data."
+    )
+
+    action <- paste(
+      "Try uploading a portfolio with PACTA relavant holdings."
+    )
+
+    log_user_errors(
+      error_name = error_name,
+      suggested_action = action,
+      description = desc,
+      immediate = TRUE
+    )
+
+    if (!interactive()) {
+      # using message and quit, rather than stop(), because server silently fails
+      # if docker container exits with anything other than a 0 status code
+      message("port_holdings_date contains multiple distinct values")
+      quit(save = "no", status = 0L, runLast = FALSE)
+    }
+
+    stop("no PACTA relevant data")
+  }
+}
--- a/R/quit_if_no_pacta_relevant_data.R
+++ b/R/quit_if_no_pacta_relevant_data.R
@@ -8,7 +8,7 @@ quit_if_no_pacta_relevant_data <- function(portfolio) {
     )
 
     action <- paste(
-      "Try uploading a portfolio with PACTA relavant holdings."
+      "Try uploading a portfolio with PACTA relevant holdings."
     )
 
     log_user_errors(
--- a/R/quit_if_no_pacta_relevant_data.R
+++ b/R/quit_if_no_pacta_relevant_data.R
@@ -21,7 +21,7 @@ quit_if_no_pacta_relevant_data <- function(portfolio) {
     if (!interactive()) {
       # using message and quit, rather than stop(), because server silently fails
       # if docker container exits with anything other than a 0 status code
-      message("port_holdings_date contains multiple distinct values")
+      message("Portfolio does not have any holdings with PACTA relevant data")
       quit(save = "no", status = 0L, runLast = FALSE)
     }
 
--- a/tests/testthat/test-quit_if_no_pacta_relevant_data.R
+++ b/tests/testthat/test-quit_if_no_pacta_relevant_data.R
@@ -12,6 +12,7 @@ test_that("`quit_if_no_pacta_relevant_data()` quits if no ABCD in non-interactiv
         log_path <<- outputs_path
         report_path <- file.path(outputs_path, portfolio_name_ref_all, "report")
         dir.create(path = report_path, recursive = TRUE)
+        setwd(here::here())
         quit_if_no_pacta_relevant_data(port_with_no_data)
         TRUE
       },
--- a/tests/testthat/test-quit_if_no_pacta_relevant_data.R
+++ b/tests/testthat/test-quit_if_no_pacta_relevant_data.R
@@ -1,25 +1,24 @@
 test_that("`quit_if_no_pacta_relevant_data()` quits if no ABCD in non-interactive", {
+  tmp_dir <- withr::local_tempdir()
+  port_name <- "TestPort"
   expect_null({
-    callr::r(function() {
-      devtools::load_all()
-      source(here::here("0_global_functions.R"))
-      port_with_no_data <- data.frame(has_asset_level_data = FALSE)
-      outputs_path <<- tempdir()
-      portfolio_name_ref_all <<- "TestPort"
-      log_path <<- outputs_path
-      dir.create(
-        path = file.path(outputs_path, portfolio_name_ref_all, "report"),
-        recursive = TRUE
-      )
-      dir.create(
-        path = file.path("inst", "rmd"),
-        recursive = TRUE
-      )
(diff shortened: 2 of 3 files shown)


AlexAxthelm:
Looks good, but I'm wondering if we should rename `R/user_error_checks.R` in line with this PR.

cjyetman:
Linux tests are failing because of a [temporary issue](https://github.com/fstpackage/fstcore/issues/8) with the CRAN version of {fstcore}, but otherwise tests are passing.

cjyetman:
{fstcore} issue has resolved itself and Linux tests are now passing again

R/quit_if_no_pacta_relevant_data.R
```diff
@@ -0,0 +1,30 @@
+quit_if_no_pacta_relevant_data <- function(portfolio) {
```
jdhoffa, line 1: Should this function be documented?

jdhoffa: My gut feeling is that if the function belongs in its own dedicated file (ie. `quit_if_no_pacta_relevant_data.R`) then it should be documented, otherwise maybe just add it to the bottom of `utils.R`?

R/quit_if_no_pacta_relevant_data.R
```diff
@@ -0,0 +1,30 @@
...
+  if (!any(portfolio$has_asset_level_data, na.rm = TRUE)) {
+    error_name <- "Portfolio does not have any holdings with PACTA relevant data"
+
+    desc <- paste(
+      "This portfolio does not have any direct or indirect investments in ",
+      "companies for which we have PACTA relevant, asset based company data."
+    )
+
+    action <- paste(
+      "Try uploading a portfolio with PACTA relavant holdings."
```
jdhoffa, line 11: Typo: "relevant"

tests/testthat/test-quit_if_no_pacta_relevant_data.R
```diff
@@ -0,0 +1,51 @@
+test_that("`quit_if_no_pacta_relevant_data()` quits if no ABCD in non-interactive", {
```
jdhoffa, line 1: NIT: In my tests, I usually omit the function name (`quit_if_no_pacta_relevant_data()`), since it is assumed that any test in the file `test-quit_if_no_pacta_relevant_data.R` relates to that function). 

That's why you will see tests [here](https://github.com/2DegreesInvesting/r2dii.analysis/blob/main/tests/testthat/test-target_market_share.R) with names like "w/ bad data, errors with informative message"). 

This could, for example, just be "quits if no ABCD in non-interactive session". Just saves clutter when writing many tests.

tests/testthat/test-quit_if_no_pacta_relevant_data.R
```diff
@@ -0,0 +1,51 @@
...
+  skip_if_R_CMD_check()
+
+  tmp_dir <- withr::local_tempdir()
+  port_name <- "TestPort"
+  expect_null({
```
jdhoffa, line 6: Does this `expect_null()` call actually do anything? Can you just run `callr::r`, and then the `expect_true()` later?

web_tool_script_2.R
```diff
@@ -35,6 +35,18 @@ create_portfolio_subfolders(portfolio_name_ref_all)
 
 port_col_types <- set_col_types(grouping_variables, "ddddccccddclc")
 
+
+# quit if there's no relevant PACTA assets --------------------------------
+
+ttlport <- file.path(proc_input_path, portfolio_name_ref_all, "total_portfolio.rda")
```
jdhoffa, line 41: NIT: object could just explicitly be called `total_portfolio` to stick with snake-case?

jdhoffa (commented):
Preliminary comments. Will review locally now.

R/quit_if_no_pacta_relevant_data.R
jdhoffa: Ah, actually just noticed that Alex mentioned changing the name to `user_error_checks.R`. That could also make sense.

tests/testthat/test-quit_if_no_pacta_relevant_data.R
```diff
@@ -0,0 +1,51 @@
...
+    callr::r(
+      func = function(tmp_dir, port_name) {
+        devtools::load_all()
+        source(here::here("0_global_functions.R"))
+        port_with_no_data <- data.frame(has_asset_level_data = FALSE)
+        outputs_path <<- tmp_dir
+        portfolio_name_ref_all <<- port_name
+        log_path <<- outputs_path
```
jdhoffa, line 14: Ok so it looks like the function `quit_if_no_pacta_relevant_data()` actually requires the invisible arguments `outputs_path`, `portfolio_name_ref_all` and `log_path`, (I guess because of the `log_user_errors()` functuon also has those invisible dependencies). 

Changing `log_user_errors` is beyond the scope of this PR, but maybe it's good to make it obvious in the function trace of `quit_if_no_pacta_relevant_data()` that those objects are required...

R/quit_if_no_pacta_relevant_data.R
cjyetman: I'm not opposed to documenting it, but I am opposed to giving ANY suggestion at all that this function should be used by anyone, anywhere other than where I've precisely put it in this PR. Frankly, this is a dangerous function that is only necessary because of a very specific, peculiar behavior of the TM server that we need to play nice with.

On the other hand, I do want this to be/remain tested, and the unique R script for it follows the {testthat} mode of putting a function in it's own script, naming the script as the name of the function, and having a corollary {testthat} script in the tests directory. If there's a better/reasonable way to test it that doesn't conflict with standard tidyverse recommendation (I'm really trying to stick to it for your benefit), then I'm open to those suggestions.

tests/testthat/test-quit_if_no_pacta_relevant_data.R
cjyetman: > Ok so it looks like the function `quit_if_no_pacta_relevant_data()` actually requires the invisible arguments `outputs_path`, `portfolio_name_ref_all` and `log_path`, (I guess because of the `log_user_errors()` functuon also has those invisible dependencies).

💯 

> Changing `log_user_errors` is beyond the scope of this PR, but maybe it's good to make it obvious in the function trace of `quit_if_no_pacta_relevant_data()` that those objects are required...

What exactly does it mean to "make it obvious in the function trace"?

web_tool_script_2.R
cjyetman: I modified this to use separate object names for the two different objects: `total_portfolio_path` and `total_portfolio`
(Hopefully neither of those conflict with one of those magic global parameter object names)

tests/testthat/test-quit_if_no_pacta_relevant_data.R
jdhoffa: I mean adding these arguments directly to the function. I know it won't do much, but at least it will make it very clear when looking at the function that's it is necessary that those objects exist.

cjyetman: I think that would cause more confusion than benefit... if we add those as arguments, whatever gets passed to them will not be used because `log_user_errors()` will still look in the global environment for them, not the function's local environment... so then it's doubling down on the confusion of where those values are coming from.

jdhoffa: Yup that is fair.

cjyetman:
force merging because this has been up for review for more than a week

PR #629, opened 2022-05-12
EDA of FactSet fund data

jdhoffa:
In this PR, I have drafted a script that is an exploratory analysis of FactSet data. I leave the script exploratory for now, so @fariiaakh can follow along with the workflow/ process, however I will tidy up the script more before merging so it is executable and simply pulls the relevant fund by timestamp. 

In the analysis I have: 
* Load up the `fund_data` from the https://github.com/2DegreesInvesting/pacta-data/blob/master/2020Q4/cleaned_files/fund_data.fst
* Pull 1000 random ISINs from that dataset
* Using those 1000 random ISINs, generate an equivalent `fund_data` dataset using FactSet data (from `2020-12-31`)

Some quick coverage stats:
- 100% of the randomly pulled ISINs have an associated factset id (`fsym_id`)
- Of those 1000 ISINs and fsym_ids, 949 had equivalent `factset_fund_id`
- Of all funds in FactSet data, every holding has a valid market value (this is not terribly surprising, but good sanity check)
- When splitting each fund holding by: `c("Fixed Income", "Equity", "Convertible Equity", "Other", "Cash/ST Invts")`, found that 100% of those holdings classified as "Equity" have an associated holding ISIN.
- Of the 949 funds that have `factset_fund_id`s, 620 ISINs have SOME coverage by holding ISIN
- Of those, only 430 funds have a coverage higher than 50% (by ISIN)

ONE IMPORTANT TO DO: I have not expanded potential funds of funds. This may increase coverage dramatically

Despite the potentially not great coverage, I was able to produce an equivalent dataset to what we pull from Lipper, which can be a drop in for the `fund_data.fst`, with equivalent rows and information. 

Given the potential upside to having one single data provider for financial data, my assesment is that FactSet is an appropriate replacement for Lipper fund data.

### DIFF
--- a/FactSet/factset_fund_data.R
+++ b/FactSet/factset_fund_data.R
@@ -1,0 +1,198 @@
+library(here)
+library(fst)
+library(odbc)
+library(RPostgres)
+library(dplyr)
+library(dbplyr)
+library(readr)
+library(readxl)
+library(keyring)
+
+# Connection functions ----------------------------------------------------
+
+connect_factset_db <-
+  function(dbname = "charlie") {
+    require(keyring)
+    require(odbc)
+    require(RPostgres)
+
+    host <- "data-eval-db.postgres.database.azure.com"
+    port <- 5432L
+
+    keyring_service_name <- "2dii_factset_database"
+    username <- "postgres@data-eval-db"
+
+    if (!username %in% key_list(service = keyring_service_name)$username) {
+      key_set(
+        service = keyring_service_name,
+        username = username,
+        prompt = "Enter password for the FactSet database (it will be stored in your system's keyring): "
+      )
+    }
+    password <- key_get(service = keyring_service_name, username = username)
+
+    odbc::dbConnect(
+      drv = RPostgres::Postgres(),
+      dbname = dbname,
+      host = host,
+      port = port,
+      user = username,
+      password = password,
+      options = "-c search_path=fds"
+    )
+  }
+
+# Set global vars ----------------------------------------------------------
+
+data_timestamp <- "2020-12-31"
+
+# Load PACTA-formatted Lipper fund data -----------------------------------
+
+fund_data <- read_fst(here("../pacta-data/2020Q4/cleaned_files/fund_data.fst"))
+
+# Recreate fund data using FactSet ----------------------------------------
+
+# Here I will pull a random sub-set of 1000 funds from our current Lipper fund
+# data, and try to create an equivalent dataset from the FactSet data
+
+# pull a reproducibly random sample of 1000 fund ISINs from our current data
+set.seed(666)
+fund_isins <- distinct(fund_data, fund_isin)
+sampled_fund_isins <- sample_n(fund_isins, 1000)
+
+# Pull the same ISINs from factset, and the corresponding factset_fund_id
+factset_db <- connect_factset_db()
+
+# symbology containing the ISIN to fsym_id link
+sym_isin <- tbl(factset_db, "sym_v1_sym_isin") %>%
+  collect()
+
+# a dataset of the sampled ISINs and the corresponding fsym_id
+data <- sampled_fund_isins %>%
+  left_join(sym_isin, by = c(fund_isin = "isin"))
+
+# FactSet has a dataset linking their internal fund IDs (factset_fund_id) to
+# another internal identifier (fsym_id). Note there may be multiple fsym_id per
+# fund.
+
+# get the fsym_ids that are relevant to our 1000 pulled fund ISINs
+sampled_fsym_ids <- unique(data$fsym_id)
+
+# pull all `factset_fund_id`s that are relevant to our ISIN/ fsym_id list
+own_ent_fund_identifiers <- tbl(factset_db, "own_v5_own_ent_fund_identifiers") %>%
+  filter(fund_identifier %in% sampled_fsym_ids) %>%
+  collect()
+
+# Joining in all identifiers
+data <- data %>%
+  left_join(own_ent_fund_identifiers, by = c(fsym_id = "fund_identifier"))
+
+# Sanity Check: We can see for which ISIN and fsym_id, we have a
+# factset_fund_id. We can begin to get a sense of coverage here, seeing if there
+# are any ISINs in the Lipper dataset, that don't have a fund_id in FactSet
+data %>%
+  mutate(factset_fund_id_exists = !is.na(factset_fund_id)) %>%
+  group_by(factset_fund_id_exists) %>%
+  summarize(isin_count = n())
+# of the 1000 randomly pulled ISIN funds, 59 are missing altogether from the
+# FactSet dataset (around 94% coverage, by ISIN)
+
+sampled_factset_fund_ids <- unique(data$factset_fund_id)
+
+# `own_fund_detail` contains the holding details of each fund, by
+# `factset_fund_id`. Here, we will pull all relevant funds
+own_fund_detail <- tbl(factset_db, "own_v5_own_fund_detail") %>%
+  filter(factset_fund_id %in% sampled_factset_fund_ids) %>%
+  filter(report_date == data_timestamp) %>%
+  select(factset_fund_id, fsym_id, reported_mv) %>%
+  collect()
+
+# connect all the factset funds (where available) to details containing:
+# - fsym_id associated with each holding in fund
+# - reported market value associated with each holding
+data <- data %>%
+  select(fund_isin, factset_fund_id) %>%
+  left_join(own_fund_detail)
+
+# Sanity Check: all funds for which a factset_fund_id exists has complete
+# coverage on reported_mv (complete in the sense that every security has an
+# associated market value, doesn't necessarily mean each fund has comprehensive
+# coverage of all securities)
+data %>%
+  mutate(factset_fund_id_exists = !is.na(factset_fund_id)) %>%
+  filter(factset_fund_id_exists) %>%
+  mutate(mv_exists = !is.na(reported_mv)) %>%
+  group_by(factset_fund_id_exists) %>%
+  summarize(isin_count = n())
+
+# Using the following two issues and asset maps, we can get info on what
+# asset_class is associated with each holding
+ref_v2_issue_type_map <- tbl(factset_db, "ref_v2_issue_type_map")
+ref_v2_asset_class_map <- tbl(factset_db, "ref_v2_asset_class_map")
+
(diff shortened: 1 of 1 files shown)


fariiaakh:
i have yet to look at the code but maybe we can ask Ryan about the chances of us requesting data on specifc funds that show up on a user's portfolio but aren't covered in the factset universe. Is there a chance we can get the data at alll and if so how long might it take, etc. Just a thought

FactSet/factset_fund_data.R
```diff
@@ -0,0 +1,198 @@
...
+# Equity, Fixed Income, Private Equity, Cash, Other
+# We don't expect Private Equity, Cash or Other to necessarily have corresponding ISINs
+# However, the Fixed Income class could be an issue.
+
+# Let's look at coverage by asset_type
+data %>%
+  filter(!is.na(asset_type)) %>%
+  mutate(holding_isin_exists = !is.na(holding_isin)) %>%
+  group_by(fund_isin, asset_type, holding_isin_exists) %>%
+  summarise(reported_mv = sum(reported_mv)) %>%
```
fariiaakh: is this line adding together all the reported mv values cells per fund_id?
image

FactSet/factset_fund_data.R
```diff
@@ -0,0 +1,198 @@
...
+  filter(!is.na(asset_type)) %>%
+  mutate(has_isin = !is.na(holding_isin)) %>%
+  group_by(fund_isin, has_isin) %>%
+  summarise(reported_mv = sum(reported_mv)) %>%
+  group_by(fund_isin) %>%
+  mutate(coverage = reported_mv / sum(reported_mv)) %>%
+  filter(has_isin) %>%
+  select(fund_isin, coverage)
+
+# 617 funds by ISIN
```
fariiaakh: I think this line of code and the code block above is related to this comment from ur PR:

> Of the 949 funds that have factset_fund_ids, 620 ISINs have SOME coverage by holding ISIN

but what does this mean? that not all factset funds have holding information?

FactSet/factset_fund_data.R
```diff
@@ -0,0 +1,198 @@
...
+coverage %>%
+  filter(coverage > 0.5) %>%
+  nrow()
+
+# format the final FactSet data that can be a place in for fund_data
+factset_fund_data <- data %>%
+  group_by(fund_isin) %>%
+  mutate(total_mv = sum(reported_mv)) %>%
+  group_by(fund_isin, holding_isin) %>%
+  summarise(isin_weight = sum(reported_mv)/sum(total_mv)) %>%
```
fariiaakh, line 211: just to clarify, we are adding together the market values related to each fund ID to get the total and then dividing each holding's market value from the same fund with this total? 

The definition of reported market value is this according to factset: Market value held by the fund as reported in the source, represented in USD
so i am still confused as to why the reported market value would change per row... why isn't there a 'holdings value' to show the value of the holdings within the fund? (using the same image as before) 
image

fariiaakh (approved):
LGTM :) I like the EDA script better. easier to understand the process. just have a lot of questions... 😅😅
i will refer back to the EDA script to create a sort of 'funds processing in words' type document.

FactSet/factset_fund_data.R
```diff
@@ -0,0 +1,198 @@
...
+# a dataset of the sampled ISINs and the corresponding fsym_id
+data <- sampled_fund_isins %>%
+  left_join(sym_isin, by = c(fund_isin = "isin"))
+
+# FactSet has a dataset linking their internal fund IDs (factset_fund_id) to
+# another internal identifier (fsym_id). Note there may be multiple fsym_id per
+# fund.
+
+# get the fsym_ids that are relevant to our 1000 pulled fund ISINs
+sampled_fsym_ids <- unique(data$fsym_id)
```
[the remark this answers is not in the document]
jdhoffa: Ya actually it's a totally unnecessary step. I just have a habit of running those commands, whenever I'm about to pass a list to a call to `filter()` (just a common pattern).

FactSet/factset_fund_data.R
jdhoffa: Exactly. Notice how I'm not actually assigning this to anything, I just wanted to print the output to console as a sanity check. It allows you to see, per fund and per asset_class, what is the approximate coverage in terms of ISINs. This is interesting, because we might see that most funds have 100% coverage for the "Equity" asset class, and 0% coverage for the "Cash" asset class, which is legit because these different classes may or may not have ISINs.

FactSet/factset_fund_data.R
jdhoffa: Market value HELD by fund = market value of the holding. 

In the table above: 
- `factset_fund_id` indicates one single fund (a portfolio of holdings)
- `fsym_id` indicates one single holding
- `reported_mv` indicates the reported market value of that holding. 

`reported_holding` and `reported_mv` are similar values, the former is the "share amount" of the holding, the latter is the "market value" of the holding 

To get the market value of the fund, you just have to aggregate the market value of all holdings.

fariiaakh: wow. I can't believe it took me this long to understand this. Thanks Jackson!

jdhoffa:
> LGTM :) I like the EDA script better. easier to understand the process. just have a lot of questions... 😅😅 i will refer back to the EDA script to create a sort of 'funds processing in words' type document.

If you want to explain it in words, I would just do so in an Rmarkdown

FactSet/factset_fund_data.R
cjyetman: > To get the market value of the fund, you just have to aggregate the market value of all holdings.


assuming you have/know *all* holdings

fariiaakh: soo waiittt what if some of the holdings are cash and don't have an isin, then u don't have info on ALL the holdings right? can you still do this aggregation and add the caveat that "non-isin-containing-financial-vehicles" are not included in this aggregation?

cjyetman: that is why there are often holdings (at least in the Lipper data) that have `"Cash"` or `"MissingWeight"` in the `isin` column

cjyetman: and that is why we typically, or should aggregate all non-ivestment vehicle holdings into one `"Other"` or `"MissingWeight"` row

cjyetman: however, it should be a concern if we believe the data *does not* have *all* holdings, including non-investment vehicles, i.e. no isin, holdings.... or if we cannot validate or infer what is missing otherwise

cjyetman: or if we know from elsewhere that the total value of the fund is `100` and we have

| fund_isin  | holding_isin | value |
| ------------- | ------------- | ------------- |
| 1286742  | 9028323  | 20 |
| 1286742  | 4359823  | 70 |

we can infer the rest as

| fund_isin  | holding_isin | weight |
| ------------- | ------------- | ------------- |
| 1286742  | 9028323  | 0.2 |
| 1286742  | 4359823  | 0.7 
| 1286742  | Unknown  | 0.1 |

cjyetman: or if we know from elsewhere that the total value of the fund is 100 and we have

| fund_isin  | holding_isin | value |
| ------------- | ------------- | ------------- |
| 1286742  | 9028323  | 20 |
| 1286742  | 4359823  | 40 |
| 1286742  | Gold  | 10 |
| 1286742  | Silver  | 10 |
| 1286742  | Cash  | 10 |

we can infer

| fund_isin  | holding_isin | weight |
| ------------- | ------------- | ------------- |
| 1286742  | 9028323  | 0.2 |
| 1286742  | 4359823  | 0.4 | 
| 1286742  | Unknown  | 0.4 |

cjyetman: there are many possibilities, but we need an end result that looks like

| fund_isin  | holding_isin | weight |
| ------------- | ------------- | ------------- |
| 1286742  | 9028323  | 0.2 |
| 1286742  | 4359823  | 0.4 | 
| 1286742  | Unknown  | 0.4 |

fariiaakh: but in our case, we have to calculate the weight and the total market value.. we may not necessarily know if whether we have all the holdings or not from the factset data so how can we make the inference..?

cjyetman: and if "Unknown" is >= 0.5 (or some threshold) for any given fund, then we should consider whether or not we consider that a reliable source for that fund

cjyetman: That is why I added the caveat "assuming you have/know all holdings". I didn't work on this, so I'm not sure how @jdhoffa got around that problem. But as I've demonstrated, there are many possible ways of getting around that (I only demonstrated a few).

fariiaakh: i get that but we don't know what the unknown would be since we have to calculate the total market value from individual holding information... like the total may not even be the full total if there is holding information that is missing and we don't have an external source to confirm what the actual total value of the fund is... does what i am saying make sense? so can we really be confident about our numbers?

cjyetman: yes..... like I said, this is something that @jdhoffa should explain

fariiaakh: yea sorry, i didn't see ur other msg before😅. i was trying to formulate my thoughts 😅

jdhoffa: Ya so this could be an issue. The response from Ryan (FactSet) was that the total size of the fund is determined by `reported_mv`, which to my understanding, means aggregating the size of the fund, but there is no proof/ external validation that we are missing weights. 

That said, the FactSet data DOES at least take into account: "Equity", "Fixed Income", "Cash", "Other", etc. , so in principle, we know that most asset classes are covered, the issue would arise from lack of coverage. 

I did manually check the `isin_weight` of a handful of the randomly sampled funds between the `pacta-data/fund_data` and `factset_fund_data`, and found all the weights to be in the same range +/-0.1% or so, but granted this could still be a problem, and not entirely sure how to get around it.

fariiaakh: @AlexAxthelm FYI

jdhoffa: Ya, perhaps you can just elaborate on the last e-mail? 

The question being if there is a way to determine the total fund size other than aggregating `reported_mv`, as a way to cross-validate how much of the fund is covered

jdhoffa:
Closing as this is handled in `2DegreesInvesting/pacta.data.preparation`

PR #632, opened 2022-05-16
Import functions from {fst} that are used

cjyetman:


jdhoffa (approved):
Sorry for being slow on this, lgtm

PR #635, opened 2022-08-04
Prune r folder

jdhoffa:
Removing unused functions from the `R/` folder.

PR #637, opened 2022-08-04
Prune functions from R folder

jdhoffa:
Trying #635 again

Notice I basically only remove two functions:
`get_param()` (and support), which don't seem to be used by anything directly. 
`update_dockerfile_packages()` this seems to be a utility function that maintainers can use to write a new dockerfile with updated dependencies. This seems like overkill to me, but @cjyetman not sure if you use it or think we should keep it?


cjyetman (approved):
thanks... I think the Docker stuff is from a former process we had to build the Docker image, but now it's done with a completely different process in `transitionmonitor_docker/`

